PR #6, opened 2018-01-23
Safe/utility arguments wrapper

NikVolf:
As seen from example, leads to much simpler and error-prone user code

### DIFF
--- a/src/host.rs
+++ b/src/host.rs
@@ -1,7 +1,8 @@
 use std::any::TypeId;
-use value::RuntimeValue;
+use value::{RuntimeValue, TryInto};
 use Error;
 
+/// Safe wrapper for list of arguments
 pub struct RuntimeArgs<'a>(&'a [RuntimeValue]);
 
 impl<'a> From<&'a [RuntimeValue]> for RuntimeArgs<'a> {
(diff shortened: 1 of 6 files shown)


src/host.rs
```diff
@@ -1,7 +1,37 @@
...
+}
+
+impl<'a> RuntimeArgs<'a> {
+
+	/// Extract argument by index `idx` returning error if cast is invalid or not enough arguments
+	pub fn nth<T>(&self, idx: usize) -> Result<T, Error> where RuntimeValue: TryInto<T, Error> {
+		Ok(self.nth_value(idx)?.try_into().map_err(|_| Error::Value("Invalid argument cast".to_owned()))?)
+	}
+
+	/// Extract argument as a runtime value by index `idx` returning error is not enougn arguments
```
pepyakin: typo: enougn

src/host.rs
```diff
@@ -45,8 +75,23 @@ impl Externals for NopExternals {
...
+	use value::RuntimeValue;
+	use super::RuntimeArgs;
+
+	#[test]
+	fn i32_runtime_args() {
+		let args: RuntimeArgs = (&[RuntimeValue::I32(0)][..]).into();
+		let val: i32 = args.nth(0).unwrap();
+		assert_eq!(val, 0);
+	}
+
```
pepyakin: Could you add at least one more test for an error condition?

NikVolf:
@pepyakin addressed

PR #14, opened 2018-01-24
Second iteration on documenation.

pepyakin:
cc #3


src/host.rs
```diff
@@ -62,6 +98,77 @@ impl HostError {
...
+/// 		}
+/// 	}
+/// }
+///
+/// impl HostExternals {
+/// 	fn check_signature(
+/// 		&self,
+/// 		index: usize,
+/// 		signature: &Signature
+///		) -> bool {
```
NikVolf: identation a bit wierd?

PR #15, opened 2018-01-24
Fuzz loading/validation against wabt.

pepyakin:


fuzz/fuzz_targets/load.rs
```diff
@@ -0,0 +1,13 @@
+#![no_main]
+#[macro_use] extern crate libfuzzer_sys;
+extern crate wasmi;
+extern crate wabt;
+
+fuzz_target!(|data: &[u8]| {
+    let wasmi_result = wasmi::load_from_buffer(data);
```
NikVolf: identation? :)

pepyakin: Fixed!

PR #24, opened 2018-01-28
Deref for NotStartedModule

NikVolf:


pepyakin:
I’m mildly against this change. Because user should understand that the module isnt fully initialized. So forcing user to do inconvenient call to intermediate fn to get module seems good to me

PR #29, opened 2018-01-31
Wasm function can only trap

pepyakin:
Wasm function can fail only by trapping (e.g. invoking wasm function can't result in linkage error). This means we can benefit from introducing special error type for traps.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -116,6 +116,7 @@ pub enum Trap {
 	DivisionByZero,
 	InvalidConversionToInt,
 	StackOverflow,
+	Host(Box<host::HostError>),
 }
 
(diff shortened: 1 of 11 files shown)


src/lib.rs
```diff
@@ -106,6 +106,19 @@ use std::fmt;
 use std::error;
 use std::collections::HashMap;
 
+#[derive(Debug)]
+pub enum Trap {
```
NikVolf: Oh its' pretty easy to document this from official spec?

pepyakin: Yup! TBH I just forgot to add the docs :s

pepyakin:
@NikVolf please check

PR #33, opened 2018-02-01
Fix memory max pages

pepyakin:


src/memory.rs
```diff
@@ -98,14 +98,18 @@ impl MemoryInstance {
...
+
+		if initial_pages > LINEAR_MEMORY_MAX_PAGES {
+			return Err(Error::Memory(format!("initial memory size must be at most {} pages", LINEAR_MEMORY_MAX_PAGES)));
+		}
+		if maximum_pages > LINEAR_MEMORY_MAX_PAGES {
+			return Err(Error::Memory(format!("maximum memory size must be at most {} pages", LINEAR_MEMORY_MAX_PAGES)));
+		}
+
+		let maximum_size = maximum_pages.saturating_mul(LINEAR_MEMORY_PAGE_SIZE);
+		let initial_size = calculate_memory_size(0, limits.initial(), maximum_size).expect("ok");
```
NikVolf: expect should describe why it can't ever fail

pepyakin: nice catch!

Issue #28, opened 2018-01-29
Add origin information about Wasm traps


tailhook:
Hi,

Is there any way to debug where trap happened? At least function would be good, file+line and/or traceback is even better.

I'm using rust's wasm32-unknown-unknown for generating wasm code itself (and the trap is a panic). Maybe that's something to tweak at the compiler side?

pepyakin:
In general, there is no way to get location of the trap yet (although this should be very useful, and I think this is somwthing we want sooner rather than later)

However, it is possible to print information provided in panic payload, at least if you have a custom panic fmt. But im not sure how panics are reported in wasm32-unknown-unknown with std.

tailhook:
> However, it is possible to print information provided in panic payload, at least if you have a custom panic fmt. But im not sure how panics are reported in wasm32-unknown-unknown with std.

It looks like payload is ignored by rust compiler: https://github.com/rust-lang/rust/blob/master/src/libpanic_unwind/wasm32.rs#L28

Also, what do you mean by "print"? I don't think there is some builtin way to print something in wasm.

pepyakin:
AFAIK, actual printing happens in the panic hook, ie [here](https://github.com/rust-lang/rust/blob/92c32d2d8ca4dc7fb8ddf9296580b0c573350c6a/src/libstd/panicking.rs#L370) which happens before `libpanic_*` takes control (just take a look at other impls or at libpanic_abort).

> Also, what do you mean by "print"? I don't think there is some builtin way to print something in wasm.

Yeah, you're right! However, in some envs such as web JS or wasmi we can define what it means to print something.
In fact, wasm std lib supports basic i/o however under special flag [`DEBUG`](https://github.com/rust-lang/rust/blob/70f7d5842f29d4900f24420b030f144d21f3c5fc/src/libstd/sys/wasm/mod.rs#L39)

Hope this helps!

tailhook:
Wow, that worked like a charm! Thanks. [Here is the code](https://github.com/tailhook/verwalter/commit/82d16ad9b86a1cdf8b7018c1f084f142cc09d902) that catches panics and logs payload and location for anyone trying to figure it out.

However, I think this issue (i.e. tracking traps and not only panics out of the box) depends on [debugging symbols](https://github.com/rust-lang-nursery/rust-wasm/issues/26) and should not be closed yet. Right?

pepyakin:
Im very happy thats worked!

[...]

PR #36, opened 2018-02-05
Refine errors

pepyakin:
- `FuncInstance::invoke` now is public
- `Trap` was separated on struct `Trap` and enum `TrapKind`. This should peave the way to #28

### DIFF
--- a/src/func.rs
+++ b/src/func.rs
@@ -132,10 +132,10 @@ impl FuncInstance {
 	/// # Errors
 	///
 	/// Returns `Err` if `args` types is not match function [`signature`] or
-	/// if [`TrapKind`] at execution time occured.
+	/// if [`Trap`] at execution time occured.
 	///
 	/// [`signature`]: #method.signature
-	/// [`TrapKind`]: #enum.TrapKind.html
+	/// [`Trap`]: #enum.Trap.html
 	pub fn invoke<E: Externals>(
 		func: &FuncRef,
 		args: &[RuntimeValue],
--- a/src/func.rs
+++ b/src/func.rs
@@ -141,7 +141,7 @@ impl FuncInstance {
 		args: &[RuntimeValue],
 		externals: &mut E,
 	) -> Result<Option<RuntimeValue>, Trap> {
-		check_function_args(func.signature(), &args).map_err(|_| Trap::new(TrapKind::UnexpectedSignature))?;
+		check_function_args(func.signature(), &args).map_err(|_| TrapKind::UnexpectedSignature)?;
 		match *func.as_internal() {
 			FuncInstanceInternal::Internal { .. } => {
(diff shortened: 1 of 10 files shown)


src/lib.rs
```diff
@@ -248,15 +290,13 @@ impl error::Error for Error {
 			Error::Table(ref s) => s,
 			Error::Memory(ref s) => s,
 			Error::Global(ref s) => s,
-			Error::Stack(ref s) => s,
 			Error::Value(ref s) => s,
-			Error::Trap(_) => "Trap",
+			Error::Trap(_) => "TrapKind",
```
NikVolf: what this suppose to mean?

src/runner.rs
```diff
@@ -1211,17 +1189,24 @@ impl fmt::Debug for FunctionContext {
...
 
-fn prepare_function_args(signature: &Signature, caller_stack: &mut ValueStack) -> Vec<RuntimeValue> {
-	let mut args = signature.params().iter().cloned().rev().map(|_| {
-		caller_stack.pop().expect("Due to validation stack shouldn't be empty")
-	}).collect::<Vec<RuntimeValue>>();
+fn prepare_function_args(
+	signature: &Signature,
+	caller_stack: &mut ValueStack,
+) -> Vec<RuntimeValue> {
+	let mut args = signature
```
NikVolf: should be refactored to something avoiding double reverse and double clone

src/func.rs
```diff
@@ -128,20 +127,30 @@ impl FuncInstance {
...
+	///
+	/// [`signature`]: #method.signature
+	/// [`Trap`]: #enum.Trap.html
+	pub fn invoke<E: Externals>(
 		func: &FuncRef,
 		args: &[RuntimeValue],
 		externals: &mut E,
 	) -> Result<Option<RuntimeValue>, Trap> {
-		debug_assert!(check_function_args(func.signature(), &args).is_ok());
+		check_function_args(func.signature(), &args).map_err(|_| Trap::new(TrapKind::UnexpectedSignature))?;
```
NikVolf: why not `impl From<TrapKind> for Trap` and just just `?` everywhere?

src/runner.rs
NikVolf: why is it still cloned and collected?..

src/host.rs
```diff
@@ -207,7 +207,7 @@ impl Externals for NopExternals {
 		_index: usize,
 		_args: RuntimeArgs,
 	) -> Result<Option<RuntimeValue>, Trap> {
-		Err(Trap::Unreachable)
+		Err(Trap::new(TrapKind::Unreachable))
```
NikVolf: TrapKind::Unreachable.into()

?

NikVolf: `Err(TrapKind::Unreachable)?`

?

PR #39, opened 2018-02-07
Update testsuite

### LINKED ISSUE #1: Fix spec testsuite
Current testsuite actually is far from current. I tried to upgrade spec suite and it appears that wasmi couldn't pass it (although, I'm not sure why, maybe it is fault in decoding in [parity-wasm](https://github.com/paritytech/parity-wasm).

pepyakin:
Resolves #1 

- add missing validation for exports with duplicate names,
- implement polymorphic stack explicitly, fix related bug,
- hold `last_module` explicitly in spec runner instead of name mangling,
- use `wast2json` in spec runner,
- update fixtures,


src/validation/func.rs
```diff
@@ -610,47 +608,50 @@ impl<'a> FunctionValidationContext<'a> {
...
-		Ok(self.value_stack.top().map(Clone::clone)?)
+	fn tee_value(&mut self, value_type: StackValueType) -> Result<StackValueType, Error> {
+		let value = self.pop_value(value_type)?;
+		self.push_value(value)?;
+		Ok(value)
 	}
 
 	fn unreachable(&mut self) -> Result<(), Error> {
-		Ok(self.value_stack.push(StackValueType::AnyUnlimited)?)
+		let frame = self.frame_stack.top_mut()?;
```
NikVolf: Not sure I agree with this change

Much cleaner to push something to stack than to mutate the last entry in obscure fasion

Is there really no way to preserve former logic?

pepyakin: And I'm not sure about having some special value with unclear behavior to signal that stack is in polymorphic state is much better :-) 

What's bad in taking current frame and marking it is transitioned to polymorphic stack state?

PR #55, opened 2018-02-13
Update wabt version

pepyakin:


NikVolf:
I don't think you need to make pr for this, just update master and publish

pepyakin:
Yep, but I'm lazy and just wanted travis to do the checks for me :P

pepyakin:
but yeah, this might be too much : )

PR #56, opened 2018-02-14
Move spec testsuite into tests.

pepyakin:
Move all spec testsuite into `/tests` directory.

So it means that several changes will happen:

- Compilation will share common dependencies in target directory. This should make CI builds faster. This also should reduce space requirements.
- Common Cargo.lock file
- `cargo test` will include spec tests.
- `cargo check --tests` will also include spec tests.


NikVolf:
Lgtm

PR #57, opened 2018-02-14
Merge signature mismatch traps

pepyakin:
If you look close enough you will realize that these are actually same thing.


pepyakin:
Depends on #56

PR #71, opened 2018-03-11
Args refactor

pepyakin:
When working on a integration of sandbox into a polkadot, I've encountered a situation when I needed to use `RuntimeArgs::nth` in a generic context but was unable to do so since it is impossible to call `nth` with type `T` without a `TryInto` constraint and it is also impossible to specify such constraint since `TryInto` is not exported.

Furthermore, even though `TryInto` is just a utility and user can always define conversion operations by himself, it is infeasible to extract `[RuntimeValue]` from the `RuntimeArgs` since it's not public.

### DIFF
--- a/src/host.rs
+++ b/src/host.rs
@@ -2,7 +2,10 @@ use std::any::TypeId;
 use value::{RuntimeValue, FromRuntimeValue};
 use {TrapKind, Trap};
 
-/// Safe wrapper for list of arguments.
+/// Wrapper around slice of [`RuntimeValue`] for using it
+/// as an argument list conveniently.
+///
+/// [`RuntimeValue`]: enum.RuntimeValue.html
 #[derive(Debug)]
 pub struct RuntimeArgs<'a>(&'a [RuntimeValue]);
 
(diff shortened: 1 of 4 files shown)


NikVolf:
Can you provide an example of what became possible that was not possible before?

pepyakin:
Yeah, sure:

For example, you want to have a method that looks like this:

```rust
fn first_arg(args: RuntimeArgs) -> i32 {
  args.nth(0)
}
```

this will work fine since `RuntimeValue` has an implementation of `TryInto<i32, _>`. 
But when you introduce generic parameter

```rust
fn first_arg<T>(args: RuntimeArgs) → T {
  args.nth(0)
}
```

it won't compile, since `nth::<T>` requires that `T` have a bound `RuntimeValue: TryInto<T, _>` and you have no way to specify it, since you can't import it.

One way to fix it would be to expose the fact that `RuntimeArgs` is essentially just a wrapper around a slice. That at least will allow you to define your own mechanism to convert `RuntimeValue` to some `T`. This also allows you to use an iterator on `RuntimeValue` to get args one by one. I found this useful when working in a macro context: instead of having  `RuntimeArgs` and a position of the current argument, you can just use the iterator for fetching the args.

And exposing the `FromRuntimeValue` (fka `TryInto<T, _>`) trait is allows user to preserves the convenience of using the `RuntimeArgs::nth` in a generic.

So now, it is possible to have a construction like that:

```rust
fn first_arg<T: FromRuntimeValue>(args: RuntimeArgs) → T {
  args.nth(0)
}
```

PR #82, opened 2018-04-04
Fix the name of the 32bits feature

tomaka:
So that it matches the Cargo.toml


pepyakin:
Thanks!

tomaka:
Could you publish a new version of this crate so that I can integrate it in Parity for https://github.com/paritytech/parity/pull/8036 ?

pepyakin:
[Done](https://crates.io/crates/wasmi/0.1.2)!

PR #83, opened 2018-04-16
Optionally deny floating point operations

### LINKED ISSUE #49: Add float ban as optional validation check

Vurich:
Closes #49

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -430,9 +430,7 @@ impl Module {
 		})
 	}
 
-	/// Create `Module` from `parity_wasm::elements::Module`.
-	///
-	/// This function will load, validate and prepare a `parity_wasm`'s `Module`.
+	/// Fail if the module contains any floating-point operations
 	///
 	/// # Errors
 	///
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -472,6 +472,19 @@ impl Module {
 	///
 	/// let module = wasmi::Module::from_buffer(&wasm_binary).expect("Parsing failed");
 	/// assert!(module.deny_floating_point().is_err());
+	///
+	/// let wasm_binary: Vec<u8> =
+	///     wabt::wat2wasm(
+	///         r#"
+	///         (module
+	///          (func $add (param $lhs f32) (param $rhs f32) (result f32)
+	///                get_local $lhs))
+	///         "#,
+	///     )
+	///     .expect("failed to parse wat");
+	///
+	/// let module = wasmi::Module::from_buffer(&wasm_binary).expect("Parsing failed");
+	/// assert!(module.deny_floating_point().is_err());
 	/// ```
 	pub fn deny_floating_point(&self) -> Result<(), Error> {
 		validation::deny_floating_point(&self.module).map_err(Into::into)
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -385,7 +385,6 @@ pub struct Module {
 }
 
 impl Module {
-
 	/// Create `Module` from `parity_wasm::elements::Module`.
 	///
 	/// This function will load, validate and prepare a `parity_wasm`'s `Module`.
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -431,6 +430,55 @@ impl Module {
 		})
 	}
 
+	/// Create `Module` from `parity_wasm::elements::Module`.
```
pepyakin: These seems to be not relevant. Let's update this to actually reflect the method's intention

src/validation/mod.rs
```diff
@@ -50,6 +50,97 @@ impl ::std::ops::Deref for ValidatedModule {
...
+			use parity_wasm::elements::Opcode::*;
+
+			macro_rules! match_eq {
+                    ($pattern:pat) => {
+                        |val| if let $pattern = *val { true } else { false }
+                    }
+                }
+
+			const DENIED: &[fn(&Opcode) -> bool] = &[
+				match_eq!(F32Load(_, _)),
```
pepyakin: nice!

src/validation/mod.rs
```diff
@@ -50,6 +50,97 @@ impl ::std::ops::Deref for ValidatedModule {
...
+				match_eq!(I64TruncSF32),
+				match_eq!(I64TruncUF32),
+				match_eq!(I64TruncSF64),
+				match_eq!(I64TruncUF64),
+				match_eq!(I32ReinterpretF32),
+				match_eq!(I64ReinterpretF64),
+			];
+
+			if DENIED.iter().any(|is_denied| is_denied(op)) {
+				return Err(Error(format!("Unimplemented opcode: {:?}", op)));
```
pepyakin: How about returning something like more specific, like "usage of a forbidden opcode"?

pepyakin:
the thing that worries me is floats are still possible to use, e.g.:

```WebAssembly
(module
   (func (export "derp") (param f32) (result f32)
     get_local 0     
   )
)
```

I didn't found a case where it would be a problem so far, at least in our specific use cases (i.e kovan-wasm). @NikVolf what do you think ?

src/lib.rs
Vurich: Ah crap, I copied it from another method

Vurich:
@pepyakin Isn't that denied by denying `F32Load`?

NikVolf:
Floats are already disabled by gas metering instrumentation in kovan-wasm, so feel free to postulate your own requirements for wasmi @pepyakin

pepyakin:
No, that code will be allowed (will pass the `deny_floating_point` check) since it doesn't contain `f32.load` (or any other fp specific) operations: instructions such as `get_local`/`get_global` etc are generic and their type depends on the location they refer to.

In my understanding, without any floating point operations there is nothing the user can do with these values, only juggling with them with `get_local`/`get_global`/`call`/etc, this should be ok but i'm not 100% sure.

> Floats are already disabled by gas metering instrumentation in kovan-wasm

Ah, indeed, this isn't currently a problem in kovan-wasm and I don't yet see any further uses of float ban. Do we actually need one in `wasmi`?

Vurich:
Isn't the reason that we want to deny floating point to avoid differences in rounding? I don't think it's a problem to allow `get_{global,local}`. I guess we can allow `{f32,f64}.load` too

sunfishcode:
WebAssembly scalar floating point has completely deterministic rounding. The parts that are nondeterministic are the bits of NaN values returned from arithmetic instructions, and support for subnormal values in proposed SIMD arithmetic instructions.

pepyakin:
Hm, I had an impression that we want to deny floating point ops to avoid differences in bit patterns of `NaN` and I wasn't aware of anything regarding to rounding. The [spec states](https://webassembly.github.io/spec/core/exec/numerics.html#rounding) that the rounding is fully defined operation (am I correct?)

And operations with NaN values aren't, their payload might be different on the input values and the exact operation. If you store this value with `fN.store` (or somehow observe their bitpattern, i guess  `copysign`?) you can have non deterministic result thus consensus problems.

PR #90, opened 2018-05-23
Add a simple bench

pepyakin:
A step forward on #5


benches/src/lib.rs
```diff
@@ -0,0 +1,44 @@
...
+	let mut buf = Vec::new();
+	file.read_to_end(&mut buf)?;
+	Ok(Module::from_buffer(buf)?)
+}
+
+#[bench]
+fn bench_tiny_keccak(b: &mut Bencher) {
+	let wasm_kernel = load_from_file(
+		"./wasm-kernel/target/wasm32-unknown-unknown/release/wasm_kernel.wasm",
+	).expect("failed to load wasm_kernel. Is `build.rs` broken?");
```
NikVolf: expect could be inside the `load_from_file`

benches/src/lib.rs
```diff
@@ -0,0 +1,44 @@
...
+
+	let instance = ModuleInstance::new(&wasm_kernel, &ImportsBuilder::default())
+		.expect("failed to instantiate wasm module")
+		.assert_no_start();
+
+	let test_data = match instance
+		.invoke_export("prepare_tiny_keccak", &[], &mut NopExternals)
+		.unwrap()
+	{
+		Some(v @ RuntimeValue::I32(_)) => v,
```
NikVolf: `assert_matches!` ?

benches/wasm-kernel/src/lib.rs
```diff
@@ -0,0 +1,54 @@
...
+				data: &DATA,
+				result: &mut RESULT,
+			});
+		}
+		TEST_DATA.as_ref().unwrap() as *const TinyKeccakTestData
+	}
+}
+
+#[no_mangle]
+pub extern "C" fn bench_tiny_keccak(test_data: *const TinyKeccakTestData) {
```
NikVolf: the whole function could be just unsafe

benches/wasm-kernel/src/lib.rs
```diff
@@ -0,0 +1,54 @@
...
+	unsafe {
+		let mut keccak = Keccak::new_sha3_256();
```
dvdplm: Not exactly sure what it is we're benchmarking here, but just thought I'd point out that this is *not* the keccak used in parity; this one uses the `0x06` delimiter whereas the "normal" keccak 256 uses `0x01`.

benches/src/lib.rs
pepyakin: True, but in this case it will take a few more expects with less clear and/or redundant messages. Precisely because I wanted to avoid this and use `?` I extracted one expect to the outside of the function.

benches/src/lib.rs
pepyakin: I think it will do.

benches/wasm-kernel/src/lib.rs
pepyakin: That's true, 

but motivation behind using `unsafe` as a block is that the block is much more visible than the cluttered function signature.

Compare:

```rust
pub extern "C" unsafe fn bench_tiny_keccak(test_data: *const TinyKeccakTestData) {
    let mut keccak = Keccak::new_sha3_256();
    keccak.update((*test_data).data);
    keccak.finalize((*test_data).result);
}
```

and

```rust
#[no_mangle]
pub extern "C" fn bench_tiny_keccak(test_data: *const TinyKeccakTestData) {
    unsafe {
        let mut keccak = Keccak::new_sha3_256();
        keccak.update((*test_data).data);
        keccak.finalize((*test_data).result);
    }
}
```

In my opinion, the `unsafe` block in the latter example standouts more than in the former one.

However, I think this a insubstantial matter and I'm ready to change it if you wish.

benches/wasm-kernel/src/lib.rs
pepyakin: Nice to know, thanks! 

TBH I didn't bothered too much and I just have ~ported~^W stolen the bench from [tiny-keccak repo](https://github.com/debris/tiny-keccak/blob/7d8d499f8027044b38f73b30d130c36b6859faed/benches/sha3.rs#L16-L26) 😄

NikVolf: better to use keccak224-512, it's not regular sha3

pepyakin: Could you expand on that? I really have no idea why keccak224-512 is better over this sha3-256 for benchmarking

pepyakin: Changed to `keccak256`!

PR #91, opened 2018-05-31
Provide direct access to the underlying buffer

### LINKED ISSUE #27: Ensure we can have memory views directly to linear memory

pepyakin:
This allows zero-copy access to the linear memory.

Closes #27

### DIFF
--- a/src/memory.rs
+++ b/src/memory.rs
@@ -48,7 +48,7 @@ impl ::std::ops::Deref for MemoryRef {
 ///
 /// [`LINEAR_MEMORY_PAGE_SIZE`]: constant.LINEAR_MEMORY_PAGE_SIZE.html
 pub struct MemoryInstance {
-	/// Memofy limits.
+	/// Memory limits.
 	limits: ResizableLimits,
 	/// Linear memory buffer.
 	buffer: RefCell<Vec<u8>>,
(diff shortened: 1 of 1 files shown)


src/memory.rs
```diff
@@ -330,14 +330,43 @@ impl MemoryInstance {
...
+	/// # Panics
+	///
+	/// Any calls that requires either read or write access to memory (such as [`get`], [`set`], [`copy`], etc) made
+	/// within the closure will panic. Proceed with caution.
+	///
+	/// [`get`]: #method.get
+	/// [`set`]: #method.set
+	/// [`copy`]: #method.copy
+	pub fn with_direct_access_mut<R, F: FnOnce(&mut [u8]) -> R>(&self, f: F) -> R {
+		let mut buf = self.buffer.borrow_mut();
```
NikVolf: why not make it panic-free with `try_borrow` and `Result`, or at least add such option?

pepyakin: Indeed we can return a `Result` here but it will not solve the general problem. 

Consider:

```rust
let same_mem = mem.clone();
mem.with_direct_access_mut(|buf| 
  same_mem.set(0, &[1, 2, 3])?;
  Ok(())
)?;
```

In this example, `set` will still panic when called.

I think it might be possible to solve this issue by re-architecting whole memory API, however, I think it's not feasible right now.

PR #96, opened 2018-06-18
Remove VecDeque

Vurich:
This is unnecessary and about 5% slower than using `Vec` on the benchmark


pepyakin:
Haha, I'm working on optimizing wasmi right now (not yet published the PR) and this is one of the optimizations I've also done! : )

Vurich:
@pepyakin I'll close this one then and let it be a part of yours :smile:

PR #97, opened 2018-06-18
Add some more benches

pepyakin:


benches/wasm-kernel/src/rev_complement.rs
```diff
@@ -0,0 +1,155 @@
...
+		*y = table[replace(x, table[*y as usize]) as usize];
+	}
+}
+
+/// Compute the reverse complement on chunks from opposite ends of a sequence.
+///
+/// `left` must start at the beginning of a line. If there are an odd number of
+/// bytes, `right` will initially be 1 byte longer than `left`; otherwise they
+/// will have equal lengths.
+fn reverse_complement_left_right(mut left: &mut [u8],
```
Vurich: You shouldn't have indentation that relies on the length of identifiers if you're using tabs for indentation. Just `rustfmt` and this will be fixed.

benches/wasm-kernel/src/lib.rs
```diff
@@ -48,3 +40,84 @@ pub extern "C" fn bench_tiny_keccak(test_data: *const TinyKeccakTestData) {
 		keccak.finalize((*test_data).result);
 	}
 }
+
+pub struct RevComplementTestData {
+	input: ManuallyDrop<Box<[u8]>>,
```
Vurich: :+1: great, I really like this change

benches/wasm-kernel/src/lib.rs
```diff
@@ -48,3 +40,84 @@ pub extern "C" fn bench_tiny_keccak(test_data: *const TinyKeccakTestData) {
...
+	unsafe {
+		(*test_data).input.as_mut_ptr()
+	}
+}
+
+#[no_mangle]
+pub extern "C" fn bench_regex_redux(test_data: *mut RegexReduxTestData) {
+	unsafe {
+		let result = regex_redux::run(&*(*test_data).input);
+		(*test_data).output = Some(result);
```
Vurich: This doesn't drop the old data any more, I was thinking we avoid dropping the long-lived data but the data that we only need for one iteration (i.e. this data) gets dropped.

benches/wasm-kernel/src/regex_redux.rs
```diff
@@ -0,0 +1,17 @@
...
+//!
+//! [1]: https://benchmarksgame-team.pages.debian.net/benchmarksgame/program/regexredux-rust-2.html
+
+lazy_static! {
+	static ref REGEX: ::regex::bytes::Regex =
+		{ ::regex::bytes::Regex::new("agggtaa[cgt]|[acg]ttaccct").unwrap() };
+}
+
+pub fn prepare() {
+	&REGEX;
```
Vurich: This should be `&*REGEX`, AFAIK currently this doesn't do anything. Even better:

```rust
pub fn prepare() {
    // Force initialization of REGEX
    ::lazy_static::initialize(&REGEX);
}
```

Using the [`initialize` function in `lazy_static`](https://docs.rs/lazy_static/1.0.1/lazy_static/fn.initialize.html)

pepyakin: Oh, that's really nice!

benches/wasm-kernel/src/lib.rs
pepyakin: Hm, are you sure? My impression was that it only doesn't drop contained data on dropping of `ManuallyDrop` itself; however, an assignment is a different case. 

I'd expect that it desugars like that:

```rust
let tmp = (*test_data).output;
drop(tmp)
(*test_data).output = new_output;
```

benches/wasm-kernel/src/regex_redux.rs
pepyakin: This one is really nice. It is much cleaner than obscure `&REGEX`/`&*REGEX`

benches/wasm-kernel/src/lib.rs
pepyakin: Yep, can confirm that:

https://play.rust-lang.org/?gist=1f13c35e1f5f03e99757e7737e50fcb6&version=stable&mode=debug

benches/wasm-kernel/src/lib.rs
pepyakin: yeah, a similar thing, but without unsafe : )

benches/wasm-kernel/src/lib.rs
Vurich: That's really strange, I wouldn't have expected that at all. I guess that's just how the semantics of `DerefMut` works though, so it makes sense.

PR #98, opened 2018-06-18
Optimizations

pepyakin:
Based on #97 

Wasm isn't the greatest instruction set for interpretation. The reason for that is that Wasm is more optimized for ease of validation and compilation. 

The central theme of this PR is a transformation of Wasm structured stack machine into a plain stack machine. This transformation is performed once before the interpretation alongside with validation.

See "isa.rs" for more details on this topic.

Beyond this there a couple of optimizations here and there. 

With all of this, the overall increase in the performance is about almost 2x on some benchmarks.


src/common/stack.rs
```diff
@@ -43,19 +42,17 @@ impl<T> StackWithLimit<T> where T: Clone {
...
 
-	pub fn limit(&self) -> usize {
-		self.limit
-	}
-
 	pub fn top(&self) -> Result<&T, Error> {
+		let len = self.values.len();
 		self.values
-			.back()
+			.get(len - 1)
```
Vurich: This should be `.last` and `.last_mut`

src/isa.rs
```diff
@@ -0,0 +1,298 @@
...
+	/// Similar to the Wasm ones, but instead of a label depth
+	/// they specify direct PC.
+	Br(Target),
+	BrIfEqz(Target),
+	BrIfNez(Target),
+
+	/// Last one is the default.
+	///
+	/// Can be less than zero.
+	BrTable(Box<[Target]>),
```
Vurich: I dislike the pointer indirection here, it's not important to fix now but I wonder if we can do something about it. We could have `BrTable` have one field (the length) and then have a series of `Br` instructions afterwards that act as the list of targets.

src/runner.rs
```diff
@@ -3,110 +3,132 @@ use std::ops;
...
-	Branch(usize),
+	/// Branch to an instruction at the given position.
+	Branch(isa::Target),
 	/// Execute function call.
 	ExecuteCall(FuncRef),
-	/// End current frame.
-	End,
 	/// Return from current function block.
-	Return,
+	Return(isa::DropKeep),
```
Vurich: Can the `drop`/`keep` instructions be encoded as extra instructions at the end of the function (so that we don't have to have special-case handling)? That would mean we can make `Branch` and `Return` the same here.

src/runner.rs
```diff
@@ -3,110 +3,132 @@ use std::ops;
...
+					// Weird, but bail out anyway.
+					|_| Trap::from(TrapKind::StackOverflow)
+				)?;
+		}
+
+		let initial_frame = FunctionContext::new(func.clone());
 
-		let mut function_stack = VecDeque::new();
-		function_stack.push_back(context);
+		let mut call_stack = Vec::new();
```
Vurich: Can we cache this vec in `self` so that we don't need to realloc every call to `start_execution`? Especially useful because we start empty. We just call `Vec::clear` after `run_interpreter_loop`.

src/runner.rs
```diff
@@ -117,322 +139,274 @@ impl<'a, E: Externals> Interpreter<'a, E> {
...
-			&Opcode::I64ReinterpretF64 => self.run_reinterpret::<F64, i64>(context),
-			&Opcode::F32ReinterpretI32 => self.run_reinterpret::<i32, F32>(context),
-			&Opcode::F64ReinterpretI64 => self.run_reinterpret::<i64, F64>(context),
+		Ok(RunResult::Return)
+	}
+
+	#[inline(always)]
+	fn run_instruction(&mut self, context: &mut FunctionContext, instruction: &isa::Instruction) -> Result<InstructionOutcome, TrapKind> {
+		match instruction {
+			&isa::Instruction::Unreachable => self.run_unreachable(context),
```
Vurich: Are these `run_*` functions inlined by LLVM? That'd be a really good source of easy wins if not.

src/runner.rs
```diff
@@ -1252,65 +1108,92 @@ pub fn check_function_args(signature: &Signature, args: &[RuntimeValue]) -> Resu
 	Ok(())
 }
 
+#[derive(Debug)]
 struct ValueStack {
```
Vurich: Why `Box<[RuntimeValue]>` + `usize` and not `Vec`? This seems basically equivalent to a `Vec`.

src/validation/func.rs
```diff
@@ -99,397 +223,1071 @@ impl Validator {
...
+			GetGlobal(index) => {
+				Validator::validate_get_global(context, index)?;
+				context.sink.emit(isa::Instruction::GetGlobal(index));
+			}
+			SetGlobal(index) => {
+				Validator::validate_set_global(context, index)?;
+				context.sink.emit(isa::Instruction::SetGlobal(index));
+			}
+
+			I32Load(align, offset) => {
```
Vurich: Could you generate this with a macro? It seems error-prone.

src/validation/util.rs
```diff
@@ -10,14 +10,36 @@ use validation::Error;
...
-	pub fn new(params: &'a [ValueType], local_groups: &'a [Local]) -> Locals<'a> {
-		Locals {
+	/// Create a new wrapper around declared variables and parameters.
+	pub fn new(params: &'a [ValueType], local_groups: &'a [Local]) -> Result<Locals<'a>, Error> {
+		let mut acc = params.len() as u32;
+		for locals_group in local_groups {
+			acc = acc
+				.checked_add(locals_group.count())
+				.ok_or_else(||
+					Error(String::from("Locals range no in 32-bit range"))
```
Vurich: Typo: should say "not"

src/isa.rs
pepyakin: https://github.com/paritytech/wasmi/issues/100 !

src/runner.rs
pepyakin: What kind of special-case handling do you mean? Also, btw, `drop`/`keep` are not instructions per se, but immediate arguments of these instructions.

src/common/stack.rs
pepyakin: Lol definitely!

src/runner.rs
pepyakin: That will not make any difference since for now, an `Interpreter` is disposable. That is, every call to `FuncInvoke::invoke` will create a new instance of the `Interpreter` and then drop it after it finishes the execution.

Thankfully, this happens only once when invoking a wasm module exported function, which should be pretty infrequent.

src/runner.rs
pepyakin: Not sure about all of them, but I experimented with inlining of hot functions (such as `set_local`/`get_local` for `fac-opt`) and it didn't make a difference.

src/runner.rs
pepyakin: That's actually for a reason. This way it's more explicit that we don't use `Vec`, and in particular, that we are trying to avoid `Vec::push` method.

src/validation/func.rs
pepyakin: Yeah, you're right but I would rather not touch this code! All instructions, which happened to be placed after `TeeLocal`, were mechanically and carefully edited with the help of multiline cursor feature :D

src/validation/util.rs
pepyakin: Thanks!

src/runner.rs
Vurich: Special case handling of returning vs branching. Especially in a stack machine you shouldn't need to have a difference.

src/runner.rs
```diff
@@ -1252,65 +1108,92 @@ pub fn check_function_args(signature: &Signature, args: &[RuntimeValue]) -> Resu
...
-	stack_with_limit: StackWithLimit<RuntimeValue>,
+	buf: Box<[RuntimeValue]>,
+	/// Index of the first free place in the stack.
+	sp: usize,
 }
 
 impl ValueStack {
 	fn with_limit(limit: usize) -> ValueStack {
+		let mut buf = Vec::new();
+		buf.resize(limit, RuntimeValue::I32(0));
```
Vurich: If we unify the value types as in #99 we can make this use `calloc`

src/runner.rs
Vurich: I feel like that's unnecessary and just leads to having to write many `0`s into the `Box<[RuntimeValue]>`. If we used `Vec` we could have the extra space be transparently lazilly allocated by jemalloc

src/validation/func.rs
Vurich: If this code ever needs to be changed we have to convert it to be a macro. Will tests fail if we mess it up?

src/runner.rs
pepyakin: Oh wow, that's an interesting idea, haven't thought about it. It requires some substantial changes though. 

Firstly, all branches currently are static, i.e. they encode branch destination (and drop keep) as immediates. So we need an instruction that pops a return address and branches to it. (It seems that the only instruction that needs such kind of indirect jump, so why don't keep calling it return? : ) )

Secondly, all code currently contained in a separate vectors. So you can't branch from one function and land in code for another. Although keeping all code in a single vec is more efficient and desired. I didn't go with this way since it might be more error-prone and decided to leave it for later.

Also, when I thought about making drop/keep separate instructions, I decided not to, because large part of interpreters overhead comes from a dispatch. So it seems it is profitable to fuse several instructions into one.

src/runner.rs
pepyakin: AFAIR, using `Vec` in this case actually slows down quite a lot 

> I feel like that's unnecessary

What exactly? Do you mean avoiding using `Vec::push`? 

> just leads to having to write many 0s into the Box<[RuntimeValue]>

But if we use `Vec` we still need to use `resize`, which is basically the same?

src/runner.rs
pepyakin: Yeah, that's the plan! (And because of this I've asked you about zero-ing FPs the other day : ) )

src/validation/func.rs
pepyakin: > If this code ever needs to be changed we have to convert it to be a macro.

Agree. 

> Will tests fail if we mess it up?

I guess so

src/runner.rs
Vurich: When this gets merged I'll write a PR to try to convert it to use a flat vector. We can even generate the bytecode for each individual function in parallel and then flatten it at the end.

You've convinced me that this is fine to keep as a special case, though.

src/runner.rs
Vurich: Why do we need to use `resize` instead of `with_capacity`? Surely one of the validation steps of the bytecode is to make sure that it doesn't read uninitialised stack data.

src/runner.rs
Vurich: Great! I don't know whether it will work with a union but we can do `Vec::<u64>::new()` and then convert that to a `Vec<RuntimeValue>`. That'll definitely use `calloc`.

src/runner.rs
pepyakin: If we use `with_capacity` only and leave `len` unchanged, then we basically can't use `get`, right?

Hm, I think I lost it somewhere and don't follow...

Vurich: I'm thinking something like this:

```rust
#[derive(Debug)]
struct ValueStack {
	buf: Vec<RuntimeValue>,
}

impl ValueStack {
	fn with_limit(limit: usize) -> ValueStack {
		let buf = Vec::with_capacity(limit);

		ValueStack {
			buf: buf,
		}
	}

	#[inline]
	fn drop_keep(&mut self, drop_keep: isa::DropKeep) {
		if drop_keep.keep == isa::Keep::Single {
			let top = *self.top();
			*self.pick_mut(drop_keep.drop as usize + 1) = top;
		}

		let cur_stack_len = self.len();
		self.buf.truncate(cur_stack_len - drop_keep.drop as usize);
	}

	#[inline]
	fn pop_as<T>(&mut self) -> T
	where
		T: FromRuntimeValue,
	{
		let value = self.pop();
		value.try_into().expect("Due to validation stack top's type should match")
	}

	#[inline]
	fn pop_pair_as<T>(&mut self) -> (T, T)
	where
		T: FromRuntimeValue,
	{
		let right = self.pop_as();
		let left = self.pop_as();
		(left, right)
	}

	#[inline]
	fn pop_triple(&mut self) -> (RuntimeValue, RuntimeValue, RuntimeValue) {
		let right = self.pop();
		let mid = self.pop();
		let left = self.pop();
		(left, mid, right)
	}

	#[inline]
	fn top(&self) -> &RuntimeValue {
		self.pick(1)
	}

	fn pick(&self, depth: usize) -> &RuntimeValue {
		&self.buf[self.buf.len() - depth]
	}

	#[inline]
	fn pick_mut(&mut self, depth: usize) -> &mut RuntimeValue {
	    let old_len = self.buf.len();
		&mut self.buf[old_len - depth]
	}

	#[inline]
	fn pop(&mut self) -> RuntimeValue {
	    // TODO: Use `get_unchecked` since we always have at least one value
		self.buf.pop().unwrap()
	}

	#[inline]
	fn push(&mut self, value: RuntimeValue) -> Result<(), TrapKind> {
	    if self.buf.len() == self.buf.capacity() {
	        return Err(TrapKind::StackOverflow);
	    }
	    self.buf.push(value);
		Ok(())
	}

	#[inline]
	fn len(&self) -> usize {
		self.buf.len()
	}
}
```

pepyakin: I just re-run benchmarks with this version and they show that this version is indeed slower (not as slower as I expected though).

benches/Cargo.toml
```diff
@@ -7,3 +7,6 @@ authors = ["Sergey Pepyakin <s.pepyakin@gmail.com>"]
 wasmi = { path = ".." }
 assert_matches = "1.2"
 wabt = "0.3"
+
+[profile.bench]
+debug = true
```
NikVolf: why

Vurich: For cachegrind

pepyakin: Yeah, or Instruments in my case : )

src/runner.rs
```diff
@@ -117,322 +139,274 @@ impl<'a, E: Externals> Interpreter<'a, E> {
...
 	}
 
-	fn run_br_table(&mut self, context: &mut FunctionContext, table: &[u32], default: u32) -> Result<InstructionOutcome, TrapKind> {
-		let index: u32 = context.value_stack_mut()
+	fn run_br_table(&mut self, table: &[isa::Target]) -> Result<InstructionOutcome, TrapKind> {
+		let index: u32 = self.value_stack
 			.pop_as();
-		Ok(InstructionOutcome::Branch(table.get(index as usize).cloned().unwrap_or(default) as usize))
+
+		let dst =
```
NikVolf: strange identation below

src/runner.rs
```diff
@@ -452,8 +426,8 @@ impl<'a, E: Externals> Interpreter<'a, E> {
 		context: &mut FunctionContext,
 		signature_idx: u32,
 	) -> Result<InstructionOutcome, TrapKind> {
-		let table_func_idx: u32 = context
-			.value_stack_mut()
+		let table_func_idx: u32 = self
+			.value_stack
```
NikVolf: new line for this is overkill?

src/runner.rs
```diff
@@ -478,46 +452,46 @@ impl<'a, E: Externals> Interpreter<'a, E> {
 		Ok(InstructionOutcome::ExecuteCall(func_ref))
 	}
 
-	fn run_drop(&mut self, context: &mut FunctionContext) -> Result<InstructionOutcome, TrapKind> {
-		let _ = context
-			.value_stack_mut()
+	fn run_drop(&mut self) -> Result<InstructionOutcome, TrapKind> {
+		let _ = self
+			.value_stack
```
NikVolf: ditto too many lines

src/runner.rs
```diff
@@ -623,418 +597,401 @@ impl<'a, E: Externals> Interpreter<'a, E> {
...
-	fn run_shl<T>(&mut self, context: &mut FunctionContext, mask: T) -> Result<InstructionOutcome, TrapKind>
+	fn run_shl<T>(&mut self, mask: T) -> Result<InstructionOutcome, TrapKind>
 	where RuntimeValue: From<<T as ops::Shl<T>>::Output>, T: ops::Shl<T> + ops::BitAnd<T, Output=T> + FromRuntimeValue {
-		let (left, right) = context
-			.value_stack_mut()
-			.pop_pair_as::<T>()
-			.expect("Due to validation stack should contain pair of values");
+		let (left, right) = self
+			.value_stack
+			.pop_pair_as::<T>();
```
NikVolf: is ::<T> needed here ?

src/validation/func.rs
```diff
@@ -99,397 +223,1071 @@ impl Validator {
...
+			}
+			Return => {
+				if let BlockType::Value(value_type) = context.return_type()? {
+					tee_value(&mut context.value_stack, &context.frame_stack, value_type.into())?;
+				}
+
+				let drop_keep = drop_keep_return(
+					&context.locals,
+					&context.value_stack,
+					&context.frame_stack
```
NikVolf: missing trailing comma ;)

src/runner.rs
pepyakin: In this particular, yes, it is needed. Rustc is having problems with infering the types.

src/runner.rs
Vurich: Maybe we can use `rustfmt` from here on?

pepyakin: I'd like to, but let's postpone it until the next PR!

Vurich:
:tada:

Issue #85, opened 2018-04-18
interpreter pauses


Geal:
hello,

I'm currently developing [serverless-wasm](https://github.com/geal/serverless-wasm) and I'm wondering how I could implement two features:
- putting a hard limit on the run time for a function (real time, or number of opcodes)
- supporting asynchronous networking in host functions. Something like, WASM side calls tcp_connect, host creates the connection, stops the interpreter right there, and resumes once the connection is established

From what I understand, I could proceed like this (for the async part):
- have some host function return a custom trap to say "WouldBlock" and registers that we're waiting for a specific socket
- write another version of `Interpreter::start_execution` that would store the function context when receiving that trap from `Interpreter::run_interpreter_loop` instead of dropping it
- once the code calling the interpreter knows the connection is established (or we got an error), modify the function stack to push the return value
- resume executing from there

It looks like it would be doable, and could be used to implement breakpoints, too.

The other feature, limiting the run time of a function, would probably be easier to implement, with the loop regularly checking that we do not exceed some limit passed as arguments.

Do you think those features would be useful in wasmi?
The serverless-wasm project is in very early stage, and I admit the readme is not too serious, but I really want to get it to a state where it's nice to use, and I need at least the async networking feature to get it.

pepyakin:
This might be something that would be useful for wasmi, but before working on the details, let's discuss other, non-invasive, solutions.

So it sounds to me as you trying to achieve async calls with sync interface. And i'm wondering, have you considered using asynchronous interface? That is, `tcp_connect` returns handle, host creates the connection and calls back to some exported function in wasm to push update with that handle.

As for limiting run time. In blockchains we have a similar problem: we need to limit execution of wasm code by the run time of an average machine that runs the node and do it deterministically. To solve this problem we use a mechanism called `gas metering`. The idea is simple: every opcode has a cost. Before execution of wasm code we instrument it with gas metering calls. At the beggining of each basic block a call to a special function (e.g. `use_gas(i32)`) is inserted. The parameter passed in that function contains a sum of the costs of all following instructions in the basic block.
So I wondering, maybe you can also use instrumentation for that? Maybe you can even reuse our gas metering solution with gas cost = 1 for every instruction.

[...]

PR #110, opened 2018-07-06
Resumable function invocation

sorpaas:
Implements #85. If a trap of kind Host happens, we allow the caller to feed back in another return value for that host function, and then continue the execution.

* For interpreter, the call stack is moved to `Interpreter` struct itself. We drop the interpreter altogether for normal executions anyway. `resume_execution`, which takes the feeded-back return value, will continue the execution with the loop.
* For public interface, `invoke_resumable` is added to `FuncInstance`. This function returns a handle `FuncInvocation`. The caller can then use this to start or resume an execution.

### DIFF
--- a/src/func.rs
+++ b/src/func.rs
@@ -3,8 +3,9 @@ use std::fmt;
 use parity_wasm::elements::Local;
 use {Trap, TrapKind, Signature};
 use host::Externals;
-use runner::{check_function_args, Interpreter};
+use runner::{check_function_args, Interpreter, InterpreterState};
 use value::RuntimeValue;
+use types::ValueType;
 use module::ModuleInstance;
 use isa;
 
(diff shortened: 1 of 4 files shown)


sorpaas:
Added a simple test. And for resumable calls, I moved passing `&mut Externals` from initializing the interpreter to calling `start/resume`. The rationale is that for most of the time, when dealing with resumables, we would mostly want do interpretation, modify externals, do interpretation, modify externals, ... Passing externals on initialization would mean this routine would face lifetime issues.

I think this PR is ready for review now.

src/runner.rs
```diff
@@ -70,14 +87,66 @@ impl<'a, E: Externals> Interpreter<'a, E> {
...
+		})
+	}
+
+	pub fn state(&self) -> &InterpreterState {
+		&self.state
+	}
+
+	pub fn start_execution<'a, E: Externals + 'a>(&mut self, externals: &'a mut E) -> Result<Option<RuntimeValue>, Trap> {
+		// Ensure that the VM has not been executed.
+		assert!(self.state == InterpreterState::Initialized);
```
NikVolf: Not sure about all those asserts

This all supposed to be called in places with zero tolerance to panics

Can we at least write why this assert will always pass?

sorpaas: The public interface will not panic. In `func` mod's `FuncInvocation::start_execution`, we check whether the state is `Initialized` first before calling `Interpreter::start_execution`. But if there's an internal incorrect usage in our code, then panic do happen. Do we prefer to return additional errors here or is this panic okay?

src/runner.rs
```diff
@@ -70,14 +87,66 @@ impl<'a, E: Externals> Interpreter<'a, E> {
...
+
+		self.state = InterpreterState::Started;
+		self.run_interpreter_loop(externals)?;
+
+		let opt_return_value = self.return_type.map(|_vt| {
+			self.value_stack.pop()
+		});
+
+		// Ensure that stack is empty after the execution.
+		assert!(self.value_stack.len() == 0);
```
NikVolf: What about this assert?
How is that guaranteed that there are no more values?

sorpaas: I'm not sure about this either. This assert is added in https://github.com/paritytech/wasmi/pull/98. @pepyakin Do we have guarantees on this or should we change this to a Trap?

pepyakin: sorry my bad. Yeah we do, it’s guaranteed by the validation properties. I should have add a comment

src/func.rs
```diff
@@ -144,15 +145,140 @@ impl FuncInstance {
...
+	}
+}
+
+/// A resumable invocation error.
+#[derive(Debug)]
+pub enum ResumableError {
+	/// Trap happened.
+	Trap(Trap),
+	/// The invocation is not resumable.
+	NotResumable,
```
pepyakin: Could we expand a little bit more on why the invocation can be not resumable?

src/func.rs
```diff
@@ -144,15 +145,140 @@ impl FuncInstance {
...
 			FuncInstanceInternal::Host {
 				ref host_func_index,
 				..
 			} => externals.invoke_index(*host_func_index, args.into()),
 		}
 	}
+
+	/// Invoke the function, get a resumable handle.  This handle can then be used to actually start the execution. If a
+	/// Host trap happens, caller can use `resume_execution` to feed the expected return value back in, and then
+	/// continue the execution.
```
pepyakin: I think it is a good idea to add a note that this is an experimental API. Also it might be a good idea to also warn the user that they might not be able to find anything like this API on other execution engines.

src/runner.rs
NikVolf: Oh, an explanation is ok

PR #112, opened 2018-07-26
Avoid allocations on memory operations

### LINKED ISSUE #101: Optimization: Value handling operations
Related to #99 
Depends on #98

For now, functions like `into_little_endian` do an allocation. This is very unfortunate for the functions that are called upon every access to wasm memory.

At the very least, we can use change definition from

```rust
pub trait LittleEndianConvert where Self: Sized {
    fn into_little_endian(self) -> Vec<u8>;
    fn from_little_endian(buffer: &[u8]) -> Result<Self, Error>;
}
```

to something like this

```rust
pub trait LittleEndianConvert where Self: Sized {
    fn into_little_endian(&self, out: &mut [u8]); // Will panic if `out` of not appropriate size
    fn from_little_endian(buffer: &[u8]) -> Result<Self, Error>;
}
```

this will avoid allocations for `into_little_endian` case.

arkpar:
Fixes #101

### DIFF
--- a/src/memory.rs
+++ b/src/memory.rs
@@ -7,6 +7,7 @@ use std::cell::RefCell;
 use parity_wasm::elements::ResizableLimits;
 use Error;
 use memory_units::{RoundUpTo, Pages, Bytes};
+use value::LittleEndianConvert;
 
 /// Size of a page of [linear memory][`MemoryInstance`] - 64KiB.
 ///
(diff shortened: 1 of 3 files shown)


pepyakin:
```
 name               master-a605175ab.bench ns/iter  ark-allocs-73fbccaa4.bench ns/iter  diff ns/iter   diff %  speedup
 bench_regex_redux  9,525,512                       7,109,347                             -2,416,165  -25.37%   x 1.34
 bench_rev_comp     12,549,005                      8,504,604                             -4,044,401  -32.23%   x 1.48
 bench_tiny_keccak  5,654,512                       4,880,091                               -774,421  -13.70%   x 1.16
 fac_opt            51,511                          50,128                                    -1,383   -2.68%   x 1.03
 fac_recursive      54,433                          53,466                                      -967   -1.78%   x 1.02
 recursive_ok       1,435,432                       1,363,597                                -71,835   -5.00%   x 1.05
 recursive_trap     180,653                         177,061                                   -3,592   -1.99%   x 1.02
```

PR #114, opened 2018-07-30
Bump wabt's version to 0.4 and add two more test cases

guanqun:
It needs some interface changes.

It passes the "cargo test" and "cargo test".


pepyakin:
Awesome!

PR #115, opened 2018-07-31
Allocate memory on demand

arkpar:


NikVolf:
What is the advantage of this?

You still do `Vec::with_capacity` which allocates, no?

src/memory.rs
```diff
@@ -533,12 +554,13 @@ mod tests {
 	#[test]
 	fn zero_copy() {
 		let mem = MemoryInstance::alloc(Pages(1), None).unwrap();
+		mem.set(100, &[0]).expect("memory set should not fail");
 		mem.with_direct_access_mut(|buf| {
-			assert_eq!(buf.len(), 65536);
+			assert_eq!(buf.len(), 101);
```
pepyakin: Oh, actually this can be surprising for the user. So the user must account that the amount of allocated memory may or may not be equal to the length of the buffer. 

Furthermore, the user might want to differentiate between out-of-bounds access (this should trap) and just untouched memory. The complication here is that we can't really tell the difference between uninitialized memory and zeroed memory because all memory is implicitly zeroed. So it is possible that some host function (aka "externalities func") can be called on (fully or partially) untouched  region. 

On top of that, in the worst case the user might not be able to write anything to the `buf`, because the region that is inteded to write just have never been touched and basically the user should fallback to non-direct access, that IMO complicates matters a lot.

pepyakin:
benches looks nice

```
 bench_regex_redux  7,109,347                           5,979,126                              -1,130,221  -15.90%   x 1.19
 bench_rev_comp     8,504,604                           8,291,579                                -213,025   -2.50%   x 1.03
 bench_tiny_keccak  4,880,091                           4,453,374                                -426,717   -8.74%   x 1.10
 fac_opt            50,128                              44,756                                     -5,372  -10.72%   x 1.12
 fac_recursive      53,466                              46,933                                     -6,533  -12.22%   x 1.14
 recursive_ok       1,363,597                           1,278,726                                 -84,871   -6.22%   x 1.07
 recursive_trap     177,061                             158,144                                   -18,917  -10.68%   x 1.12
```

src/memory.rs
arkpar: They can resize the buffer as they need. The percieved allocated size is returned in `current_size` so it is easy enough to to differentiate between out-of-bounds access.

The user should not be using such low level API in the first place. There's a tradeoff between providing a low level API and making efficient abstractions.

pepyakin: >  There's a tradeoff between providing a low level API and making efficient abstractions.

100% indeed users should already use this function with caution since it might lead to panics. But I wonder if there is a better approach to this, e.g. returning wrapper struct that handles such details? 

Anyway, I wouldn't block the PR from merge and I think a plain comment that cautions the user about this special thing would do.

arkpar: I've added a comment and changed `with_direct_access_mut` so that the closure can control allocation.

pepyakin:
Nice!

PR #118, opened 2018-08-02
make the value stack limit and call stack limit as a config parameter

guanqun:
This is rather naive way, I simply expose the InterpreterConfig interface.  Tell me if you ack this design or not. :)

### DIFF
--- a/src/module.rs
+++ b/src/module.rs
@@ -9,6 +9,7 @@ use {Module, Error, Signature, MemoryInstance, RuntimeValue, TableInstance};
 use imports::ImportResolver;
 use global::{GlobalInstance, GlobalRef};
 use func::{FuncRef, FuncBody, FuncInstance};
+use runner::InterpreterConfig;
 use table::TableRef;
 use memory::MemoryRef;
 use host::Externals;
@@ -155,6 +156,7 @@ pub struct ModuleInstance {
 	memories: RefCell<Vec<MemoryRef>>,
 	globals: RefCell<Vec<GlobalRef>>,
 	exports: RefCell<HashMap<String, ExternVal>>,
+	config: InterpreterConfig,
 }
 
 impl ModuleInstance {
@@ -166,6 +168,7 @@ impl ModuleInstance {
 			memories: RefCell::new(Vec::new()),
 			globals: RefCell::new(Vec::new()),
 			exports: RefCell::new(HashMap::new()),
+			config: Default::default(),
 		}
 	}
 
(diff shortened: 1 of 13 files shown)


src/module.rs
```diff
@@ -515,6 +522,7 @@ impl ModuleInstance {
 	pub fn new<'m, I: ImportResolver>(
 		loaded_module: &'m Module,
 		imports: &I,
+		config: &InterpreterConfig
```
NikVolf: maybe leave `new` function as it is now and make another constructor `with_limit` to limit public api changes?

src/module.rs
```diff
@@ -781,7 +792,8 @@ mod tests {
 		);
 		ModuleInstance::new(
 			&module_with_start,
-			&ImportsBuilder::default()
+			&ImportsBuilder::default(),
+			&InterpreterConfig::default()
```
NikVolf: trailing comma missing for this and other cases

pepyakin:
AFAIR, we have a few different things we also want to be configurable. I think the best way forward is to compile down the approximate list and then proceed with general solution. Otherwise, we risk poluting the API and/or making lifes of wasmi users a little bit harder by breaking changes : )

I'll get back with the compiled list later on

src/module.rs
guanqun: I totally agree that changing the `new` function will affect all the callers... Adding a new `with_limits` sounds better, but probably it should be `with_limit_and_externals`, the current API in this struct doesn't quite allow chain API design. Another idea is to use some kind of builder pattern.

src/module.rs
guanqun: Sure, didn't know that. will update.

guanqun:
I'll wait for the compiled list. I should've discussed with you before changing the API interfaces.

pepyakin:
Apart from configuring value and stack limits, and probably other runtime limits there are:

- validation configuration. We already have such an example with `deny_floating_point`. Someone also might want to configure [these](https://github.com/paritytech/wasmi/blob/0409913a26ad16a622f49b7f885fc6415e80f26d/src/validation/func.rs#L12-L15).
- loading configuration (e.g. flag for loading additional data, like debugging names). This should go in `Module`. 

Now, I don't think that we need a solution for configuring all of these in one place. At the very least we most probably want to separate runtime configuration from all others. Just imagine, if we want to run some functions with a different configuration then we need to load and instantiate module again : )

So in general, embedding configuration only in `Module` is ok.  However, the proposed solution unfortunately falls into aforementioned problem: it specifies configuration upon the instantiation, which means that if we wanted to change some values between instantiation we would need to re-instantiate the module. 

Moving the configuration to invoking functions (i.e. all those that takes `Externals` parameter and calls into `FuncInstance`) should work

src/module.rs
pepyakin: Yeah, great idea about builder pattern!

guanqun:
closing this as I don't have enough time working on this... sorry.

PR #122, opened 2018-08-24
Add no_std support

jrakow:
This PR should contain no breaking changes.

* Add a `std` feature, enabled by default and sync it with `parity-wasm/std`
* Remove uses of `std::io`
* Change to a HashMap implementation that works on `alloc`
* Disable `std::error` for `no_std`
* The biggest change is using `core` and `alloc` everywhere and substituting `std` if available.
This is -- as far as I know -- the accepted way to support `no_std` (as mentioned in the [`alloc` RFC](https://github.com/SimonSapin/rfcs/blob/liballoc/text/0000-liballoc.md)).

`no_std` should probably be added to the Travis configuration.
Nom has a working `no_std` Travis: https://github.com/Geal/nom/blob/master/.travis.yml.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,6 +10,11 @@ description = "WebAssembly interpreter"
 keywords = ["wasm", "webassembly", "bytecode", "interpreter"]
 exclude = [ "/res/*", "/tests/*", "/fuzz/*", "/benches/*" ]
 
+[features]
+default = ["std"]
+# Disable for no_std support
+std = []
+
 [dependencies]
 parity-wasm = "0.31"
 byteorder = "1.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,7 @@ default = ["std"]
 std = ["parity-wasm/std"]
 # Enable for no_std support
 # hashmap_core only works on no_std
-no_std = ["hashmap_core"]
+core = ["hashmap_core"]
 
 [dependencies]
 parity-wasm = { version = "0.31", default-features = false }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,4 +27,5 @@ libm = { version = "0.1.2", optional = true }
 
 [dev-dependencies]
 assert_matches = "1.1"
+rand = "0.4.2"
 wabt = "0.4"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,7 +23,6 @@ parity-wasm = { version = "0.31", default-features = false }
 byteorder = { version = "1.0", default-features = false }
 hashmap_core = { version = "0.1.9", optional = true }
 memory_units = "0.3.0"
-nan-preserving-float = "0.1.0"
 libm = { version = "0.1.2", optional = true }
 
 [dev-dependencies]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,6 +18,7 @@ std = ["parity-wasm/std"]
 [dependencies]
 parity-wasm = { version = "0.31", default-features = false }
 byteorder = "1.0"
+hashmap_core = "0.1.9"
 memory_units = "0.3.0"
 nan-preserving-float = "0.1.0"
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,10 +13,10 @@ exclude = [ "/res/*", "/tests/*", "/fuzz/*", "/benches/*" ]
 [features]
 default = ["std"]
 # Disable for no_std support
-std = []
+std = ["parity-wasm/std"]
 
 [dependencies]
-parity-wasm = "0.31"
+parity-wasm = { version = "0.31", default-features = false }
 byteorder = "1.0"
 memory_units = "0.3.0"
 nan-preserving-float = "0.1.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,7 @@ default = ["std"]
 std = ["parity-wasm/std", "byteorder/std"]
 # Enable for no_std support
 # hashmap_core only works on no_std
-core = ["hashmap_core"]
+core = ["hashmap_core", "libm"]
 
 [dependencies]
 parity-wasm = { version = "0.31", default-features = false }
@@ -24,6 +24,7 @@ byteorder = { version = "1.0", default-features = false }
 hashmap_core = { version = "0.1.9", optional = true }
 memory_units = "0.3.0"
 nan-preserving-float = "0.1.0"
+libm = { version = "0.1.2", optional = true }
 
 [dev-dependencies]
 assert_matches = "1.1"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,14 +13,14 @@ exclude = [ "/res/*", "/tests/*", "/fuzz/*", "/benches/*" ]
 [features]
 default = ["std"]
 # Disable for no_std support
-std = ["parity-wasm/std"]
+std = ["parity-wasm/std", "byteorder/std"]
 # Enable for no_std support
 # hashmap_core only works on no_std
 core = ["hashmap_core"]
 
 [dependencies]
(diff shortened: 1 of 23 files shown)


jrakow:
Also had to add a `no_std` feature to disable `hashmap_core` in `no_std` environments.

CryZe:
Features should never / can't ever be negative. So the `no_std` feature is a bug waiting to happen.

jrakow:
How else would you enable `hashmap_core` *only* in `no_std`?
AFAIK, you need a feature to enable a dependency.

CryZe:
Does it not work via a cfg dependency? It may actually not, because cargo has tons of bugs around features and targets.

jrakow:
I was not aware of cfg dependencies, thanks for the tip.
I tried it now, but sadly https://github.com/rust-lang/cargo/issues/2524.

pepyakin:
Hello , thanks for such a great work! I don’t have much time to review because I’m travelling now

@CryZe is right, using negative features is not recommended.

You can add a new feature ‘core’ for this though. Take a look to a cranelift for example https://github.com/CraneStation/cranelift/blob/master/lib/codegen/Cargo.toml

jrakow:
> The "std" feature enables use of libstd. The "core" feature enables use of some minimal std-like replacement libraries. At least one of these two features need to be enabled.

This is just using different names:
With this PR, either `std` or `no_std` has to be enabled.
At Cranelift's, either `std` or `core` has to be enabled.

But sure, I can rename it, if you want me to?

pepyakin:
oh sorry havent had a look into the PR. Yes, that’s should work! I think I would vote for ‘core’ because I think i’ve seen it more often used in such situations

src/global.rs
```diff
@@ -1,5 +1,5 @@
-use std::rc::Rc;
-use std::cell::Cell;
+use alloc::rc::Rc;
+use core::cell::Cell;
```
sunfishcode: One thing we do in Cranelift is to remap `core` to `std` in the top-level lib.rs [here](https://github.com/CraneStation/cranelift/blob/master/lib/codegen/src/lib.rs#L110) so that most files can just use `std::`. This obviously isn't necessary, but since we have some crates that support no_std and some that don't, it's nice to the code more consistent between the two.

fckt:
> Yes, that’s should work! I think I would vote for ‘core’ because I think i’ve seen it more often used in such situations

But in Parity we use feature with name `std` everywhere. It was started by analogy with https://github.com/BurntSushi/byteorder/blob/master/Cargo.toml#L23

jrakow:
@fckt `core` is a negative feature, which has to be enabled iff `no_std` is to be used.
The default-on normal feature is still called `std`.
We have to use an ugly negative feature, because cfg dependencies are buggy:  rust-lang/cargo#2524.

jrakow:
Only the float handling in `value.rs` is left with `std` dependencies.
This relies on `nan-preserving-float` and some bits of the standard library that are not in `core` (for example [`std::f32::fract`](https://doc.rust-lang.org/std/primitive.f32.html#method.fract)).

I have filed a PR for `no_std` support in `nan-preserving-float`: Vurich/nan-preserving-float#1.
With that and [`libm`](https://crates.io/crates/libm), I was able to compile `wasmi` for `no_std`.
`libm` has some panics in debug mode (japaric/libm#4), so we still use `std` when available.

If the PR to `nan-preserving-float` is accepted and released, we're good to go.

Cargo.toml
```diff
@@ -10,9 +10,18 @@ description = "WebAssembly interpreter"
 keywords = ["wasm", "webassembly", "bytecode", "interpreter"]
 exclude = [ "/res/*", "/tests/*", "/fuzz/*", "/benches/*" ]
 
+[features]
+default = ["std"]
+# Disable for no_std support
+std = ["parity-wasm/std", "byteorder/std"]
+# Enable for no_std support
```
NikVolf: are those comments outdated?

src/memory.rs
```diff
@@ -1,9 +1,11 @@
...
-use std::cmp;
-use std::fmt;
-use std::rc::Rc;
-use std::cell::{Cell, RefCell};
+#[allow(unused_imports)]
+use alloc::prelude::*;
+use alloc::rc::Rc;
+use core::u32;
+use core::ops::Range;
+use core::cmp;
```
NikVolf: core imports might be grouped

Cargo.toml
jrakow: No. They match the readme section about `no_std`.

sorpaas:
I'm trying to get this branch working with a real embedded environment (Rux microkernel!), but so far, it still needs some changes, and I would hope we can fix them!

First, `hashmap_core` crate doesn't work well in my environment. I hope we can remove that crate, but instead, use something like this:

```
#[cfg(feature = "std")]
use std::collections::{HashMap as Map};
#[cfg(not(feature = "std"))]
use alloc::btree_map::{BTreeMap as Map};
```

`BTreeMap`/`BTreeSet` are actually sufficient for all our use-cases, and using hashmap would impose unnecessary additional memory consumption for those embedded environments.

The second issue I'm facing is missing `fract` functions for floating points. That doesn't seem to always be defined in some environments, but we can do something like this in `impl_float!` in `value.rs`:

```
			#[cfg(not(feature = "std"))]
			fn fract(self) -> $type {
				self - self.trunc()
			}
```

I have a branch that's working for Rux microkernel's demo propose at https://github.com/sorpaas/wasmi/tree/jrakow-nostd. That branch also contains some further modifications that make it work with older Rust nightly versions, which I guess we don't need here.

jrakow:
I had actually solved the float issues locally by depending on `libm` for the implementation of `fract`.
These commits are now published here.

The remaining blocker is Vurich/nan-preserving-float#1, which is inactive.
I am not sure how to proceed without waiting on that PR.

sorpaas:
@jrakow Would you mind also fixing the `hashmap_core` issue described above? We can safely use `BTreeMap` for `no_std` environments.

jrakow:
@sorpaas I was just writing a response to that as a second comment:

As for `BTreeMap` vs. `hashmap_core`, I am fine with using `BTreeMap`.
If you would open a PR against https://github.com/jrakow/wasmi/tree/no_std with your changes on that issue, I would happily merge that.

jrakow:
BTW, the Rust issue for float operations in libcore is rust-lang/rfcs#2505.
Using a Rust `libm` is the second solution highlighted there.

sorpaas:
@jrakow Thanks! I still need to test whether your solutions work for me regarding `libm`. The issue is that I need to compile to targets that both support and does not support native floating point. I haven't tested whether `libm` works in all those cases. Even if it does, I think we would miss some opportunities to use native floating point ops to make code run faster -- All we're missing in `libcore` is the one single `fract` function.

Another situation is that it looks like current floating point implementations actually depend on some extern functions (which I have not yet understood why, see [this](https://github.com/sorpaas/wasmi-on-rux/blob/master/sinits/examples/wasmi.rs)).

One of the possible solutions is to add a feature flag to disable all floating points operations from compiling -- we don't need them in deterministic environments like blockchain virtual machines, so it might be a good feature to have.

Anyway, the above looks not trivial so I might just get another PR if I figured out something (and will definitely fix `BTreeMap` issue, but probably also just in that PR, to make pulling logistics simple!). Please feel free to continue and merge this!

pepyakin:
After chatting with @Vurich in private we decided to embed `nan-preserving-float` into `wasmi` repo.

jrakow:
Looks like the Travis cache is getting too big.

pepyakin:
Finally we merged it! Thanks for the great job, @jrakow !

src/global.rs
Vurich: This would have made rebasing a lot nicer, too :upside_down_face:

PR #123, opened 2018-08-26
Proofread the library-level docs

tbu-:


src/lib.rs
```diff
@@ -1,48 +1,48 @@
 //! # wasmi
 //!
-//! This library allows to load WebAssembly modules in binary format and invoke functions on them.
+//! This library allows to load WebAssembly modules in binary format and invoke their functions them.
```
NikVolf: `and invoke their functions them?`

is this ok english?

src/lib.rs
```diff
@@ -1,48 +1,48 @@
...
-//! WebAssembly (wasm) is a safe, portable, compact format that designed for efficient execution.
+//! WebAssembly (wasm) is a safe, portable and compact format that designed for efficient execution.
 //!
-//! Wasm code is distributed in a form of modules, that contains definitions of:
+//! Wasm code is distributed in the form of modules that contains definitions of:
 //!
 //! - functions,
 //! - global variables,
-//! - linear memories,
+//! - linear memory spaces and
```
NikVolf: "memory instances" will fit probably

we don't use "memory space" anywhere I think

PR #124, opened 2018-08-29
Expose the `validate_module` function and `ValidatedModule` to crate users

pchickey:
This just makes the function and struct public in `lib.rs`, so that they can be used outside of the crate.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -380,6 +380,7 @@ pub use self::module::{ModuleInstance, ModuleRef, ExternVal, NotStartedModuleRef
 pub use self::global::{GlobalInstance, GlobalRef};
 pub use self::func::{FuncInstance, FuncRef, FuncInvocation, ResumableError};
 pub use self::types::{Signature, ValueType, GlobalDescriptor, TableDescriptor, MemoryDescriptor};
+pub use self::validation::{ValidatedModule, validate_module};
 
 /// WebAssembly-specific sizes and units.
 pub mod memory_units {
--- a/src/validation/mod.rs
+++ b/src/validation/mod.rs
@@ -40,8 +40,11 @@ impl From<stack::Error> for Error {
 }
 
 #[derive(Clone)]
+/// A deserialized and validated module
 pub struct ValidatedModule {
+	/// The instructions in each Function, represented using isa::Instructions
 	pub code_map: Vec<isa::Instructions>,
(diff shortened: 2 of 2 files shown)


src/lib.rs
```diff
@@ -380,6 +380,7 @@ pub use self::module::{ModuleInstance, ModuleRef, ExternVal, NotStartedModuleRef
 pub use self::global::{GlobalInstance, GlobalRef};
 pub use self::func::{FuncInstance, FuncRef, FuncInvocation, ResumableError};
 pub use self::types::{Signature, ValueType, GlobalDescriptor, TableDescriptor, MemoryDescriptor};
+pub use self::validation::{ValidatedModule, validate_module};
```
pepyakin: This will expose `ValidatedModule` which in turn will ultimately expose `isa::Instruction`, which is the implementation detail.

pepyakin:
I have mixed feelings about this.

On the one hand, this module is called validation but it is more about "preparation" or something like that. It validates the code and generates internal representation along the way, and all that touches the internal representation is an implementation detail.

On the other hand, I have a feeling that validation logic is too coupled with the compilation, and I'd rather decouple them. If we do so, then this will open up the way for the usage of this crate in other projects, which sounds good to me. The thing is that it is not the simplest thing to do. 

So, if possible I would prefer to do the proper extraction. If not, then we probably might go with some feature gate that exports these items. This however will not give you any guarantees about stability of these items.

pchickey:
Thanks for the feedback! I sympathize with your concerns. For my use case, I didn't care about the `isa::Instruction` detail.

I'll scope out what it takes to do a proper extraction. The other implementation I've looked at is https://github.com/yurydelendik/wasmparser.rs/blob/master/src/validator.rs, but it has a very different architecture because it is designed for streaming. I'll be off on vacation next week so it may be a little bit before I get back to this.

pchickey:
I'm not sure of the best forum for this discussion, so please feel free to redirect -

I copied the validation code from this repo and removed all of the dependencies on the `isa` and other code generation details. I tried to make as few change as possible at first - it may be possible to remove a bunch of the logic related to labels and control flow, but for the moment I left a bunch of it in.

https://github.com/pchickey/pwasm-validator

It passes the tests that were in this repo (where appropriate, again trimming out any isa-related details) and the official spec suite.

PR #126, opened 2018-09-17
implement extend() and remove one TODO

guanqun:
Notice that this branch has been there for some time, sending a PR now...

### DIFF
--- a/src/runner.rs
+++ b/src/runner.rs
@@ -1,6 +1,5 @@
 use std::ops;
 use std::{u32, usize};
-use std::cmp;
 use std::fmt;
 use std::iter::repeat;
 use parity_wasm::elements::Local;
@@ -1262,13 +1261,15 @@ impl ValueStack {
 
 	#[inline]
 	fn extend(&mut self, values: &[RuntimeValue]) -> Result<(), TrapKind> {
-		let len = cmp::min(values.len(), self.buf.len() - self.sp);
-		self.buf[self.sp..(self.sp+len)].copy_from_slice(&values[..len]);
-		self.sp += len;
-		if len < values.len() {
-			Err(TrapKind::StackOverflow)
+		let avail_len = self.buf.len() - self.sp;
+		let to_copy_len = values.len();
+
+		if to_copy_len <= avail_len {
+		    self.buf[self.sp..(self.sp+to_copy_len)].copy_from_slice(&values[..to_copy_len]);
+		    self.sp += to_copy_len;
+		    Ok(())
 		} else {
-			Ok(())
+		    Err(TrapKind::StackOverflow)
 		}
 	}
 
@@ -1286,8 +1287,13 @@ mod tests {
 	#[test]
 	fn test_stack_extend() {
 		let mut stack = ValueStack::with_limit(5);
-		let values = vec![RuntimeValue::I32(1), RuntimeValue::I32(2), RuntimeValue::I32(3),
-			RuntimeValue::I32(4), RuntimeValue::I32(5)];
+		let values = vec![
+			RuntimeValue::I32(1),
+			RuntimeValue::I32(2),
+			RuntimeValue::I32(3),
+			RuntimeValue::I32(4),
+			RuntimeValue::I32(5)
+		];
 
 		assert!(stack.extend(&values).is_ok());
 		assert_eq!(stack.len(), 5);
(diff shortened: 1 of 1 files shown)


src/runner.rs
```diff
@@ -1263,8 +1260,47 @@ impl ValueStack {
 		Ok(())
 	}
 
+	#[inline]
+	fn extend(&mut self, values: &[RuntimeValue]) -> Result<(), TrapKind> {
+		let len = cmp::min(values.len(), self.buf.len() - self.sp);
+		self.buf[self.sp..(self.sp+len)].copy_from_slice(&values[..len]);
+		self.sp += len;
+		if len < values.len() {
+			Err(TrapKind::StackOverflow)
```
pepyakin: I wonder why do we return `StackOverflow` here, after we've tried to perform copy (even though it is copy of zero elements) and we've modified the stack pointer?

It's actually violates consistency of the stack. Consider this test:


```rust
let values = vec![
    RuntimeValue::I32(1), 
    RuntimeValue::I32(2), 
    RuntimeValue::I32(3),
    RuntimeValue::I32(4), 
    RuntimeValue::I32(5),
];

let mut stack = ValueStack::with_limit(4);
assert!(stack.extend(&values).is_err());
assert!(stack.push(RuntimeValue::I32(0)).is_ok());
```

this test is going to fail since `sp` will be modified and will point out of the underlying buffer.

src/runner.rs
```diff
@@ -1263,8 +1260,47 @@ impl ValueStack {
 		Ok(())
 	}
 
+	#[inline]
+	fn extend(&mut self, values: &[RuntimeValue]) -> Result<(), TrapKind> {
+		let len = cmp::min(values.len(), self.buf.len() - self.sp);
```
pepyakin: Consider giving a name to `self.buf.len() - self.sp` expression

src/runner.rs
```diff
@@ -1109,11 +1110,7 @@ impl FunctionContext {
 			.map(RuntimeValue::default)
 			.collect::<Vec<_>>();
 
-		// TODO: Replace with extend.
-		for local in locals {
-			value_stack.push(local)
-				.map_err(|_| TrapKind::StackOverflow)?;
-		}
+		value_stack.extend(&locals)?;
```
pepyakin: I wonder if this TODO is actually [still] relevent. Have you performed benchmarks before/after?

src/runner.rs
guanqun: I think we need a bit clarification on the semantics of the api, I was trying to follow the previous code's semantics:

```
               for local in locals {
                       value_stack.push(local)
                               .map_err(|_| TrapKind::StackOverflow)?;
               }
```

This code will insert `locals' until it overflows, when it overflows, all previous entries have been pushed into value_stack already. That's why I do a copy first and then if we figure out we're not able to copy all, then simply return a 'overflow' error.

src/runner.rs
guanqun: sure.

src/runner.rs
guanqun: Not  yet, I assume the `copy_from_slice` can do something like `memmove`, and it would be better than pushing values one by one.

pepyakin: Yeah, agree with your logic here. But at the same time, I think there is (a rather slight) chance that LLVM could recognize pattern and turn it into `mem{cpy,move}`. If that's the case, then I would rather keep the code simple.

guanqun:
Fixed the implementation of `extend()`, but I doubt the speedup will reflect in the current benchmark, as this is not a critical path. I probably need to write a specific benchmark around this interface.

pepyakin:
```
 name               master-438eab9ad.bench ns/iter  remove-one-todo-039c75589.bench ns/iter  diff ns/iter   diff %  speedup
 bench_regex_redux  7,061,385                       7,051,162                                     -10,223   -0.14%   x 1.00
 bench_rev_comp     9,723,503                       9,236,616                                    -486,887   -5.01%   x 1.05
 bench_tiny_keccak  5,141,418                       5,225,626                                      84,208    1.64%   x 0.98
 fac_opt            53,578                          61,003                                          7,425   13.86%   x 0.88
 fac_recursive      55,997                          57,160                                          1,163    2.08%   x 0.98
 recursive_ok       1,464,346                       1,291,170                                    -173,176  -11.83%   x 1.13
 recursive_trap     199,255                         174,759                                       -24,496  -12.29%   x 1.14
```

hm, there is a slight regression in some of the benchmarks.

guanqun:
Thanks for testing this out.  Interesting...

guanqun:
closing this as I don't have enough time working on this... sorry.

pepyakin:
No problem, thank you so much for your PRs.

PR #128, opened 2018-10-01
Transfer function

pepyakin:
Add a `transfer` function to `MemoryInstance`, which allows copying memory between two, possibly distinct, memory instances.

### DIFF
--- a/src/memory.rs
+++ b/src/memory.rs
@@ -352,6 +352,29 @@ impl MemoryInstance {
 		Ok(())
 	}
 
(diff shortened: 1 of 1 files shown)


pepyakin:
Yeah, sure. This use-case is pretty niche for the moment because every module instance can have at most one memory instance. This function can become handy if we were working with several module instances at the same time and these instances communicate through memory but without sharing one memory instance. And indeed it happened that in substrate we have such a requirement: in sandboxing we need to copy memory between supervisor and guest module instances and do it without using intermediate buffer.

Issue #100, opened 2018-06-20
Optimization: Use compact encoding of bytecode


pepyakin:
Depends on #98

For now, every instruction is represented by an enum value. It means that each instruction occupies size required for a tag plus a size of the instruction that has the largest payload (probably `BrTable`) (which requires to be properly aligned). Now each instruction occupies about 24 bytes, even though most instructions are 1 byte wide.

Fortunately, we don't need to address instructions by an index and can iterate them sequentially, and branches can specify the exact byte offset for their target.

Shrinking the size of one instruction will allow us to use cache more efficiently and also will reduce memory overhead greatly.

willglynn:
I got this working with a crude encoding scheme. Each `isa::Instruction` is identified with a single `u8`, followed by any operands (usually `u32`). As of now, I'm seeing a 10-15% performance improvement in `cargo bench` results.

The savings aren't as large as I had hoped. A napkin-math analysis over the test suite gives me a thought:

<img width="740" alt="Bytes per instruction" src="https://user-images.githubusercontent.com/118362/46592285-c8e9e400-ca86-11e8-8b8c-b64202a59550.png">

Most of the `u32` operands are slots or offsets or similar small integers. Maybe a variable length encoding scheme is worth a go?

pepyakin:
10-15% is a very good figure actually!

This is basically a trade-off and we can easily get more overhead than we save by compacting the bytecode. What we might want to explore is introduction of a special narrow versions of opcodes that takes, say, u16 instead of u32 immediates.

willglynn:
Well… then I started thinking about how WebAssembly is already a compact bytecode format with variable-length encoding of operands. `FunctionReader::read_instruction()` swizzles around flow control, but many more instructions are [validated then emitted 1:1](https://github.com/paritytech/wasmi/blob/master/src/validation/func.rs#L506-L1126). Maybe it's better not to produce a second bytecode, but to instead produce a mix of new/enriched flow control instructions along with references to the original bytecode.

pepyakin:
> Maybe it's better not to produce a second bytecode, but to instead produce a mix of new/enriched flow control instructions along with references to the original bytecode.

Can you expand a little on this? I'm not sure I follow.

pepyakin:
Related:

- https://github.com/paritytech/parity-wasm/issues/135
- https://github.com/paritytech/parity-wasm/pull/154
- https://github.com/paritytech/parity-wasm/issues/224
[...]

PR #129, opened 2018-10-07
Hide instruction storage details

willglynn:
This PR lays some groundwork for #100.

`isa::Instructions` is now an opaque container. `current_pc(&self) -> u32` indicates the position at which the next instruction would be added, and `push(&mut self, Instruction)` actually does the adding. Currently this happens by pushing into a `Vec<Instruction>`, but this could instead encode the `Instruction` into bytecode.

`isa::Instructions` supports reading back instructions via `iterate_from(&self, pc: u32) -> InstructionIter<'>`. Again, right now that just walks the internal `Vec<Instruction>`, but it could instead sequentially decode `Instruction`s from bytecode.

Since instruction encoding is now internal to `isa::Instructions`, it also gains `patch_relocation(&mut self, Reloc, dst_pc: u32)`, and that enumeration moves to `isa::Reloc`. `func::Sink::resolve_label()` still controls the process as usual, but since the `isa::Instruction` is now hidden inside `isa::Instructions`, `isa::Instructions` needs to perform the required patching.

Finally, I noticed a few places where `usize` is used as a program counter. I changed those to `u32`, since a) that seems like enough and b) `isa::Instruction` is `u32` already.

### DIFF
--- a/src/runner.rs
+++ b/src/runner.rs
@@ -175,7 +175,7 @@ impl Interpreter {
 			let function_return =
 				self.do_run_function(
 					&mut function_context,
-					&function_body.code.code,
+					&function_body.code,
 				).map_err(Trap::new)?;
 
 			match function_return {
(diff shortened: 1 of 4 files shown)


willglynn:
Sorry for the noise. The original Travis build failed due to [a transient issue](https://travis-ci.org/paritytech/wasmi/jobs/438341674#L961). Restarting builds requires write access to the repo, which I don't have – however, opening a PR triggers a new build, and closing/reopening was within my power. That second build [is green](https://travis-ci.org/paritytech/wasmi/builds/438351385).

willglynn:
I pushed another commit into this PR, one which lets tests continue working once `Target{ pc: }` refers to byte location rather than index.

This was the last change I needed in order to actually implement a bytecode `Instruction` encoding – an encoding which works but which causes large performance regressions, and so is staying out of this PR :-)

src/runner.rs
```diff
@@ -229,18 +229,24 @@ impl Interpreter {
...
 	}
 
-	fn do_run_function(&mut self, function_context: &mut FunctionContext, instructions: &[isa::Instruction]) -> Result<RunResult, TrapKind> {
+	fn do_run_function(&mut self, function_context: &mut FunctionContext, instructions: &isa::Instructions)
+		-> Result<RunResult, TrapKind>
+	{
+		let mut iter = instructions.iterate_from(function_context.position);
 		loop {
-			let instruction = &instructions[function_context.position];
+			let instruction = iter.next().expect("instruction");
```
pepyakin: Can you expand this a little bit, giving a clue for the reader what went wrong?

PR #132, opened 2018-10-26
Travis maintenance

pepyakin:
- use gcc-8
- update wabt to 0.6.
- don't run cargo-deadlinks in CI. It keeps getting us problems, e.g. [this](https://travis-ci.org/paritytech/wasmi/jobs/444458704#L1311-L1313). It's time to get rid of it.


Vurich:
LGTM :+1:

PR #133, opened 2018-10-26
Remove tag from RuntimeValue

### LINKED ISSUE #99: Optimization: use unions for representing RuntimeValue
Depends on #98

Currently, a `RuntimeValue` is represented by a rust enum. 

As a refresher: A rust enum requires space for the variant tag plus the size of a payload of the largest variant, which also requires to be properly aligned. Thus, a `RuntimeValue` takes 8 bytes for the payload (for 64-bit wide values) and 8 bytes for the alignment (on x86_64).

We can shelve 8 bytes by removing a tag. We can achieve it by using C-like unions for representing runtime values internally. It is possible because after validation it is statically guaranteed that each operation will be used with operands of appropriate types (i.e., `f32.*` operators will always be used with `f32` operands).

Shrinking `RuntimeValue` may potentially improve cache efficiency for the operand stack and remove branching to check operand types.

Vurich:
This closes #99. Benchcmp results:

```
 name               preopt.bench ns/iter  postopt.bench ns/iter  diff ns/iter   diff %  speedup 
 bench_regex_redux  4,471,532             4,117,821                  -353,711   -7.91%   x 1.09 
 bench_rev_comp     10,363,517            9,552,537                  -810,980   -7.83%   x 1.08 
 bench_tiny_keccak  5,475,809             5,044,934                  -430,875   -7.87%   x 1.09 
 fac_opt            42,472                26,942                      -15,530  -36.57%   x 1.58 
 fac_recursive      43,852                28,770                      -15,082  -34.39%   x 1.52 
 recursive_ok       1,423,602             1,363,951                   -59,651   -4.19%   x 1.04 
 recursive_trap     182,536               160,872                     -21,664  -11.87%   x 1.13 
```

All tests still pass. I expect most of these improvements are due to the fact that we double our cache efficiency by halving the size of `RuntimeValue` (since `RuntimeValue` includes a `u64`, its tag must be padded to 64 bits, making the whole value 2x the size of a u64). However, some improvements may be due to the fact that we remove some validity checking on function arguments. This was duplicated effort anyway, since that is already checked statically.

### DIFF
--- a/src/runner.rs
+++ b/src/runner.rs
@@ -1,30 +1,126 @@
 #[allow(unused_imports)]
 use alloc::prelude::*;
+use common::{DEFAULT_MEMORY_INDEX, DEFAULT_TABLE_INDEX};
+use core::fmt;
 use core::ops;
 use core::{u32, usize};
-use core::fmt;
-use core::iter::repeat;
-use parity_wasm::elements::Local;
-use {Trap, TrapKind, Signature};
-use module::ModuleRef;
-use memory::MemoryRef;
-use func::{FuncRef, FuncInstance, FuncInstanceInternal};
-use value::{
-	ArithmeticOps, ExtendInto, Float, FromRuntimeValueInternal, Integer, LittleEndianConvert,
-	RuntimeValue, RuntimeValueInternal, TransmuteInto, TryTruncateInto, WrapInto,
-};
+use func::{FuncInstance, FuncInstanceInternal, FuncRef};
 use host::Externals;
-use common::{DEFAULT_MEMORY_INDEX, DEFAULT_TABLE_INDEX};
-use types::ValueType;
-use memory_units::Pages;
-use nan_preserving_float::{F32, F64};
 use isa;
+use memory::MemoryRef;
+use memory_units::Pages;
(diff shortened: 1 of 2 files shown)


pepyakin:
Could you rebase this please?

Vurich:
Per mine and @pepyakin's discussion in private, we seem to get such a huge improvement on the factorial benchmarks because zeroing allocation can use `memset`/`calloc` now instead of having to write bytes individually. Before it was setting stack values to `I32(0)`, which (seemingly) doesn't have the tag `0` and so `memset(0)` wouldn't work, but now because it's essentially just writing a zeroed u64 Rust can generate a `memset` or `calloc` (depending on whether we're on the stack or the heap, respectively).

pepyakin:
Could you rebase this?

Vurich:
@pepyakin After this rebase can you just squash+merge (unless it's incorrect of course) because I have to manually merge every time because all our changes touch `runner.rs`

pepyakin:
yeah sure! Just ping me in every possible way you can whenever you've done with rebasing : )

pepyakin:
Perfect, thank you!

PR #135, opened 2018-10-29
Remove redundent check_function_args

sorpaas:
In `FuncInstance::invoke`, `check_function_args` is immediately called again.

### DIFF
--- a/src/module.rs
+++ b/src/module.rs
@@ -623,7 +623,6 @@ impl ModuleInstance {
 			}
 		};
 
-		check_function_args(func_instance.signature(), &args)?;
 		FuncInstance::invoke(&func_instance, args, externals)
 			.map_err(|t| Error::Trap(t))
 	}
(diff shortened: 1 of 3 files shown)


src/module.rs
```diff
@@ -623,7 +623,6 @@ impl ModuleInstance {
 			}
 		};
 
-		check_function_args(func_instance.signature(), &args)?;
```
pepyakin: In this case, there is not a lot of sense in actually formatting error message inside of `check_function_args`. Maybe we can get rid of it?

pepyakin:
As potential follow-up: Why do we use collect there? Does this involve allocation of a temporary vector?

sorpaas:
@pepyakin Yeah I think that's unnecessary. Removed!

Vurich:
You can still use collect and collect into a `Result<(), _>` instead of a `Result<Vec<_>, _>`, although in this case there's no need to. `()` implements `FromIterator<Item = ()>`.

Issue #139, opened 2018-11-06
Remove `Box<[Target]>` from `Instruction`


Vurich:
Currently copying a `Vec<Instruction>` is very expensive. Although most variants are `Copy`-able, `BrTable` contains a `Box<[Target]>`, which means that copying an instruction always requires checking the variant and you can only copy one at a time in serial. We should instead use an encoding like so:

```rust
// Note the addition of `Copy`
#[derive(Debug, Copy, Clone, PartialEq)]
enum Instruction {
    // ...
    BrTable { count: u32 },
    BrTableTarget(Target),
    // ...
}
```

This would mean that we can `memcpy` a vector of `Instruction`s, _much_ faster. It also reduces pointer chasing, see #136.

Finally, this is good preparation work for #100, since it means that decoding doesn't mean allocating a `Box<[Target]>`.

PR #141, opened 2018-11-12
Remove `Box<[Target]>` from `Instruction`

Vurich:
Closes https://github.com/paritytech/wasmi/issues/139. This also allows `Instruction` to be `Copy`, which massively speeds
up `<Instructions as Clone>::clone` since it can now just `memcpy`
the bytes using SIMD instead of having to switch on every single
element. I haven't looked at the disassembly of `InstructionIter::next`
yet, it could be that there are even more improvements yet to be gained
from either:

* Only doing work on `BrTable` (this might already be the case depending
  on the whims of the optimiser),
* Using `unsafe` to make it a noop (we really don't want to do this,
  obviously, since it means that `Instructions` has to be immovable).

This also paves the way for an implementation of #100 (in theory we shouldn't have to do any API breakage to implement that feature), although one wonders if it will actually be a win.

`cargo benchcmp` results:

```
 name               box.bench ns/iter  nobox.bench ns/iter            diff ns/iter   diff %  speedup 
 bench_regex_redux  4,644,263          4,050,022                          -594,241  -12.80%   x 1.15 
 bench_rev_comp     10,337,697         9,149,109                        -1,188,588  -11.50%   x 1.13 
 bench_tiny_keccak  5,389,921          4,824,605                          -565,316  -10.49%   x 1.12 
 fac_opt            15,230             15,110                                 -120   -0.79%   x 1.01 
 fac_recursive      17,188             16,773                                 -415   -2.41%   x 1.02 
 recursive_ok       1,351,895          1,200,655                          -151,240  -11.19%   x 1.13 
 recursive_trap     142,108            133,746                              -8,362   -5.88%   x 1.06 
```

WARNING: This is a breaking change to `isa::Instruction` and so we should bump the major version once this lands.

### DIFF
--- a/src/isa.rs
+++ b/src/isa.rs
@@ -74,7 +74,6 @@ use alloc::prelude::*;
 ///
 /// Note that this is a `enum` since Wasm doesn't support multiple return
 /// values at the moment.
-#[repr(u8)]
 #[derive(Copy, Clone, Debug, PartialEq, Eq)]
 pub enum Keep {
 	None,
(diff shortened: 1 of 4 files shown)


src/isa.rs
```diff
@@ -332,35 +537,35 @@ impl Instructions {
 		self.vec.len() as u32
 	}
 
-	pub fn push(&mut self, instruction: Instruction) {
+	pub(crate) fn push(&mut self, instruction: InstructionInternal) {
```
Vurich: TODO: We might want to make it so you supply a `Instruction` and we convert it to an `InstructionInternal`, then have a `pub(crate) fn push_internal`

src/isa.rs
```diff
@@ -90,30 +89,49 @@ pub struct DropKeep {
...
 	/// the specified pc.
-	BrTable {
-		pc: u32,
-		idx: usize,
-	},
+	BrTable { pc: u32, idx: usize },
+}
+
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub struct Targets<'a> {
```
pepyakin: Since this is only for `br_table` and will not be used for other cases in the foreseen future, maybe it would be better  to rename this to `BrTargets`?

src/runner.rs
```diff
@@ -338,10 +338,11 @@ impl Interpreter {
 		instructions: &isa::Instructions,
 	) -> Result<RunResult, TrapKind> {
 		let mut iter = instructions.iterate_from(function_context.position);
+
 		loop {
-			let instruction = iter.next().expect("instruction");
+			let instruction = iter.next().expect("Function ended without explicit return");
```
pepyakin: Yeah it's somewhat more complicated: it's not only an explicit return that can return from a function but also an implicit `end`. Moreover, function execution can also be terminated by a trap.

How about "Due to validation: it's impossible to reach the end of function."?

PR #142, opened 2018-11-14
Use rustfmt

bddap:
This pr adds a `rustfmt.toml` and formats all existing code.


bkchr:
Hey, thanks for your pr. However, we do not want to use `rustfmt` as it does not support our [style guidelines](https://wiki.parity.io/Substrate-Style-Guide).

Issue #51, opened 2018-02-12
Reconsider the limits


pepyakin:
We need to reconsider the limits (particularly maximal value and frame stack height).
Ideally we should provide a way for a user to change this limits.

Also, we might want to synchronize with limits.h

bddap:
`Interpreter::new` creates new value stack with size `DEFAULT_VALUE_STACK_LIMIT`.

`Interpreter::run_interpreter_loop` checks it's frame stack against `DEFAULT_CALL_STACK_LIMIT`.

I think the first step should be `Interpreter::new` taking value stack limit and frame stack limit as arguments.

bddap:
Ok, backing out of the weeds here.

On a higher level, It seems like `ModuleInstance` should hold value and frame stack limits. That way the user can optionally set those values in when they create the module.

pepyakin:
I would disagree here.

1. A user may want to execute two functions on a same instance with two different limits.
2. One instance is able to call functions on other instance. This raises a question, when calling an instance with different limits should you preserve limits from the calling instance or should you just adjust to a new limits?

In my understanding, stack limits are properties of an execution, not of a (module) instance.

bddap:
Looks like someone (@pepyakin according to git blame) had a similar idea when they wrote `common/stack.rs`. `StackWithLimit` is almost the same thing as `BoundedStack`. The main difference is that `BoundedStack` doesn't need its contents to be `Clone` because dummy values aren't used for pre-allocation.

Since we don't need dummy values, `BoundedStack` can contain `FunctionContext` and can be used as a call stack.

I'm going to merge `BoundedStack` into `StackWithLimit` then submit a pr.

pepyakin:
> I've been fiddling with runner.rs and I think it would make most sense to write a stack struct with upper bound checking.

Can you expand a little bit on this? Isn't the current implementation already has upper bounds?

bddap:
Yes, the `ValueStack` struct has upper bounds checking, but the interpreter call stack is currently a `Vec<_>` so it has to be checked manually.

pepyakin:
Ah, I've looked to your implementation and it's looks like the previous implementation of stack with limit, yes : )

The new stack was introduced for optimization in an attempt to remove unnecessary checks that `Vec` does, e.g. to detect the condition for growing. Preallocation up to the specified upper bound allowed us to get rid of that check, AFAIR.

pepyakin:
[...]

PR #146, opened 2018-11-20
Common bounded stack

bddap:
This pr addresses #51. The goal of this pr is to make stack limits configurable without sacrificing performance.
A useful side effect of these changes is that value and frame stacks become reuseable.

Benches:

| name              | control.trace ns/iter | variable.trace ns/iter | diff ns/iter | diff %  | speedup |
| -                 | -                     | -                      | -            | -       | -       |
| bench_regex_redux | 2,791,066             | 2,812,252              | 21,186       | 0.76%   | x 0.99  |
| bench_rev_comp    | 6,600,544             | 6,651,448              | 50,904       | 0.77%   | x 0.99  |
| bench_tiny_keccak | 3,512,802             | 3,434,810              | -77,992      | -2.22%  | x 1.02  |
| fac_opt           | 19,906                | 11,045                 | -8,861       | -44.51% | x 1.80  |
| fac_recursive     | 22,074                | 13,603                 | -8,471       | -38.38% | x 1.62  |
| recursive_ok      | 1,018,325             | 1,173,291              | 154,966      | 15.22%  | x 0.87  |
| recursive_trap    | 114,942               | 125,829                | 10,887       | 9.47%   | x 0.91  |

I changed `Interpreter` to hold a value and call stack. It no longer remembers the function it is running or the arguments it was passed.

This makes it possible to reuse an interpreter object, (with value and call stack) for subseqeunt function invocations. These are the results of a benchmark with interperters being recycled:

| name              | reuse-control.trace ns/iter | variable.trace ns/iter | diff ns/iter | diff %  | speedup |
| -                 | -                           | -                      | -            | -       | -       |
| bench_regex_redux | 2,791,066                   | 2,781,772              | -9,294       | -0.33%  | x 1.00  |
| bench_rev_comp    | 6,600,544                   | 6,584,948              | -15,596      | -0.24%  | x 1.00  |
| bench_tiny_keccak | 3,512,802                   | 3,432,738              | -80,064      | -2.28%  | x 1.02  |
| fac_opt           | 19,906                      | 1,860                  | -18,046      | -90.66% | x 10.70 |
| fac_recursive     | 22,074                      | 4,194                  | -17,880      | -81.00% | x 5.26  |
| recursive_ok      | 1,018,325                   | 1,129,373              | 111,048      | 10.90%  | x 0.90  |
| recursive_trap    | 114,942                     | 111,131                | -3,811       | -3.32%  | x 1.03  |

Other changes:

- public typesafe api for configuring frame and value stacks (see `StackSize` common/stack.rs)
- made `StackWithLimit` pre-allocate and dynamically resize when appropriate
- optimized `StackWithLimit` implementation (push and pop_unchecked are speedy now). As @pepykin found, Vec::push and Vec::pop were the main bottleneck.
- added `invoke_configurable` function which borrows an interpreter and uses it to run a wasm function
- made copies of benchmarks that recycle interpreters. There were 7, now there are 14. I think its best we get rid of the old ones.

There is still one todo, namely:

- Interpreter::start_execution does not need return type passed as an argument.

### DIFF
--- a/benches/.gitignore
+++ b/benches/.gitignore
@@ -1,3 +1,2 @@
-/target
+target
 *.trace
-
--- a/benches/.gitignore
+++ b/benches/.gitignore
@@ -1,2 +1,3 @@
-target
+/target
 *.trace
+
--- a/benches/src/reuse.rs
+++ b/benches/src/reuse.rs
@@ -1,7 +1,7 @@
 use std::error;
 use std::fs::File;
 use test::Bencher;
-use wasmi::{FuncInstance, ImportsBuilder, Interpreter, Module, ModuleInstance, NopExternals, RuntimeValue};
+use wasmi::{ImportsBuilder, Interpreter, Module, ModuleInstance, NopExternals, RuntimeValue};
 
 // Load a module from a file.
 fn load_from_file(filename: &str) -> Result<Module, Box<error::Error>> {
@@ -39,7 +39,7 @@ fn bench_tiny_keccak(b: &mut Bencher) {
 
 	b.iter(|| {
 		interpreter.reset();
-		FuncInstance::invoke_configurable(&func, &[test_data_ptr], &mut NopExternals, &mut interpreter).unwrap()
+		interpreter.invoke(&func, &[test_data_ptr], &mut NopExternals).unwrap()
 	});
 }
 
@@ -85,7 +85,7 @@ fn bench_rev_comp(b: &mut Bencher) {
 
 	b.iter(|| {
 		interpreter.reset();
-		FuncInstance::invoke_configurable(&func, &[test_data_ptr], &mut NopExternals, &mut interpreter).unwrap()
+		interpreter.invoke(&func, &[test_data_ptr], &mut NopExternals).unwrap()
 	});
 
 	// Verify the result.
@@ -142,7 +142,7 @@ fn bench_regex_redux(b: &mut Bencher) {
 
 	b.iter(|| {
 		interpreter.reset();
-		FuncInstance::invoke_configurable(&func, &[test_data_ptr], &mut NopExternals, &mut interpreter).unwrap()
+		interpreter.invoke(&func, &[test_data_ptr], &mut NopExternals).unwrap()
 	});
 }
 
@@ -174,9 +174,9 @@ fn fac_recursive(b: &mut Bencher) {
 
 	b.iter(|| {
 		interpreter.reset();
-		let value =
-			FuncInstance::invoke_configurable(&func, &[RuntimeValue::I64(25)], &mut NopExternals, &mut interpreter)
-				.unwrap();
+		let value = interpreter
+			.invoke(&func, &[RuntimeValue::I64(25)], &mut NopExternals)
+			.unwrap();
 		assert_matches!(value, Some(RuntimeValue::I64(7034535277573963776)));
 	});
 }
@@ -214,9 +214,9 @@ fn fac_opt(b: &mut Bencher) {
 
 	b.iter(|| {
 		interpreter.reset();
-		let value =
-			FuncInstance::invoke_configurable(&func, &[RuntimeValue::I64(25)], &mut NopExternals, &mut interpreter)
-				.unwrap();
+		let value = interpreter
+			.invoke(&func, &[RuntimeValue::I64(25)], &mut NopExternals)
+			.unwrap();
 		assert_matches!(value, Some(RuntimeValue::I64(7034535277573963776)));
 	});
 }
@@ -255,9 +255,9 @@ fn recursive_ok(b: &mut Bencher) {
 
 	b.iter(|| {
 		interpreter.reset();
-		let value =
-			FuncInstance::invoke_configurable(&func, &[RuntimeValue::I32(8000)], &mut NopExternals, &mut interpreter)
-				.unwrap();
+		let value = interpreter
+			.invoke(&func, &[RuntimeValue::I32(8000)], &mut NopExternals)
+			.unwrap();
 		assert_matches!(value, Some(RuntimeValue::I32(0)));
 	});
 }
@@ -295,7 +295,8 @@ fn recursive_trap(b: &mut Bencher) {
 
 	b.iter(|| {
 		interpreter.reset();
-		FuncInstance::invoke_configurable(&func, &[RuntimeValue::I32(1000)], &mut NopExternals, &mut interpreter)
+		interpreter
+			.invoke(&func, &[RuntimeValue::I32(1000)], &mut NopExternals)
 			.unwrap_err();
 	});
 }
--- a/benches/src/reuse.rs
+++ b/benches/src/reuse.rs
@@ -1,10 +1,7 @@
 use std::error;
 use std::fs::File;
 use test::Bencher;
-use wasmi::{
-	FuncInstance, ImportsBuilder, Interpreter, Module, ModuleInstance, NopExternals, RuntimeValue, StackSize,
-	StackWithLimit,
-};
+use wasmi::{FuncInstance, ImportsBuilder, Interpreter, Module, ModuleInstance, NopExternals, RuntimeValue};
 
 // Load a module from a file.
 fn load_from_file(filename: &str) -> Result<Module, Box<error::Error>> {
@@ -23,12 +20,6 @@ fn load_kernel() -> Module {
 const REVCOMP_INPUT: &'static [u8] = include_bytes!("./revcomp-input.txt");
 const REVCOMP_OUTPUT: &'static [u8] = include_bytes!("./revcomp-output.txt");
 
-fn new_interpreter() -> Interpreter {
-	let value_stack = StackWithLimit::with_size(StackSize::from_element_count(1024 * 1024));
-	let call_stack = StackWithLimit::with_size(StackSize::from_element_count(1024 * 1024));
-	Interpreter::new(value_stack, call_stack)
-}
-
 #[bench]
 fn bench_tiny_keccak(b: &mut Bencher) {
 	let wasm_kernel = load_kernel();
@@ -44,7 +35,7 @@ fn bench_tiny_keccak(b: &mut Bencher) {
 
 	let func = instance.export_by_name("bench_tiny_keccak").unwrap();
 	let func = func.as_func().unwrap();
-	let mut interpreter = new_interpreter();
+	let mut interpreter = Interpreter::new();
 
 	b.iter(|| {
 		interpreter.reset();
@@ -90,7 +81,7 @@ fn bench_rev_comp(b: &mut Bencher) {
 
 	let func = instance.export_by_name("bench_rev_complement").unwrap();
 	let func = func.as_func().unwrap();
-	let mut interpreter = new_interpreter();
+	let mut interpreter = Interpreter::new();
 
 	b.iter(|| {
 		interpreter.reset();
@@ -147,7 +138,7 @@ fn bench_regex_redux(b: &mut Bencher) {
 
 	let func = instance.export_by_name("bench_regex_redux").unwrap();
 	let func = func.as_func().unwrap();
-	let mut interpreter = new_interpreter();
+	let mut interpreter = Interpreter::new();
 
 	b.iter(|| {
 		interpreter.reset();
@@ -179,7 +170,7 @@ fn fac_recursive(b: &mut Bencher) {
 
 	let func = instance.export_by_name("fac-rec").unwrap();
 	let func = func.as_func().unwrap();
-	let mut interpreter = new_interpreter();
+	let mut interpreter = Interpreter::new();
 
 	b.iter(|| {
 		interpreter.reset();
@@ -219,7 +210,7 @@ fn fac_opt(b: &mut Bencher) {
 
 	let func = instance.export_by_name("fac-opt").unwrap();
 	let func = func.as_func().unwrap();
-	let mut interpreter = new_interpreter();
+	let mut interpreter = Interpreter::new();
 
 	b.iter(|| {
 		interpreter.reset();
@@ -260,7 +251,7 @@ fn recursive_ok(b: &mut Bencher) {
 
 	let func = instance.export_by_name("call").unwrap();
 	let func = func.as_func().unwrap();
-	let mut interpreter = new_interpreter();
+	let mut interpreter = Interpreter::new();
 
 	b.iter(|| {
 		interpreter.reset();
@@ -300,7 +291,7 @@ fn recursive_trap(b: &mut Bencher) {
 
 	let func = instance.export_by_name("call").unwrap();
 	let func = func.as_func().unwrap();
-	let mut interpreter = new_interpreter();
+	let mut interpreter = Interpreter::new();
 
 	b.iter(|| {
 		interpreter.reset();
--- a/benches/src/reuse.rs
+++ b/benches/src/reuse.rs
@@ -1,17 +1,10 @@
-#![feature(test)]
-
-extern crate test;
-extern crate wasmi;
-#[macro_use]
-extern crate assert_matches;
-extern crate wabt;
-
 use std::error;
 use std::fs::File;
 use test::Bencher;
-use wasmi::{ImportsBuilder, Module, ModuleInstance, NopExternals, RuntimeValue};
-
-mod reuse;
+use wasmi::{
+	FuncInstance, ImportsBuilder, Interpreter, Module, ModuleInstance, NopExternals, RuntimeValue, StackSize,
+	StackWithLimit,
+};
 
 // Load a module from a file.
 fn load_from_file(filename: &str) -> Result<Module, Box<error::Error>> {
@@ -22,13 +15,23 @@ fn load_from_file(filename: &str) -> Result<Module, Box<error::Error>> {
 	Ok(Module::from_buffer(buf)?)
 }
 
+fn load_kernel() -> Module {
+	load_from_file("./wasm-kernel/target/wasm32-unknown-unknown/release/wasm_kernel.wasm")
+		.expect("failed to load wasm_kernel. Is `build.rs` broken?")
+}
+
 const REVCOMP_INPUT: &'static [u8] = include_bytes!("./revcomp-input.txt");
 const REVCOMP_OUTPUT: &'static [u8] = include_bytes!("./revcomp-output.txt");
 
+fn new_interpreter() -> Interpreter {
+	let value_stack = StackWithLimit::with_size(StackSize::from_element_count(1024 * 1024));
+	let call_stack = StackWithLimit::with_size(StackSize::from_element_count(1024 * 1024));
+	Interpreter::new(value_stack, call_stack)
+}
+
 #[bench]
 fn bench_tiny_keccak(b: &mut Bencher) {
-	let wasm_kernel = load_from_file("./wasm-kernel/target/wasm32-unknown-unknown/release/wasm_kernel.wasm")
-		.expect("failed to load wasm_kernel. Is `build.rs` broken?");
+	let wasm_kernel = load_kernel();
 
 	let instance = ModuleInstance::new(&wasm_kernel, &ImportsBuilder::default())
 		.expect("failed to instantiate wasm module")
@@ -39,17 +42,19 @@ fn bench_tiny_keccak(b: &mut Bencher) {
 		Ok(Some(v @ RuntimeValue::I32(_))) => v
 	);
 
+	let func = instance.export_by_name("bench_tiny_keccak").unwrap();
+	let func = func.as_func().unwrap();
+	let mut interpreter = new_interpreter();
+
 	b.iter(|| {
-		instance
-			.invoke_export("bench_tiny_keccak", &[test_data_ptr], &mut NopExternals)
-			.unwrap();
+		interpreter.reset();
+		FuncInstance::invoke_configurable(&func, &[test_data_ptr], &mut NopExternals, &mut interpreter).unwrap()
 	});
 }
 
 #[bench]
 fn bench_rev_comp(b: &mut Bencher) {
-	let wasm_kernel = load_from_file("./wasm-kernel/target/wasm32-unknown-unknown/release/wasm_kernel.wasm")
-		.expect("failed to load wasm_kernel. Is `build.rs` broken?");
+	let wasm_kernel = load_kernel();
 
 	let instance = ModuleInstance::new(&wasm_kernel, &ImportsBuilder::default())
 		.expect("failed to instantiate wasm module")
@@ -83,10 +88,13 @@ fn bench_rev_comp(b: &mut Bencher) {
 		.set(input_data_mem_offset, REVCOMP_INPUT)
 		.expect("can't load test data into a wasm memory");
 
+	let func = instance.export_by_name("bench_rev_complement").unwrap();
+	let func = func.as_func().unwrap();
+	let mut interpreter = new_interpreter();
+
 	b.iter(|| {
-		instance
-			.invoke_export("bench_rev_complement", &[test_data_ptr], &mut NopExternals)
-			.unwrap();
+		interpreter.reset();
+		FuncInstance::invoke_configurable(&func, &[test_data_ptr], &mut NopExternals, &mut interpreter).unwrap()
 	});
 
 	// Verify the result.
@@ -103,8 +111,7 @@ fn bench_rev_comp(b: &mut Bencher) {
 
 #[bench]
 fn bench_regex_redux(b: &mut Bencher) {
-	let wasm_kernel = load_from_file("./wasm-kernel/target/wasm32-unknown-unknown/release/wasm_kernel.wasm")
-		.expect("failed to load wasm_kernel. Is `build.rs` broken?");
+	let wasm_kernel = load_kernel();
 
 	let instance = ModuleInstance::new(&wasm_kernel, &ImportsBuilder::default())
 		.expect("failed to instantiate wasm module")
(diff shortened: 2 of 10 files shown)


bddap:
> Interpreter::start_execution does not need return type passed as an argument

Just resolved that todo.

benches/src/reuse.rs
```diff
@@ -0,0 +1,310 @@
+use std::error;
```
pepyakin: I don't think it's a good idea to duplicate entire benchmark suite for this...

Vurich:
Haven't looked at the code yet (just got to this PR), but I don't think we should get rid of the benchmarks that create a new `Interpreter` for each iteration. It's useful to know the first-run time taken as well as the time of subsequent runs.

src/runner.rs
```diff
@@ -166,53 +167,63 @@ enum RunResult {
...
-		call_stack.push(initial_frame);
-
-		let return_type = func.signature().return_type();
-
-		Ok(Interpreter {
-			value_stack,
+	/// Initialize an interpreter that will use `value_stack` and `call_stack`.
+	///
+	/// `value_stack` `call_stack` determine the allowed stack size during later executions.
+	pub fn new(
```
Vurich: Can we rename this `with_stacks` and have `new` create an `Interpreter` with sensible defaults (like your `fn new_interpreter` above).

src/runner.rs
```diff
@@ -36,7 +37,7 @@ pub const DEFAULT_CALL_STACK_LIMIT: usize = 64 * 1024;
 /// at these boundaries.
 #[derive(Copy, Clone, Debug, PartialEq, Default)]
 #[repr(transparent)]
-struct RuntimeValueInternal(pub u64);
+pub struct RuntimeValueInternal(pub u64);
```
Vurich: Why does this need to be public?

src/func.rs
```diff
@@ -139,24 +129,43 @@ impl FuncInstance {
...
+	/// Invoke this function.
+	///
+	/// # Errors
+	///
+	/// Returns `Err` if `args` types is not match function [`signature`] or
+	/// if [`Trap`] at execution time occured.
+	///
+	/// [`signature`]: #method.signature
+	/// [`Trap`]: #enum.Trap.html
+	pub fn invoke<E: Externals>(
```
Vurich: Maybe remove this entirely and have `invoke` be a method on `Interpreter`? Combined with my comment below that would mean invoking a function with a new interpreter would be `Interpreter::new().invoke(func, args, ext)` - much more explicit about the fact that you're creating a new `Interpreter`.

src/func.rs
```diff
@@ -230,20 +233,16 @@ impl From<Trap> for ResumableError {
...
 
 /// A resumable invocation handle. This struct is returned by `FuncInstance::invoke_resumable`.
-pub struct FuncInvocation<'args> {
-	kind: FuncInvocationKind<'args>,
+pub struct FuncInvocation {
+	kind: FuncInvocationKind,
 }
 
-enum FuncInvocationKind<'args> {
+enum FuncInvocationKind {
```
Vurich: It doesn't look like we were using the `args`, @pepyakin what was the intention here?

src/runner.rs
```diff
@@ -166,53 +167,63 @@ enum RunResult {
...
+			value_stack: ValueStack(value_stack),
 			call_stack,
-			return_type,
 			state: InterpreterState::Initialized,
-		})
+		}
 	}
 
-	pub fn state(&self) -> &InterpreterState {
+	// Todo, use types to prevent user from calling start_execution on an inretpreter not in Initialized
```
Vurich: Typo: "intretpreter" should be "interpreter"

src/runner.rs
```diff
@@ -166,53 +167,63 @@ enum RunResult {
...
+	// state.
+	/// Wipe all data in this interpreter, so it can be safely reused.
+	pub fn reset(&mut self) {
```
Vurich: Can't this be done automatically on `start_execution`? Why do we force the user to do this?

src/runner.rs
```diff
@@ -166,53 +167,63 @@ enum RunResult {
...
+		self.call_stack.truncate(0);
+		self.state = InterpreterState::Initialized;
+	}
+
+	pub(crate) fn state(&self) -> &InterpreterState {
 		&self.state
 	}
 
-	pub fn start_execution<'a, E: Externals + 'a>(
+	pub(crate) fn start_execution<E: Externals>(
```
Vurich: Why did you make this crate-internal?

src/runner.rs
```diff
@@ -221,10 +232,11 @@ impl Interpreter {
 		Ok(opt_return_value)
 	}
 
-	pub fn resume_execution<'a, E: Externals + 'a>(
+	pub(crate) fn resume_execution<'a, E: Externals + 'a>(
```
Vurich: As above: why is this crate-internal?

src/runner.rs
```diff
@@ -221,10 +232,11 @@ impl Interpreter {
...
 		&mut self,
 		return_val: Option<RuntimeValue>,
 		externals: &'a mut E,
+		return_type: Option<ValueType>,
```
Vurich: Shouldn't we store this in the `Interpreter` instead of forcing the user to pass it in? It just opens the door to logic errors.

src/runner.rs
```diff
@@ -1247,7 +1167,7 @@ impl Interpreter {
 }
 
 /// Function execution context.
-struct FunctionContext {
+pub struct FunctionContext {
```
Vurich: Why does this need to be public?

src/runner.rs
```diff
@@ -1259,7 +1179,7 @@ struct FunctionContext {
 }
 
 impl FunctionContext {
-	pub fn new(function: FuncRef) -> Self {
+	pub(crate) fn new(function: FuncRef) -> Self {
```
Vurich: `pub(crate)` is usually a code smell, it shouldn't be used apart from for internal functions that you want to expose to tests. This can be `pub` or you can rearrange the module hierarchy to allow you to make it private but accessible from the right places. Probably the former.

src/runner.rs
```diff
@@ -1278,21 +1198,15 @@ impl FunctionContext {
 		self.is_initialized
 	}
 
-	pub fn initialize(
-		&mut self,
-		locals: &[Local],
-		value_stack: &mut ValueStack,
-	) -> Result<(), TrapKind> {
+	fn initialize(&mut self, locals: &[Local], value_stack: &mut ValueStack) -> Result<(), TrapKind> {
```
Vurich: Why does this have to be private?

src/runner.rs
```diff
@@ -1419,37 +1311,47 @@ impl ValueStack {
 
 	#[inline]
 	fn top(&self) -> &RuntimeValueInternal {
-		self.pick(1)
-	}
-
-	fn pick(&self, depth: usize) -> &RuntimeValueInternal {
-		&self.buf[self.sp - depth]
+		self.0.top().expect("pre-validated")
```
Vurich: Please make this error message better, something like: "The stack is empty, this should never be possible due to the validation step".

src/runner.rs
```diff
@@ -1419,37 +1311,47 @@ impl ValueStack {
 
 	#[inline]
 	fn top(&self) -> &RuntimeValueInternal {
-		self.pick(1)
-	}
-
-	fn pick(&self, depth: usize) -> &RuntimeValueInternal {
```
Vurich: Why did you get rid of `pick`?

src/validation/func.rs
```diff
@@ -1598,14 +1372,19 @@ fn make_top_frame_polymorphic(
 	let frame = frame_stack
 		.top_mut()
 		.expect("make_top_frame_polymorphic is called with empty frame stack");
-	value_stack.resize(frame.value_stack_len, StackValueType::Any);
+
+	// shrink value stack to match top frame
+	debug_assert!(frame.value_stack_len <= value_stack.len());
+	while value_stack.len() > frame.value_stack_len {
+		value_stack.pop().expect(
+			"frame.value_stack_len >= 0, value_stack.len() > frame.value_stack_len :. value_stack.len() > 0; qed",
```
Vurich: I know we've got terrible "... qed" error messages everywhere, but can you make this from the perspective of if a user ever actually hit it? So something like, "The value stack is empty, this should never happen since `frame.value_stack_len >= 0` and `value_stack.len() > frame.value_stack_len`, so therefore `value_stack.len()` must be more than 0". "Impossible" cases do happen, we've had users come to us with "qed" error messages that they're really confused by since this style of message actually tells you the _opposite_ of what is true (since they're designed to describe the case in which the `expect` did not fire).

src/validation/func.rs
```diff
@@ -1623,19 +1402,16 @@ fn pop_value(
 	let actual_value = if stack_is_empty && is_stack_polymorphic {
 		StackValueType::Any
 	} else {
-		let value_stack_min = frame_stack
-			.top()
-			.expect("at least 1 topmost block")
-			.value_stack_len;
+		let value_stack_min = frame_stack.top().expect("at least 1 topmost block").value_stack_len;
```
Vurich: Again, better error message here please.

src/validation/func.rs
```diff
@@ -1708,11 +1485,10 @@ fn top_label(frame_stack: &StackWithLimit<BlockFrame>) -> &BlockFrame {
...
 
-fn require_label(
-	depth: u32,
-	frame_stack: &StackWithLimit<BlockFrame>,
-) -> Result<&BlockFrame, Error> {
-	Ok(frame_stack.get(depth as usize)?)
+fn require_label(depth: u32, frame_stack: &StackWithLimit<BlockFrame>) -> Result<&BlockFrame, Error> {
+	frame_stack
+		.get_relative_to_top(depth as usize)
+		.ok_or_else(|| Error("non-empty stack expected".into()))
```
Vurich: Since you repeat this string verbatim many times, maybe extract it to a `const`.

src/common/stack.rs
```diff
@@ -1,88 +1,419 @@
...
+use core::usize;
+
 #[allow(unused_imports)]
 use alloc::prelude::*;
 
-#[cfg(feature = "std")]
-use std::error;
-use core::fmt;
+#[derive(Copy, Clone, Debug)]
+pub struct StackOverflow;
```
Vurich: Why remove the stack information from the error? I'm not saying that we necessarily need it, but I would like to know your justification.

src/common/stack.rs
```diff
@@ -1,88 +1,419 @@
...
+			if self.stack.len() == self.limit {
+				return Err(StackOverflow);
+			}
+			// grows exponentially, just like Vec normally does
+			// https://doc.rust-lang.org/1.26.0/src/alloc/raw_vec.rs.html#462
+			let desired_len = self
+				.stack
+				.len()
+				.checked_mul(2)
+				.unwrap_or(usize::MAX)
```
Vurich: You can use `saturating_mul` instead of `checked_mul(...).unwrap_or(usize::MAX)`

src/common/stack.rs
```diff
@@ -1,88 +1,419 @@
...
+
+	pub(crate) fn pop(&mut self) -> Option<T> {
+		self.stack.pop()
+	}
+
+	/// Remove and Return top element. Does not check for emptyness.
+	/// If this is called on a zero length stack, bad things will happen.
+	/// Do not call this method unless you can prove the stack has length.
+	#[inline]
+	pub(crate) unsafe fn pop_unchecked(&mut self) -> T {
```
Vurich: Again, use of `pub(crate)`. This can either be made `pub` or left private but rearranged in the module hierarchy.

src/common/stack.rs
```diff
@@ -1,88 +1,419 @@
...
+		debug_assert!(self.stack.len() < self.stack.capacity());
+		let len = self.stack.len();
+		unsafe {
+			ptr::write(self.stack.get_unchecked_mut(len), value);
+			self.stack.set_len(len + 1);
+		}
+		Ok(())
+	}
+
+	pub(crate) fn pop(&mut self) -> Option<T> {
```
Vurich: Use of `pub(crate)`.

src/common/stack.rs
```diff
@@ -1,88 +1,419 @@
...
+		self.stack.set_len(len - 1);
+		ptr::read(self.stack.get_unchecked(self.stack.len()))
+	}
+
+	/// Return optional reference to item `depth` distance away from top
+	///
+	/// `bstack.get_relative_to_top(0)` gets the top of the stack
+	///
+	/// `bstack.get_relative_to_top(1)` gets the item just below the stack
+	pub(crate) fn get_relative_to_top(&self, depth: usize) -> Option<&T> {
```
Vurich: Generally Rust avoids `get_noun`-format names. I would call this `nth_from_top`.

src/common/stack.rs
```diff
@@ -1,88 +1,419 @@
...
+
+		let index = self.stack.len().checked_sub(1)?.checked_sub(depth)?;
+		debug_assert!(self.stack.len() > index, "guaranteed by previous line");
+		Some(&self.stack[index])
+	}
+
+	/// Return mutable reference to item `depth` distance away from top
+	///
+	/// Does not check whether depth is in range.
+	pub(crate) fn get_relative_to_top_mut_unchecked(&mut self, depth: usize) -> &mut T {
```
Vurich: See above, I'd suggest `nth_from_top_mut_unchecked`. Also, the name is misleading because it _is_ checked, it's just that it panicks if the depth is wrong. `unchecked` is almost exclusively reserved for `unsafe` functions.

src/common/stack.rs
```diff
@@ -1,88 +1,419 @@
...
+//
+// https://rust-lang-nursery.github.io/api-guidelines/type-safety.html#c-custom-type
+
+/// Type for communicating the size of some contigous container.
+/// Used for constructing both [`StackSizeLimit`] and [`StackSizeInitial`].
+#[derive(Eq, PartialEq, Hash, Debug)]
+pub struct StackSize<T> {
+	num_elements: usize,
+	// PhantomData is a zero-sized type for keeping track of T without rustc complaining.
+	phantom: PhantomData<*const T>,
```
Vurich: Why `*const`? Why not just `T`? I think that the dropck works differently for `PhantomData<*const T>` compared to `PhantomData<T>` but I think you should at least add a comment explaining the choice.

src/common/stack.rs
```diff
@@ -1,88 +1,419 @@
...
+	/// Compute StackSize based on allowable memory.
+	///
+	/// ```
+	/// # extern crate wasmi;
+	/// # use wasmi::StackSize;
+	/// let ss = StackSize::<(u8, u8)>::from_byte_count(10);
+	/// assert_eq!(ss.element_count(), 10 / 2);
+	/// ```
+	///
+	/// # Errors
```
Vurich: This header should be "Panics" but kudos for actually adding this section (too many people forget).

benches/.gitignore
```diff
@@ -1,3 +1,2 @@
-/target
+target
```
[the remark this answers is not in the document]
bddap: It's good that you spotted this. The change doesn't belong here. Reverting.

benches/.gitignore
```diff
@@ -1,3 +1,2 @@
...
 *.trace
-
```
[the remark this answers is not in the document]
bddap: done

src/runner.rs
bddap: done

src/runner.rs
bddap:
```rust
pub fn with_stacks(
    value_stack: StackWithLimit<RuntimeValueInternal>,
    call_stack: StackWithLimit<FunctionContext>,
)
```

`RuntimeValueInternal` is in the type signature for `with_stacks`. rustc won't let us make `with_stacks` public unless `RuntimeValueInternal` is also public. This is what happens when `RuntimeValueInternal` is made private:

```
error[E0446]: private type `runner::RuntimeValueInternal` in public interface
```

src/runner.rs
bddap: Same reason as `start_execution`. It's not type safe.. yet.

src/runner.rs
bddap: 3. It doesn't type check function args.

src/runner.rs
bddap: `FunctionContext` is part of the type signature for `Interpreter::with_stacks`.

src/func.rs
bddap: Yeah, that's nice and ergonomic. I went ahead and moved `invoke` into `Interpreter`.

Perhaps `invoke_resumable` should be moved over as well?

src/runner.rs
bddap: fixed

src/common/stack.rs
bddap: Changed it to `nth_from_top_mut_unchecked`.

Do you have a less misleading name in mind?

src/common/stack.rs
bddap: The docs recommend `PhantomData<*const T>` for when your type doesn't need to drop any `T`s. Something about drop check analysis.

https://doc.rust-lang.org/std/marker/struct.PhantomData.html#ownership-and-the-drop-check

src/common/stack.rs
bddap: fixed

src/runner.rs
bddap: It can be done automatically. Your suggestion has been implemented.

src/runner.rs
bddap: Yes.

This will be easiest to implement when I change Interpreter to be type safe.

src/runner.rs
bddap: Changed from `pub(crate)` to `pub`.

src/common/stack.rs
bddap: Changed from `pub(crate)` to `pub`.

src/common/stack.rs
bddap: Changed from `pub(crate)` to `pub`.

src/runner.rs
bddap: ValueStack is private so it can't be part of a public interface.

bddap:
Thanks for the feedback @Vurich and @pepyakin . I'm still working through change requests and CI is failing because pub functions need docs. I'll work on it more tomorrow.

src/runner.rs
bddap: 4. (writing these down for completeness) There is never a situation where `resume_execution` should be called before `start_execution`.

src/validation/func.rs
bddap: done

src/validation/func.rs
bddap: Changed to "Expected a non-empty frame stack."

src/common/stack.rs
bddap: The interpreter converts `StackOverflow` to `TrapKind::StackOverflow` before passing it back to the user. Figured I'd start with a zero-length type to match `TrapKind::StackOverflow`. It would be easy to add extra fields if they are deemed useful.

src/common/stack.rs
bddap: You're right, it's dropck related. https://doc.rust-lang.org/std/marker/struct.PhantomData.html#ownership-and-the-drop-check

Added an explanatory comment.

src/common/stack.rs
bddap: Handy!

Changed it to `saturating_mul`.

src/validation/func.rs
bddap: Extracted to helper function.

```rust
fn underflow_err() -> Error {
    Error("non-empty stack expected".into())
}
```

src/common/stack.rs
bddap: Changed it to `nth_from_top_mut` and documented panic condition.

benches/src/reuse.rs
bddap: Agreed. Which benches do you think we should keep?

bddap:
CI is happy. I've responded to all the change requests. Back to you @Vurich.

src/func.rs
Vurich: Yeah, I think that would be sensible. Could you do it as part of this PR, please? It doesn't make sense to leave it half-done.

benches/src/reuse.rs
Vurich: I think his issue is with the duplication. If we want to have the entire benchmark suite run without reusing the interpreter, we could write a macro that allows you to abstract over both. That's what I did for `smallvec`.

src/common/stack.rs
Vurich: Does not seem to be fixed

bddap:
@Vurich, you mentioned issues with API design so you may like this.

The current iteration of this pr forces the user to construct two bounded stacks themselves in order to create an interpreter. The goal is to prevent users accidentally passing incorrect arguments to `with_stacks`. For example, we wouldn't want the user to mix up `value_stack_size_initial` and `frame_stack_size_limit` (attempting to follow [rust api guidelines](https://rust-lang-nursery.github.io/api-guidelines/type-safety.html)). The current solution is overkill.

It would be simpler for the user to pass in a struct:

```rust
struct InterpreterConfig {
    value_stack_size_initial: usize,
    value_stack_size_limit: usize,
    frame_stack_size_initial: usize,
    frame_stack_size_limit: usize,
}
```

The `InterpreterConfig` struct has a few benefits.
1. The user never gets a chance to mutate either stack.
2. API size is reduced. We no longer need to expose `StackWithLimit`, `RuntimeValueInternal`, etc. The only new public types would be `Interpreter` and `InterpreterConfig`.
3. The user must explicitly enter the name of each argument i.e. `value_stack_size_initial`; arguments are not position based. All four arguments have the same type so explicit names prevent some user mistakes.

I'm going to get started on the change. It will be ready by the time Berlin wakes up.

src/runner.rs
bddap: Made `RuntimeValueInternal` private in b76ad46.

src/runner.rs
bddap: Made `FunctionContext` private in b76ad46.

src/func.rs
bddap: Starting on that now.

Vurich:
So I think that I'm going to close this PR. I'm really sorry, you did some good work here and you're very clearly a passionate, talented programmer. However, I don't think we want to merge this for a couple of reasons:

1. Having a common stack doesn't actually help the performance, the performance wins come from the reuse of the allocation used to store the stacks and we can implement that in a much simpler way
2. The quality of this PR is quite low - you shouldn't feel bad about this though! I've got some tips for how to get your PR accepted next time

So, if you want to submit a PR in the future here are some tips for how to get it accepted:

1. Don't change the formatting. Yeah, it's our fault for not `rustfmt`ing our code but it makes it really hard to review what actually changed. Plus, some projects don't use `rustfmt` at all either from lack of knowledge of it, disagreement with the formatting rules, or just from bone-headed arrogance. On your own projects it's fine to enable `rustfmt`-on-save but in general on other people's code you only want to do it if it doesn't cause git churn.
2. Don't change the API unless absolutely necessary. Our library follows (or tries to follow) semver, which means that you should avoid breaking changes, but on top of that you should avoid exposing things that we want to keep hidden. Stuff like `RuntimeValueInternal` and `InstructionInternal` aren't just kept private for the fun of secrecy, we might want to change them and so we have them be private so that changing them isn't a breaking change. If you definitely need to expose new things, make sure that the API is as high-quality as possible and follows the [Rust API guidelines](https://rust-lang-nursery.github.io/api-guidelines/) - we're not always perfect at this either of course, don't think that a minor violation of these guidelines will get your PR rejected
3. Compiler errors are not justification for making changes. For example, in the most-recent commit you removed some doctests, presumably because they stopped compiling. The simplest fix to a compiler error is normally not the best one, especially in Rust. Try to work out what the problem is that the compiler is trying to tell you about, the compiler is your friend. This is especially true of lifetime errors and problems related to privacy.
4. To generalise the previous 3 points: make the minimum changes possible. Make your commits atomic (i.e. they compile and make sense individually) where possible and make the PR as a whole have the minimum-possible added/subtracted lines. There are exceptions to this, of course, but it's a good rule of thumb.
5. Try to review your own PR thoroughly before submitting it. Reviewing is not our full-time job - in fact, we don't particularly enjoy it - and if we have to have multiple rounds of going back and forth to get issues fixed we're less likely to want to continue and ultimately get the PR accepted. I actually wrote another review that pointed out issues that remain in the code before cancelling the review and writing this rejection comment instead. I'd say that this and 4 is the reason why this PR specifically got rejected, it was far too hard to review and the back-and-forth to get it up to par was taking too much time and effort on our part.
6. Try to get a sense of what constitutes fishy code (i.e. code that might look suspicious to a reviewer) and comment thoroughly. For an example of what I mean by "fishy" and "thoroughly", [here's a comment that I wrote a while ago to document some suspicious-looking `unsafe` code](https://github.com/paritytech/parity-ethereum/blob/0cd1de769b8ffe5f517f07b0d74ebdca2238ec79/ethash/src/compute.rs#L161-L169). This helps reviewers, but more importantly it helps people down the line who want to maintain the code, which includes yourself.
7. Try not to be arrogant when interacting with reviewers. I know that you don't mean to be arrogant! No-one does. It's easy to get defensive when someone is criticising your code. You are mostly OK as far as this point goes (I'm writing this comment so I can link others to it, too) but it doesn't enamour the reviewers to you if you reply to criticisms with "the compiler made me do it" and then mark the comment as resolved

Hopefully you can take this as a learning experience and not be too disheartened! I don't want you to feel like you can't submit a PR to this project, we debated whether to work on getting it merged for a long time before finally (and sadly) coming down on the side of closing it. You clearly know your way around Rust and programming in general, you just need more experience programming in a collaborative context. Good luck, and I hope to see you again.

bddap:
@Vurich, your criticism is excellent and super useful. I really appreciate how honest, empathetic, and specific it is, as well as the time you put into it.

I plan to split the bounded stack struct out into a new crate so your review work will still serve a purpose.

PR #149, opened 2018-11-23
doc(host) Fix the `Externals` example

Hywan:
The example is missing two things:

  * `index` is computed but not used,
  * `check_signature` is never used.

This patch tries to fix that.


pepyakin:
Thank you!

PR #152, opened 2018-12-11
Check type when resuming function

### LINKED ISSUE #150: Optional argument to resume_execution allows value stack underflow.
`Interperter` assumes this invariant:
> An interpreted function will not pop too many or too few values off the value stack.

Since `pop` is in a hot code path, the above invariant allows for optimizations.

If, however, we trusted the user to push a runtime value to the stack, then we popped it, the stack would be one shorter than expected. Near the end of execution, the value stack would underflow.

### Enter `func::resume_execution`

`func::resume_execution` looks like this:

```rust
pub fn resume_execution<'externals, E: Externals + 'externals>(
    &mut self,
    return_val: Option<RuntimeValue>,
    externals: &'externals mut E,
) -> Result<Option<RuntimeValue>, ResumableError> { ... }
```

That `return_val` argument gets passed to `Interpreter::resume_execution`.
When `return_val` is `None`, `Interpreter::resume_execution` ignores it.
When `return_val` is `Some`, it gets pushed to the value stack.

- If the user passes `Some(val)` and the host doesn't pop, the value stack becomes too tall.
- Conversely, passing `None` when the host expects an argument 1) causes the host to pop some random value, and 2) makes the stack one shorter than expected.

### Sugessted fix

Either get rid of the `return_val` argument, or make it mandatory.

Vurich:
Closes #150.

### DIFF
--- a/src/func.rs
+++ b/src/func.rs
@@ -293,16 +293,23 @@ impl<'args> FuncInvocation<'args> {
         return_val: Option<RuntimeValue>,
         externals: &'externals mut E,
     ) -> Result<Option<RuntimeValue>, ResumableError> {
-        match self.kind {
-            FuncInvocationKind::Internal(ref mut interpreter) => {
-                if !interpreter.state().is_resumable() {
-                    return Err(ResumableError::AlreadyStarted);
+        use crate::TrapKind;
+
+        if return_val.map(|v| v.value_type()) != self.resumable_value_type() {
+            return Err(ResumableError::Trap(Trap::new(
+                TrapKind::UnexpectedSignature,
+            )));
+        }
+
+        match &mut self.kind {
+            FuncInvocationKind::Internal(interpreter) => {
+                if interpreter.state().is_resumable() {
+                    Ok(interpreter.resume_execution(return_val, externals)?)
+                } else {
+                    Err(ResumableError::AlreadyStarted)
                 }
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -136,6 +136,7 @@ extern crate libm;
 /// Traps can't be handled by WebAssembly code, but are reported to the embedder.
 #[derive(Debug)]
 pub struct Trap {
+    // Needs to be `pub(crate)` to allow `if let` matching in tests
```
pepyakin: Do we still need that comment?

src/func.rs
```diff
@@ -293,16 +293,23 @@ impl<'args> FuncInvocation<'args> {
         return_val: Option<RuntimeValue>,
         externals: &'externals mut E,
     ) -> Result<Option<RuntimeValue>, ResumableError> {
-        match self.kind {
-            FuncInvocationKind::Internal(ref mut interpreter) => {
-                if !interpreter.state().is_resumable() {
-                    return Err(ResumableError::AlreadyStarted);
+        use crate::TrapKind;
```
pepyakin: Oh, this is a new keyword and apparently it doesn't require the 2018 edition. What is the version of a compiler is required in order to use this keyword?

Vurich: https://github.com/rust-lang/rust/issues/45477

PR #156, opened 2019-01-08
Fixed documentation for used_size

arkpar:


pepyakin:
Thanks, that makes sense!

PR #158, opened 2019-01-15
Expose globals to host

leodasvacas:
The use case is so that they may be reset to their starting value, to get stateless repeated execution of a module. This is done by making `global_by_index` pub and adding `globals` getter. I'm totally open to better ways of doing this.

### DIFF
--- a/src/module.rs
+++ b/src/module.rs
@@ -183,7 +183,10 @@ impl ModuleInstance {
         self.tables.borrow_mut().get(idx as usize).cloned()
     }
 
-    pub(crate) fn global_by_index(&self, idx: u32) -> Option<GlobalRef> {
+    /// Get global by index. Index can be obtained by calling [globals].
+    ///
+    /// [globals]: [#method.globals]
+    pub fn global_by_index(&self, idx: u32) -> Option<GlobalRef> {
         self.globals.borrow_mut().get(idx as usize).cloned()
     }
 
@@ -213,6 +216,17 @@ impl ModuleInstance {
 
     fn push_global(&self, global: GlobalRef) {
         self.globals.borrow_mut().push(global)
+    }
+
+    /// Enumerates over globals, index can be used as an argument to [global_by_index].
+    ///
+    /// [global_by_index]: #method.global_by_index
+    pub fn globals(&self) -> impl Iterator<Item = (u32, GlobalRef)> {
+        let globals = self.globals.borrow().clone();
+        globals
+            .into_iter()
+            .enumerate()
+            .map(|(idx, global)| (idx as u32, global))
     }
 
     fn insert_export<N: Into<String>>(&self, name: N, extern_val: ExternVal) {


leodasvacas:
Thanks for taking a look. Yhea I shouldn't be cloning the whole Vec. Why non-spec compliant, shouldn't the host be able to modify globals?

The use case is for [graph node](https://github.com/graphprotocol/graph-node), we repeatedly call functions on a module, but we'd like to make sure that no state carries over from one call to the other. The obvious way to do this is to re-instantiate the whole module each time. I'm trying to do a more peformant "soft reset" and only reset the globals and zero out the memory. But I haven't gotten it to work far, don't really know why, so I'll probably give up on it. Which means it's fine if you'd rather close this.

leodasvacas:
I've made this just be a getter for the globals, there was no need to expose the indexes.

pepyakin:
Yeah, [here it is](https://webassembly.github.io/spec/core/appendix/embedding.html#store). According to it, you can obtain references only to exported globals.

Oh, this is a really interesting use-case! And actually we have something similar requirement for substrate/polkadot. But I wonder if we could better...

That said, I think nothing bad will happen if I merge this one.

src/module.rs
```diff
@@ -215,6 +215,12 @@ impl ModuleInstance {
         self.globals.borrow_mut().push(global)
     }
 
+    /// Access all globals. This is a non-standard API so it's unlikely to be
+    /// portable to other engines.
+    pub fn globals<'a>(&self) -> &[GlobalRef] {
+        &*self.globals.borrow()
```
pepyakin: I wonder how does this work: is it acctually possible to return a reference returned from `borrow`?

leodasvacas: arg I pushed broken code, I've corrected this to return `cell::Ref`.

leodasvacas:
Thanks for the link, so the spec-compliant equivalent would be a store with a [write_global](https://webassembly.github.io/spec/core/appendix/embedding.html#embed-write-global) method. But wasmi's approach is already differs from APIs as they are in the spec, I think just exposing the GlobalRef is a consistent way of implementing the functionality described in the spec.

I did see some similar things in polkadot code, hopefully we'll land on a good solution to that in the future.

pepyakin:
Yeah, that's true it is differs somewhat but I believe it is still equivalent: 

`write_global(store,globaladdr,val):store | error` 

is equivalent to

https://github.com/paritytech/wasmi/blob/073e4e7f1f4ce1acc4888b5da5887a5b3189af3e/src/global.rs#L58

`&self` can be treated as `globaladdr` and `store` is implicitly represented by a heap, `val` is equal to `RuntimeValue`. `store` is implicitly modified by a side-effect so we don't need to return it and can be replaced by `()` and `error` is can be treated as `Err` : )

leodasvacas:
@pepyakin I assumed the fields on the [store record](https://webassembly.github.io/spec/core/exec/runtime.html#syntax-store) were public. Thanks for merging this!

PR #161, opened 2019-01-20
Replaced hashmap_core with hashbrown

elichai:
Hey,
So it turns out that `hashmap_core` is deprecated, I replaced it with `hashbrown` according to the maintainers suggestion here: https://github.com/Amanieu/hashmap_core/pull/10

The tests didn't pass for me locally, But I hope this is a configuration thing because I see no reason for any meaningful difference here

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,13 +15,13 @@ default = ["std"]
 # Disable for no_std support
 std = ["parity-wasm/std", "byteorder/std"]
 # Enable for no_std support
-# hashmap_core only works on no_std
-core = ["hashmap_core", "libm"]
+# hashbrown only works on no_std
+core = ["hashbrown", "libm"]
 
 [dependencies]
 parity-wasm = { version = "0.31", default-features = false }
 byteorder = { version = "1.0", default-features = false }
-hashmap_core = { version = "0.1.9", optional = true }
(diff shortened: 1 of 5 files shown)


pepyakin:
Cool, thanks!

Robbepop:
I was curious if there are some speed-ups due to this PR since hashbrown is said to be a lot faster than Rust's current hashmap.

These are the results:

```
 name               before.bench ns/iter          after.bench ns/iter           diff ns/iter   diff %  speedup 
 bench_regex_redux  3,507,389                     3,491,674                         -15,715  -0.45%   x 1.00 
 bench_rev_comp     8,519,465                     8,149,651                        -369,814  -4.34%   x 1.05 
 bench_tiny_keccak  4,363,285                     4,361,593                          -1,692  -0.04%   x 1.00 
 fac_opt            15,953                        15,800                               -153  -0.96%   x 1.01 
 fac_recursive      17,493                        17,956                                463   2.65%   x 0.97 
 recursive_ok       1,063,224                     1,181,953                         118,729  11.17%   x 0.90 
 recursive_trap     122,288                       126,955                             4,667   3.82%   x 0.96
```

Robbepop:
Ah forgot about that - good catch!
Apparently it isn't easily possible to test that at the moment as far as I can see.

PR #164, opened 2019-01-25
Procedural macro for deriving `Externals` implementation

pepyakin:
This PR introduces wasmi-derive: a crate that allows users to generate implementations of `Externals` that wraps a given impl block.

Here is an example:

```rust
struct NonStaticExternals<'a> {
    state: &'a mut usize,
}

#[derive_externals]
impl<'a> NonStaticExternals<'a> {
    pub fn hello(&self, a: u32, b: u32) -> u32 {
        a + b
    }

    pub fn increment(&mut self) {
        *self.state += 1;
    }

    pub fn traps(&self) -> Result<(), NoInfoError> {
        Err(NoInfoError)
    }
}
```

the `[derive_externals]` macro will generate a `Externals` implementation for `NonStaticExternals<'a>` and it will also add inherent method called `resolver` for getting an instance of `ModuleImportResolver`.

Types that implement `wasmi::derive_support::IntoWasmValue` are supported in argument type positions and types that implement `wasmi::derive_support::IntoWasmResult` are supported in return type positions.

The macro is simple by itself, however it generates code that depends on type-inference and uses a couple of tricks.


ddorgan:
[clabot:check]

wasmi/examples/tictactoe.rs
```diff
@@ -133,59 +128,40 @@ mod tictactoe {
...
-                Ok(None)
-            }
-            GET_FUNC_INDEX => {
-                let idx: i32 = args.nth(0);
-                let val: i32 = tictactoe::Player::into_i32(self.game.get(idx)?);
-                Ok(Some(val.into()))
-            }
-            _ => panic!("unknown function index"),
+#[wasmi_derive::derive_externals]
+impl<'a> Runtime<'a> {
```
NikVolf: is this example tested anywhere?

because otherwise there are no tests!

ddorgan:
[clabot:check]

wasmi/examples/tictactoe.rs
NikVolf: Also, there will no longer be example for implementing host *without* procedural macro?

derive/src/parser.rs
```diff
@@ -0,0 +1,97 @@
...
+    /// The generics required to implement a trait for this type.
+    pub generics: syn::Generics,
+    /// The type declaration to implement a trait, most typically
+    /// represented by a structure.
+    ///
+    /// E.g.: `Foo<'a>`, `()`
+    pub ty: Box<syn::Type>,
+}
+
+/// Parse an incoming stream of tokens into externalities definition.
```
NikVolf: what is "externalities" in `wasmi` terms?

wouldn't even "externals" be a foreign term for WebAssembly?

pepyakin: Yes, I'd agree that this term migrated from other projects here. What would you propose here?

wasmi/examples/tictactoe.rs
pepyakin: Yes, except from tests.

I don't think it is worthwhile to keep implementing `Externals` manually here since this example doesn't do anything interesting and can be easily changed with the proc-macro.

derive/src/parser.rs
NikVolf: maybe "host bindings", "host functions" or simply "host" (if it later can be modified that way that it works not only with functions) to keep with WebAssembly terminology

also, the attribute name (now "derive_externals") might also reflect this

derive/src/error.rs
```diff
@@ -0,0 +1,37 @@
+use proc_macro2::{Span, TokenStream};
+use quote::{quote_spanned, ToTokens};
+
+macro_rules! err_span {
+    ($span:expr, $($msg:tt)*) => (
+        $crate::error::CompileError::new_spanned(&$span, format!($($msg)*))
+    )
+}
+
+pub struct CompileError {
```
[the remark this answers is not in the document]
pepyakin: Thanks a lot, nice to know!

derive/src/parser.rs
pepyakin: yeah, but I'm afraid we are really too far down the road to do that. There is already `Externals` trait. But yeah would be up to change the doc.

pepyakin:
Low priority for now, closing...

PR #165, opened 2019-01-26
Implement Send & Sync for ModuleInstance and ModuleRef to enable sharing across threads

andrewdavidmackenzie:
I have a project that wishes to execute loaded Web Assembly modules in multiple threads.

To be able to do so, I need to have ModuleRef and ModuleInstance implement the marker traits "Send" and "Sync".

These are added to the modules.rs file as markers, with no code changes required.

I have run and passed tests and hope these changes seem reasonable to you.

### DIFF
--- a/src/module.rs
+++ b/src/module.rs
@@ -39,6 +39,9 @@ use {Error, MemoryInstance, Module, RuntimeValue, Signature, TableInstance};
 #[derive(Clone, Debug)]
 pub struct ModuleRef(pub(crate) Rc<ModuleInstance>);
 
+unsafe impl Sync for ModuleRef {}
+unsafe impl Send for ModuleRef {}
+
 impl ::core::ops::Deref for ModuleRef {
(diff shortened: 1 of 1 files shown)


pepyakin:
Agree. 

The multithreading story is not clear for me as for now, and I'd prefer to implement it after the offical proposal is accepted.

andrewdavidmackenzie:
Thanks for comments guys, kind of makes sense. 

Initially I had a bigger PR with multiple changes from Rc to Arc down the hierarchy, but went with the minimal to get my code compiling.

I do actually wrap ModuleRef in Arc and Mutex already - so I will need to review my own code to see why I'm imposing Send+Sync on the wrapped struct also...and see if I can avoid it here.

Good to know you are looking at multi-threading. I imagine there is some definition of how that will work with the Globals and other things. I'll leave that to you wasm experts for now.

So, closing PR.

PR #167, opened 2019-01-29
Update readme

pepyakin:


jlb6740:
Hi ... I am interested in how small I can make wasmi to see if it is feasible for resource constrained environments.  New to both wasmi and rust and without understanding much of the code, I tried the "core" build assuming this should reduce the size of both the interpreter executable and the wasmi library. For me, the no_std section in the README implied I should update the parity-wasm dependency by adding features = "core" to the list, but this does not build. There is a build error. Was this how it was done in some previous iteration of Rust or this code? Shouldn't this section of the README be updated. Not sure who the gatekeeper is for pull requests but I'd like to contribute to updates like this and also make improvements to the no_std support build.

PR #173, opened 2019-02-27
Add a check if NotStartedModuleRef has a start function

### LINKED ISSUE #172: Checking if NotStartedModuleRef has start without panicking
Hi,
I want to check if NotStartedModuleRef has a start function without panicking,
So I can't use `assert_no_start()` and don't want to use `not_started_instance()` without checking.
Can I make a PR that adds a function to check if it has start?
that way I can check before calling `not_started_instance()`

elichai:
Resolves #172

### DIFF
--- a/src/module.rs
+++ b/src/module.rs
@@ -714,6 +714,14 @@ impl<'a> NotStartedModuleRef<'a> {
         }
         self.instance
     }
+
+    /// Whether or not the module has a `start` function.
+    ///
+    ///
+    /// Returns `true` if it has a `start` function.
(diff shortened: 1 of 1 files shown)


src/module.rs
```diff
@@ -714,6 +714,14 @@ impl<'a> NotStartedModuleRef<'a> {
         }
         self.instance
     }
+
+    /// Whether or not the module has a `start` function.
+    ///
+    ///
```
pepyakin:
```suggestion
```

PR #174, opened 2019-03-12
Added the nightly feature to hashbrown

elichai:
So, when I replaced `hashmap_core` with `hashbrown` I didn't add the nightly feature to it.
This means that it uses `std`.

source: https://github.com/Amanieu/hashbrown/blob/master/src/lib.rs#L46


pepyakin:
Thanks!

PR #176, opened 2019-04-16
Extract validation into a separate crate

pepyakin:
For substrate-contracts we have a use-case for wasm module validation: before storing the code in the code cache we need to validate it. So the idea is that we will reuse the wasmi validation code for that.


src/memory.rs
```diff
@@ -111,7 +108,23 @@ impl MemoryInstance {
     ///
     /// [`LINEAR_MEMORY_PAGE_SIZE`]: constant.LINEAR_MEMORY_PAGE_SIZE.html
     pub fn alloc(initial: Pages, maximum: Option<Pages>) -> Result<MemoryRef, Error> {
-        validate_memory(initial, maximum).map_err(Error::Memory)?;
+        {
+            use std::convert::TryInto;
+            let initial_u32: u32 = initial
+                .0
+                .try_into()
+                .map_err(|_| Error::Memory("initial can't be coerced to u32".into()))?;
```
NikVolf: Value of what "can't be coerced" would be great

src/prepare/compile.rs
```diff
@@ -0,0 +1,1239 @@
...
+    require_label, top_label, BlockFrame, FunctionValidationContext, StackValueType, StartedWith,
+};
+use validation::stack::StackWithLimit;
+use validation::util::Locals;
+use validation::{Error, FuncValidator};
+
+use isa;
+
+/// Type of block frame.
+#[derive(Debug, Clone, Copy, PartialEq)]
```
NikVolf: Why `PartialEq`?

src/prepare/compile.rs
```diff
@@ -0,0 +1,1239 @@
...
+                context.step(instruction)?;
+
+                let target = require_target(
+                    depth,
+                    context.value_stack.len(),
+                    &context.frame_stack,
+                    &self.label_stack,
+                );
+
+                let target = target.expect("validation step should ensure that this doesn't fail");
```
NikVolf: the same, be more explicit about what is "this" and why "this" does not fail

src/prepare/compile.rs
```diff
@@ -0,0 +1,1239 @@
...
+                self.sink.emit_br_table(&targets, default_target);
+            }
+            Return => {
+                let drop_keep =
+                    drop_keep_return(&context.locals, &context.value_stack, &context.frame_stack);
+
+                context.step(instruction)?;
+
+                let drop_keep =
+                    drop_keep.expect("validation step should ensure that this doesn't fail");
```
NikVolf: same

src/prepare/compile.rs
```diff
@@ -0,0 +1,1239 @@
...
+        0
+    } else {
+        if actual_value_stack_height < start_value_stack_height {
+            return Err(Error(format!(
+                "Stack underflow detected: value stack height ({}) is lower than minimum stack len ({})",
+                actual_value_stack_height,
+                start_value_stack_height,
+            )));
+        }
+        if (actual_value_stack_height as u32 - start_value_stack_height as u32) < keep as u32 {
```
NikVolf: Wouldn't exhaustive match for keep be much more clear of what's going on and will avoid all this casts?

```rust
match keep {
   isa::Keep::Single => if actual_value_stack_height + 1 < start_value_stack_height { ... },
   isa::Keep::Single = { ... },
   isa::Keep::None =>  { ... }
}
```

src/prepare/compile.rs
```diff
@@ -0,0 +1,1239 @@
...
+    value_stack_height: usize,
+    frame_stack: &StackWithLimit<BlockFrame>,
+    label_stack: &[BlockFrameType],
+) -> Result<Target, Error> {
+    let is_stack_polymorphic = top_label(frame_stack).polymorphic_stack;
+    let frame = require_label(depth, frame_stack)?;
+
+    // Get the label by the given `depth`.
+    let label = label_stack
+        .get(label_stack.len() - 1 - (depth as usize))
```
NikVolf: use checked sub and expects?

src/prepare/compile.rs
```diff
@@ -0,0 +1,1239 @@
...
+    frame_stack: &StackWithLimit<BlockFrame>,
+) -> Result<isa::DropKeep, Error> {
+    if frame_stack.is_empty() {
+        return Err(Error(
+            "drop_keep_return can't be called with the frame stack empty".into(),
+        ));
+    }
+
+    let is_stack_polymorphic = top_label(frame_stack).polymorphic_stack;
+    let deepest = (frame_stack.len() - 1) as u32;
```
NikVolf: checked_sub and expect here also?

src/prepare/compile.rs
```diff
@@ -0,0 +1,1239 @@
...
+    NotResolved,
+}
+
+struct Sink {
+    ins: isa::Instructions,
+    labels: Vec<(Label, Vec<isa::Reloc>)>,
+}
+
+impl Sink {
+    fn with_instruction_capacity(capacity: usize) -> Sink {
```
NikVolf: just "with_capacity"?

src/prepare/compile.rs
```diff
@@ -0,0 +1,1239 @@
...
+    }
+
+    /// Create a new unresolved label.
+    fn new_label(&mut self) -> LabelId {
+        let label_idx = self.labels.len();
+        self.labels.push((Label::NotResolved, Vec::new()));
+        LabelId(label_idx)
+    }
+
+    /// Resolve the label at the current position.
```
NikVolf: doc comment goes nowhere

validation/src/func.rs
```diff
@@ -0,0 +1,1236 @@
...
+    fn is_any(&self) -> bool {
+        match self {
+            &StackValueType::Any => true,
+            _ => false,
+        }
+    }
+
+    fn value_type(&self) -> ValueType {
+        match self {
+            &StackValueType::Any => unreachable!("must be checked by caller"),
```
NikVolf: must be checked by caller maybe assumes panic, not unreachable

validation/src/func.rs
```diff
@@ -0,0 +1,1236 @@
...
+            module: module,
+            locals: locals,
+            value_stack: StackWithLimit::with_limit(value_stack_limit),
+            frame_stack: StackWithLimit::with_limit(frame_stack_limit),
+            return_type: return_type,
+        };
+        // push_label overflows only if the frame stack overflow happens. We assume that
+        // the limit is greater than 0 here. We don't expect `push_label` getting any other
+        // new errors.
+        let _ = push_label(
```
NikVolf: why not add an expect then?

src/prepare/compile.rs
pepyakin: It seems it is used to be used, but we can remove it now for sure!

src/memory.rs
pepyakin: Makes sense!

src/prepare/compile.rs
pepyakin: Done!

src/prepare/compile.rs
pepyakin: Hm, I am not sure what do you mean. Looking from github, I can tell that this rustdoc belongs to the `resolve_label` function.

validation/src/func.rs
pepyakin: You're right. As far as I remember there was some genius idea why it is that way, but now I doubt that it is worth it.

Although maybe we would want to consider to add a note on setting crazy limits like 1. : )

src/prepare/compile.rs
NikVolf: non-public function so no docs for it, not really important though

src/prepare/compile.rs
```diff
@@ -0,0 +1,1239 @@
+#[allow(unused_imports)]
+use alloc::prelude::v1::*;
+
+use parity_wasm::elements::{BlockType, FuncBody, Instruction};
+
+use validation::func::{
```
NikVolf: those can be merged in one `use`

src/prepare/compile.rs
pepyakin: Ah, I see. It was deliberate. I am doing cause RLS actually shows you these (altthough I dont use it anymore :) )

src/prepare/compile.rs
```diff
@@ -0,0 +1,1239 @@
...
+                        drop: 0,
+                        keep: isa::Keep::None,
+                    },
+                });
+
+                // Resolve `if_not` to here so when if condition is unsatisfied control flow
+                // will jump to this label.
+                self.sink.resolve_label(if_not);
+
+                self.label_stack.pop().unwrap();
```
NikVolf: `expect` with message instead of `unwrap`

src/prepare/compile.rs
```diff
@@ -0,0 +1,1239 @@
...
+                }
+
+                if let Some(drop_keep) = return_drop_keep {
+                    // It was the last instruction. Emit the explicit return instruction.
+                    let drop_keep = drop_keep.expect("validation should ensure this doesn't fail");
+                    self.sink.emit(isa::InstructionInternal::Return(drop_keep));
+                }
+
+                // Finally, pop the label.
+                self.label_stack.pop().unwrap();
```
NikVolf: expect with message here also

src/prepare/compile.rs
```diff
@@ -0,0 +1,1239 @@
...
+                    default,
+                    value_stack_height,
+                    &context.frame_stack,
+                    &self.label_stack,
+                );
+
+                context.step(instruction)?;
+
+                // These two unwraps are guaranteed to succeed by validation.
+                let targets = targets.unwrap();
```
NikVolf: comment might not be enough, better use expect always

NikVolf:
I also totally dislike those formatted strings as errors, but maybe it's for another pr (added #177)

pepyakin:
Yeah, they sucks. The more prominent problem is that we want to use this within runtime code and formatting machinery is not very welcomed there.

PR #181, opened 2019-05-15
Replaced std with core in the TryInto import

elichai:
The CI actually catched this: https://travis-ci.org/paritytech/wasmi/jobs/532821121#L754

So I changed the travis config to not allow nightly fails


pepyakin:
Thanks! I am not completely sure about the `allow_failures` though, but I am up for trying : ) can disable if it gets annoying.

Issue #183, opened 2019-06-07
Consider using BTreeMap for wasmi-validation


pepyakin:
In the light of wasmi-validation must be immune for untrusted input, we might want to reconsider usage of hashbrown for this usecase.

PR #185, opened 2019-06-10
Use hashbrown for std in addition to no_std builds

adam-rhebo:
This simplifies dependency management a little bit and might also yield minimal performance improvements for large modules even though the included benchmarks do not have enough exports to make this measurable.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,7 +13,7 @@ exclude = [ "/res/*", "/tests/*", "/fuzz/*", "/benches/*" ]
 [dependencies]
 wasmi-validation = { version = "0.1", path = "validation", default-features = false }
 parity-wasm = { version = "0.31", default-features = false }
-hashbrown = { version = "0.1.8", optional = true }
+hashbrown = "0.4.0"
 memory_units = "0.3.0"
 libm = { version = "0.1.2", optional = true }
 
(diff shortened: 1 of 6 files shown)


adam-rhebo:
> 
> 
> 
> 
> Many thanks,
> 
> Parity Technologies CLA Bot

[clabot:check]

pepyakin:
The thing is, we most likely want to use `BTreeMap` even for `no_std` builds for some use-cases (see https://github.com/paritytech/wasmi/issues/183).

pepyakin:
> I see. Is BTreeMap usable with no_std out of the box?

Well, it is available with `alloc` create, but this is the case for hashbrown as well. 

>  If so, shall I just retarget this to replace HashMap by BTreeMap?

I think it depends on your use-case: we want to use `BTreeMap` for now because of it's worst-case performance.

adam-rhebo:
> I think it depends on your use-case: we want to use BTreeMap for now because of it's worst-case performance.

Personally, I am interested in an embedded use case where I expect fewer than 10 exports per module and hence `BTreeMap` is fine for me (the compactness of an open-addressed hash table would be nice, but it really won't matter for me).

This MR was really just motivated by simplifying the build system a tiny bit. Hence I'll retarget this to use `alloc::collections::btree_map::BTreeMap`...

adam-rhebo:
Closing in favor of #187

PR #186, opened 2019-06-10
Add ARMv7 as CI test target

### LINKED ISSUE #43: Support 32-bit targets
To properly support them we need to setup CI for these targets

adam-rhebo:
This adds ARMv7 as test target for Travis CI so that issues on this 32-bit platform can be spotted immediately which should also resolve #43.

To do so, I opted to bump the CI to Xenial to avoid needing third-party repositories as everything just works using `crossbuild-essential-armhf` and `qemu-user-static` on this platform.

I also to install packages using `sudo apt-get` instead of the APT `addon` as the latter does not support conditional installs based on the build matrix, but I did not want to use a different installation method of the conditional and the unconditional dependencies.

**Note that the spec tests currently fail due to failures in `wasm_conversions`. I would be glad if someone could help me with fixing those.**

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -2,8 +2,14 @@
 
 set -eux
 
+EXTRA_ARGS=""
+
+if [ -n "${TARGET-}" ]; then
+    EXTRA_ARGS="--target=${TARGET} -- --test-threads=1"
+fi
+
 cd $(dirname $0)
 
-time cargo test --all
+time cargo test --all ${EXTRA_ARGS}
 
 cd -
--- a/test.sh
+++ b/test.sh
@@ -5,7 +5,11 @@ set -eux
 EXTRA_ARGS=""
 
 if [ -n "${TARGET-}" ]; then
-    EXTRA_ARGS="--target=${TARGET} -- --test-threads=1"
+    # Tests build in debug mode are prohibitively
+    # slow when ran under emulation so that
+    # e.g. Travis CI will hit timeouts.
+    EXTRA_ARGS="--release --target=${TARGET}"
+    export RUSTFLAGS="--cfg debug_assertions"
 fi
 
 cd $(dirname $0)
--- a/.cargo/config
+++ b/.cargo/config
@@ -1,0 +1,2 @@
+[target.armv7-unknown-linux-gnueabihf]
+linker = "arm-linux-gnueabihf-gcc"
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,25 +1,24 @@
-dist: trusty
-sudo: required
+dist: xenial
+
 language:
 - rust
 - cpp
-rust:
-  - nightly
-  - stable
-addons:
-  apt:
-    sources:
-    - ubuntu-toolchain-r-test
-    packages:
-    - gcc-8
-    - g++-8
-    - cmake
-env:
-- CC=/usr/bin/gcc-8 CXX=/usr/bin/g++-8
+
+matrix:
+  fast_finish: true
+  include:
+  - rust: nightly
+  - rust: stable
+  - rust: stable
+    env: TARGET=armv7-unknown-linux-gnueabihf
 
 install:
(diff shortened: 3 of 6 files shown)


adam-rhebo:
> 
> 
> 
> 
> Many thanks,
> 
> Parity Technologies CLA Bot

[clabot:check]

pepyakin:
Ok, this turned out to be trickier that I initially thought.

The test that fails is this [one](https://github.com/WebAssembly/spec/blob/7940fa456c6a9243a29dc7da67fe54c01f67a0ca/test/core/conversions.wast#L70). 

It tests `i32.trunc_f32_s` instruction, 

- [here](https://webassembly.github.io/spec/core/exec/instructions.html#exec-cvtop) is the general conversion instruction description.
- [here](https://webassembly.github.io/spec/core/exec/numerics.html#op-trunc-s) is the specific operator description.

[This is](https://github.com/paritytech/wasmi/blob/284c907b29a358d8345c1884dcbfb3e3b661185f/src/runner.rs#L566) the implementation of the opcode and it looks like the problem is in this code:

https://github.com/paritytech/wasmi/blob/284c907b29a358d8345c1884dcbfb3e3b661185f/src/value.rs#L372-L385

The comment mentions that it is UB to convert value such as Inf or NaN. However, it also looks like that it is UB to convert a value that is outside of range of the target type, see https://github.com/rust-lang/rust/issues/10184.

I briefly tried to fix it by adding explicit bounds check but it didn't help. I guess there is a chance that something else might be going on, but I don't have ARM hardware nearby nor don't have a lot of time to look into it. If you decide to fix this and bump into problems feel free to ask me questions.

adam-rhebo:
@pepyakin I had a look and I think anything that tries to detect the failed conversions using `self as $into` will not work as the UB already happened at that point. Hence I pushed a variant that uses arbitrary precision numbers from the `num` crate to perform the conversion which surely has worse performance, but does yield correct results. This can probably go away again when the upstream issue is resolved.

I am not completely sure, but I guess the same could be applied to `impl_wrap_into!(f64, f32);`?

adam-rhebo:
@pepyakin I had to check the cross-compiled tests to use `--release` builds as they would hit the Travis CI build timeouts otherwhise. Normally, this would be unfortunate as it makes for less useful stack traces, but since the tests run under `qemu-arm-static`, the stack traces are already not very useful...

test.sh
```diff
@@ -2,8 +2,14 @@
 
 set -eux
 
+EXTRA_ARGS=""
+
+if [ -n "${TARGET-}" ]; then
+    EXTRA_ARGS="--release --target=${TARGET}"
```
pepyakin: Can we leave a comment why we need this here?

pepyakin: Can we also enable debug-assertions for release mode?

adam-rhebo: I wonder, should I go for `RUSTFLAGS` only for cross-tests or do this unconditionally via `profile.release`?

adam-rhebo: Done by adjust `RUSTFLAGS` so that actual release builds in production are not affected.

adam-rhebo: Done.

adam-rhebo:
>  If anybody comes up with a better way to do this I'd be happy to merge it.

Ideally, rustc will be improved after LLVM has the necessary intrinsics and we will get things like `impl TryFrom<f32> for i32` which we can then just use here instead of the arbitrary precision numbers.

pepyakin:
Yeah. 

But TBH I am still not sure why checking the range before casting doesn't work (as in my tentative fix). I still think that it might be possible to solve it this way.

adam-rhebo:
> But TBH I am still not sure why checking the range before casting doesn't work (as in my tentative fix). I still think that it might be possible to solve it this way.

From my debugging, I saw that the value actually was in range, i.e. for x86-64 and arm, `self.truncate() <= $into::max_value() as $from` was always true. On x86-64, however, `self.truncate()` had a different sign bit from `(self as $into) as $from` which did not held on arm. This is completely "valid" as  `self as $into` is already UB.

PR #187, opened 2019-06-11
Remove hashbrown and use BTree{Map,Set} from the alloc crate

### LINKED ISSUE #183: Consider using BTreeMap for wasmi-validation

adam-rhebo:
wasmi-validation must handle untrusted input and hence we switch from
Hash{Set,Map} (whether std's or hashbrown's) to BTree{Set,Map} to avoid
algorithmic complexity attacks while retaining no_std support.

Closes #183

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,7 +13,6 @@ exclude = [ "/res/*", "/tests/*", "/fuzz/*", "/benches/*" ]
 [dependencies]
 wasmi-validation = { version = "0.1", path = "validation", default-features = false }
 parity-wasm = { version = "0.31", default-features = false }
-hashbrown = { version = "0.1.8", optional = true }
 memory_units = "0.3.0"
 libm = { version = "0.1.2", optional = true }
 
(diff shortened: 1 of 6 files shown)


adam-rhebo:
@pepyakin I added an additional commit remove the `*Set` from `wasmi-validation` entirely, as we should get the same `O(N log N)` worst-case behaviour using a sorted slice that makes due with a single allocation and should yield better memory locality than the `BTreeSet`.

pepyakin:
Thanks!

goral09:
Recent version of Rust *1.36.0* [started using `hashbrown::HashMap` implementation in the standard library](https://github.com/rust-lang/rust/blob/master/RELEASES.md#libraries). Just saying.

adam-rhebo:
> Recent version of Rust _1.36.0_ [started using `hashbrown::HashMap` implementation in the standard library](https://github.com/rust-lang/rust/blob/master/RELEASES.md#libraries). Just saying.

This was about replacing Hash Brown by a different hash map implementation (as it was for the standard library), but rather by replacing a hash map which is susceptible to collision complexity attacks by a BTree-based implemenation which has a much simpler worst-case complexity behavior and also has `no_std` support (which the standard libraries's Hash Brown implementation lacks for now). (Also the standard library uses SipHash as the default hasher instead of FxHash as Hash Brown exactly to harden itself against these complexity attacks.)

PR #188, opened 2019-06-18
Make clearing value stack between export invocations optional

adam-rhebo:
This avoids the main overhead of repeated export invocations by making it optional to clear the value stack after each interpreter run.

This is especially useful if different exports of the same module are invoked repeated so that no unintended information leaks are possible.

### DIFF
--- a/src/runner.rs
+++ b/src/runner.rs
@@ -1493,6 +1493,22 @@ impl StackRecycler {
         }
     }
 
+    /// Clears any values left on the stack to avoid
+    /// leaking them to future export invocations.
+    ///
+    /// This is a secondary defense to prevent modules from
+    /// exploiting faulty stack handling in the interpreter.
+    ///
+    /// Do note that there are additional channels that
+    /// can leak information into an untrusted module.
+    pub fn clear(&mut self) {
+        if let Some(buf) = &mut self.value_stack_buf {
+            for cell in buf.iter_mut() {
+                *cell = RuntimeValueInternal(0);
+            }
+        }
+    }
+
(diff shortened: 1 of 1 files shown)


adam-rhebo:
From how I understand WASM, clearing the stack is actually not necessary at all as WASM modules should not be able to observe "uninitialized" stack values (except for bugs in the interpreter) and we could leave it out completely. But I admittedly do not feel confident enough to make that call...

pepyakin:
Explicit clearing should not be required. If information leaks this way, then for our use case we arguably will have much more serious problems : )

It also seems to me that leaking the shadow stack (i.e. the linear memory) is much serious than leaking value stack. So I'd be ok without clearing the value stack and the call stack at all (altough I won't be against a paranoid mode).

adam-rhebo:
> It also seems to me that leaking the shadow stack (i.e. the linear memory) is much serious than leaking value stack. So I'd be ok without clearing the value stack and the call stack at all (altough I won't be against a paranoid mode).

Personally, I'd remove it completely but as you say, leaving in the paranoid mode as an explicit API call does not really hurt us as well. So if you want to keep the API, we just merge this. If you want it removed to minimize the API surface, just let me know and I change this.

pepyakin:
I think I am ok with it! Although I'd add a comment clarifying that it is not required and is useful for adding another layer of protection against leaking information and that this is not a silver bullet and the user should remember about other possibilities of leaking.

adam-rhebo:
> Although I'd add a comment clarifying

I tried to add something useful but please feel free to edit the wording before merging.

pepyakin:
Thanks @adam-rhebo !

PR #189, opened 2019-06-19
Add feature flag for opt-in thread-safety

cmichi:
This PR adds a new feature `threadsafe` which switches the concurrency-relevant types to Rust's thread-safe types (`Arc`, `Mutex`, etc.). These types are computationally more expensive than the types which are used now, hence it is a feature.

If the feature is enabled this package can be used in concurrent code as well (e.g. passing `ModuleRef` or `ModuleInstance` among threads). Without this feature this is not possible, because the inner values of e.g. a `Mutex` need to implement `Send` as well.

Background:
We need this for https://github.com/paritytech/substrate/issues/2051, where we want to try out the idea of using a pool of pre-spawned runtime instances instead of a runtime cache. To achieve this, a background thread will continuously spawn new instances to the pool by cloning `ModuleInstance`. Therefore `ModuleInstance` needs to be sent to the background thread.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,6 +17,7 @@ memory_units = "0.3.0"
 libm = { version = "0.1.2", optional = true }
 num-rational = "0.2.2"
 num-traits = "0.2.8"
+atomic = { version = "0.4", optional = true }
 
 [dev-dependencies]
 assert_matches = "1.1"
@@ -24,6 +25,7 @@ rand = "0.4.2"
 wabt = "0.6"
 
 [features]
+threadsafe = ["atomic"]
 default = ["std"]
 # Disable for no_std support
 std = [
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "wasmi"
-version = "0.4.5"
+version = "0.4.6"
 authors = ["Nikolay Volf <nikvolf@gmail.com>", "Svyatoslav Nikolsky <svyatonik@yandex.ru>", "Sergey Pepyakin <s.pepyakin@gmail.com>"]
 license = "MIT/Apache-2.0"
 readme = "README.md"
--- a/README.md
+++ b/README.md
@@ -18,6 +18,7 @@ git clone https://github.com/paritytech/wasmi.git --recursive
 cd wasmi
 cargo build
 cargo test
+cargo test --features threadsafe
 ```
 
 # `no_std` support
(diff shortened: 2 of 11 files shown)


cmichi:
Running `cargo bench` in `benches/` results in these numbers:

`master`:
```
test bench_regex_redux ... bench:   3,772,972 ns/iter (+/- 241,703)
test bench_rev_comp    ... bench:   8,057,183 ns/iter (+/- 496,574)
test bench_tiny_keccak ... bench:   4,251,015 ns/iter (+/- 647,441)
test fac_opt           ... bench:      30,175 ns/iter (+/- 5,271)
test fac_recursive     ... bench:      31,950 ns/iter (+/- 11,601)
test recursive_ok      ... bench:   1,168,398 ns/iter (+/- 167,902)
test recursive_trap    ... bench:     143,066 ns/iter (+/- 17,004)
```


This PR:
```
test bench_regex_redux ... bench:   5,417,944 ns/iter (+/- 1,852,166)
test bench_rev_comp    ... bench:  12,131,625 ns/iter (+/- 4,515,406)
test bench_tiny_keccak ... bench:   5,426,550 ns/iter (+/- 1,779,248)
test fac_opt           ... bench:      37,224 ns/iter (+/- 18,375)
test fac_recursive     ... bench:      39,011 ns/iter (+/- 14,108)
test recursive_ok      ... bench:   1,357,506 ns/iter (+/- 436,465)
test recursive_trap    ... bench:     169,120 ns/iter (+/- 56,261)
```

Vurich:
You should replace `Mutex<RuntimeValue>` with a thread-safe version of `RuntimeValue`. Mutexes are slow, but worse they add an allocation for every value and decrease the compiler's ability to optimise. A thread-safe version of `Cell<RuntimeValue>` would look like so:

```rust
struct ThreadSafeRuntimeValue {
    tag: AtomicU8,
    value: AtomicU64,
}
```

You could even remove the tag and just have the value. This should be approximately the same cost as replacing `Rc` with `Arc`, it just changes some instructions from non-atomic to atomic. Same with `Mutex<usize>`/`Mutex<u32>`/etc, just use `AtomicUsize` and `AtomicU32`. `RefCell<Vec<u8>>` is unfortunately unavoidable.

You can also use Crossbeam's [`Atomic`](https://docs.rs/crossbeam/0.7.1/crossbeam/atomic/struct.AtomicCell.html), but that probably doesn't work for `RuntimeValue` since it's too big (it will fall back to global locks).

pepyakin:
While this is something that we might want to persue in future, for now the issue of optimization of the wasm runtime in substrate/polkadot is being solved with #190.

PR #210, opened 2019-09-08
Use a Cow for the resumable parameters

tomaka:
Right now, if you call `invoke_resumable`, the function arguments are borrowed by the `FuncInstance`. This is very impractical, as you can't really store the `FuncInstance` in a struct for later.

This PR replaces the `&'args [RuntimeValue]` with a `Cow<'a, [RuntimeValue]>`.
This is technically a breaking change, because unfortunately the type `[RuntimeValue; N]` doesn't implement `Into<Cow<[RuntimeValue]>>`.
For example if someone was calling `invoke_resumable(&[a, b])`, it will stop compiling and they need to do `invoke_resumable(&[a, b][..])` instead.

(fwiw, this PR is for a personal project, not a paritytech thing)

### DIFF
--- a/src/func.rs
+++ b/src/func.rs
@@ -1,4 +1,5 @@
 use alloc::{
+    borrow::Cow,
     rc::{Rc, Weak},
     vec::Vec,
 };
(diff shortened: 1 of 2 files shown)


pepyakin:
This part of the API is highly experimental, so I guess it is OK to introduce a minor breakage.

PR #220, opened 2019-10-24
Render constrained system friendly

ithinuel:
This is to allow building and running wasmi on constrained target such as microcontroller.
This requires the ability to build without std.

### DIFF
--- a/src/module.rs
+++ b/src/module.rs
@@ -730,6 +730,26 @@ impl<'a> NotStartedModuleRef<'a> {
         Ok(self.instance)
     }
 
+    /// Executes `start` function (if any) and returns fully instantiated module.
+    ///
+    /// # Errors
+    ///
+    /// Returns `Err` if start function traps.
+    pub fn run_start_with_stack<E: Externals>(
+        self,
+        state: &mut E,
+        stack_recycler: &mut StackRecycler,
+    ) -> Result<ModuleRef, Trap> {
+        if let Some(start_fn_idx) = self.loaded_module.module().start_section() {
+            let start_func = self
+                .instance
+                .func_by_index(start_fn_idx)
+                .expect("Due to validation start function should exists");
+            FuncInstance::invoke_with_stack(&start_func, &[], state, stack_recycler)?;
+        }
+        Ok(self.instance)
+    }
+
     /// Returns fully instantiated module without running `start` function.
     ///
     /// # Panics


src/runner.rs
```diff
@@ -18,10 +18,10 @@ use value::{
 use {Signature, Trap, TrapKind, ValueType};
 
 /// Maximum number of bytes on the value stack.
-pub const DEFAULT_VALUE_STACK_LIMIT: usize = 1024 * 1024;
+pub const DEFAULT_VALUE_STACK_LIMIT: usize = 16 * 1024;
 
 /// Maximum number of levels on the call stack.
-pub const DEFAULT_CALL_STACK_LIMIT: usize = 64 * 1024;
+pub const DEFAULT_CALL_STACK_LIMIT: usize = 16 * 1024;
```
pepyakin: Why would you like to change the defaults? I think they are reasonable for defaults...

ithinuel: The initial incentive in this change is that a lot of the device don't exceed a total of 128KB of RAM.

But that change was before I added `run_start_with_stack`. I remove this from the PR.

ithinuel:
@pepyakin Thanks for your review.
This is to allow building and running wasmi on constrained target such as microcontroller.
This requires the ability to build without std.
I updated the PR's description.

Since your first review I updated the truncation related changes. I removed the dependency on `BigRational`. It does not break the tests and allows to build with nostd.

ithinuel:
This work is also related to https://github.com/paritytech/parity-wasm/pull/283

src/nan_preserving_float.rs
```diff
@@ -63,6 +62,9 @@ macro_rules! float {
             }
 
             pub fn fract(self) -> Self {
+                #[cfg(feature = "libm")]
+                #[allow(unused_imports)]
+                use libm::{F32Ext, F64Ext};
```
pepyakin: Why do we need this?

src/value.rs
```diff
@@ -369,11 +369,7 @@ macro_rules! impl_try_truncate_into {
     (@primitive $from: ident, $into: ident, $to_primitive:path) => {
         impl TryTruncateInto<$into, TrapKind> for $from {
             fn try_truncate_into(self) -> Result<$into, TrapKind> {
-                // Casting from a float to an integer will round the float towards zero
-                num_rational::BigRational::from_float(self)
-                    .map(|val| val.to_integer())
-                    .and_then(|val| $to_primitive(&val))
-                    .ok_or(TrapKind::InvalidConversionToInt)
+                $to_primitive(&self.trunc()).ok_or(TrapKind::InvalidConversionToInt)
```
pepyakin: As far I remember the current code was introduced to support 32-bit x86 platforms. Your PR is essentially reverting it here, right?

ithinuel:
Hi there, sorry for the way too long delay since my last update.
I ran some tests and removed the changes that are no longer required (around the float primitives).

pepyakin:
Thanks, @ithinuel !

PR #224, opened 2019-11-04
Futures

sampaioletti:
Hello, thanks for the great project.

We are using it in a demo for a new rust based run-time for our product and needed to be able to do asynchronous host functions.

Use case being running multiple wasm modules on the same thread, when some/all have asynchronous io so we dont want to prevent other work from occurring.

I first tried to make it work with the unstable async/await rust api and wasn't happy with all the added workarounds that were required (mostly closures/traits) and we are currently using tokio 1.2 for our demo...so i just reworked some of the functions to return futures.

Currently missing good docs, tests, and a lot of finesse, to be clear I'm new to rust as a language so if this is just plain bad code i apologize ahead of time (but comments gladly accepted)

Some of the work in the interpreter loop seems like it could be done better and I would like a few things to be more compatible with the current api
- RuntimeArgs are passed as vecs instead of slices so they can be owned
- Rc around externals doesn't work with imports builder

Only other thought would be the use of the term x_async in the methods, with the upcoming 'async/await' these methods might be better called x_future so they dont get confused with the new rust async api.

Not sure if this would be of any use to the project, but would appreciate comments either way!


pepyakin:
Hello,

Your use case does make sense! However, isn't [`invoke_resumable`](https://crates.parity.io/wasmi/struct.FuncInstance.html#method.invoke_resumable) is what you want here? It is a low-level API, however, I believe that async/futures implementation could be implemented on top of it.

sampaioletti:
@pepyakin ! yes I tried that originally and for some reason (escapes me at the moment) I couldn't make it work, or it was messy in my scenario. Since i've had more experience with wasmi at this point I'll try it again and make sure I can make a case for the addition of this api. Thanks for your help.

sampaioletti:
Oh, and sorry my build failed from formatting errors...rookie mistake (:

athei:
Closed for lack of activity.

PR #230, opened 2020-02-18
Fix wasmi no_std support

mystor:
Two issues are fixed:
 1. `num-bigint 0.2` requires `std`, and is used to perform float to int conversions since #186. This patch disables this change in no_std environments. This change can be reverted once `num-bigint 0.3` is released, which will support no_std.
 2. The `f{32,64}::fract` method is currently not implemented by `core`. This patch uses the no_std supporting implementation from `num-trait`'s `FloatCore` trait.

Travis tests were also updated to check the crate with `cargo no-std-check` which should fail the build if new std dependencies are added in the future.


mystor:
[clabot:check]

PR #232, opened 2020-03-31
made libc optional so that it isnt pulled when not using libstd

reuvenpo:
Currently when compiling wasmi to targets that do not support libc, building fails even when using the `core` feature instead of the `std` feature. This PR fixes that :smiley:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,7 @@ memory_units = "0.3.0"
 libm = { version = "0.1.2", optional = true }
 num-rational = { version = "0.2.2", default-features = false }
 num-traits = { version = "0.2.8", default-features = false }
-libc = "0.2.58"
+libc = { version = "0.2.58", optional = true}
 errno = { version = "0.2.4", optional = true }
 
 [dev-dependencies]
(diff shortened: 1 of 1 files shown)


reuvenpo:
[clabot:check]

reuvenpo:
Reviewers, please note that the CI failed because the `rustfmt` component is missing for the nightly compiler configured in your settings.

reuvenpo:
Tagging recent participants in the git tree:
@bddap @pepyakin @taegyunkim @mystor

reuvenpo:
`wasm32-unknown-unknown`, for one :smile:.
Specifically I'm trying to compile wasmi as no_std into an SGX enclave using a custom target provided by the Baidu Rust SGX SDK... but you can test this by compiling to any target that does not support libc, which is about half the tier-2 targets.

pepyakin:
Hm, both 

```
cargo +nightly build --target x86_64-fortanix-unknown-sgx --no-default-features --features core
cargo build --target wasm32-unknown-unknown --no-default-features --features core
```

work for me **without** this PR.

reuvenpo:
interesting... perhaps libc has some support for them in case it is being compiled to those targets..?
In any case, this PR should improve compile times for users of no_std because libc won't be pulled and built when wasmi isn't using it

reuvenpo:
Yup, this is indeed the case:
https://github.com/rust-lang/libc/blob/7477c68275b1b6da3e8ce81d0545c9484236b35f/src/lib.rs#L145
libc provides support for compiling to official targets.

PR #234, opened 2020-07-05
Added conversion from wasmi::Error to Option<Box<dyn HostError>>

reuvenpo:
This sort of conversion is useful for extracting custom error types from the runtime when the error type is not Clone. Even when it is, it's more efficient to unwrap it that to clone it.


reuvenpo:
Seems like tests are failing in CI in code that's not related to my PR

pepyakin:
Thanks for your PR! Yeah, the CI failure doesn't seem related at all.

Issue #240, opened 2020-11-02
Update libm to fix floating point panics for debug no_std builds


kettle11:
On the README there's this note:
> Floating point operations in no_std use libm, which sometimes panics in debug mode (https://github.com/japaric/libm/issues/4). So make sure to either use release builds or avoid WASM with floating point operations, for example by using deny_floating_point.

The linked issue(https://github.com/japaric/libm/issues/4) was resolved a while back, so `wasmi` can update the version of `libm` to fix the problematic behavior described in the above quote.

PR #241, opened 2020-11-29
Upgrade libm and make no-std config testable

brson:
Just looking for something to hack on I found https://github.com/paritytech/wasmi/issues/240 and decided to give it a try.

Along they way I needed to address a few other things, so they are all included here:

- Make the `--no-default-features --features=core` configuration testable. This mostly involves manually importing std into test modules.
  - One ugly part here is that I had to add some cfgs to the example program so that it works with or without the 'std' config.
- Add the core testing configuration to CI.
- Change the way vec_bytebuf was constructing itself so that it passes the test suite fast
- Restructure how the functions in libm are called since that library no longer includes the traits that were being used.

I can't verify that the libm panicking bug is actually fixed since it doesn't seem to be triggered by the test suite.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,6 +30,7 @@ script:
 # Check that `vec_memory` feature works.
 - cargo check --features vec_memory
 - travis_wait 60 ./test.sh
+- TEST_NO_STD=1 travis_wait 60 ./test.sh
 - ./doc.sh
 
 after_success: |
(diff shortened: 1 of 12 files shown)


pepyakin:
Thanks!

> I can't verify that the libm panicking bug is actually fixed since it doesn't seem to be triggered by the test suite.

Unfortunately, FWIW, I've never bumped into this bug. Still I am up for merging this, so hopefully this issue is resolved.

Issue #194, opened 2019-07-10
Look into integration of cargo-clippy


pepyakin:
Now it shows quite some warnings. On the other hand, many of them are false positives.

Robbepop:
With #244 and #245 merged we should be able to close this issue. @pepyakin

pepyakin:
Thanks @Robbepop !

PR #244, opened 2021-01-30
Applies all clippy suggestions

Robbepop:
Implements: https://github.com/paritytech/wasmi/issues/194

In a few cases a clippy annotation was provided to ignore the clippy warning.

- https://rust-lang.github.io/rust-clippy/master/index.html#expect_fun_call
- https://rust-lang.github.io/rust-clippy/master/index.html#manual_range_contains

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -96,6 +96,8 @@
 
 #![warn(missing_docs)]
 #![cfg_attr(not(feature = "std"), no_std)]
+#![allow(clippy::len_without_is_empty)]
+#![allow(clippy::new_ret_no_self)]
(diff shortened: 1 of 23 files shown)


src/lib.rs
```diff
@@ -333,6 +332,7 @@ impl Error {
     }
 }
 
+#[allow(clippy::from_over_into)]
```
Robbepop, line 335: Generally I don't think that this impl is good and would rather remove it. However, I did not want to break users with this PR.

pepyakin: Agreed

PR #246, opened 2021-02-07
Add MemoryInstance::direct_access and direct_access_mut

tomaka:
These are more flexible alternatives compared to `with_direct_access` and `with_direct_access_mut` which require passing closures.


pepyakin:
Agreed

PR #248, opened 2021-03-04
Support deny only f64 and check max memory size

jiayihu:
- Updated parity-wasm to `0.42.0`
- Added method to deny only f64 values. Many ARM devices have f32 support, but not f64
- Added method to enforce the max number of pages in memory. This also is useful for constrained devices where the memory is limited by design and hardware capacity


jiayihu:
@ithinuel can you provide input on these changes? Your input about the stack size was very valuable and today I've been able to migrate to the solution you proposed. I hope these changes make sense

pepyakin (commented):
Hey, @jiayihu. To be completely honest, the `deny_floating_point` was a mistake in a hindsight. Mainly, it was introduced to cover the use cases for smart-contract execution: they require full determinism and FP operations are one of the sources that may produce non deterministic results. 

With time, we've realized that there are a lot of other use-cases that require inspection and/or instrumentation of a wasm module before execution. Thus, in our downstream crates we moved to processing of the wasm files before execution and no longer use `deny_floating_point`. I would've removed it, but it may still be marginally useful because of some edge-cases. 

However, going forward, I would prefer to avoid adding these kind of functionality into wasmi. That shouldn't be a big problem though since this logic can be implemented outside wasmi.

jiayihu:
Thanks for the kind reply, that makes a lot of sense. I wondered myself to opt-out `validation` from wasmi in order to decrease the binary size, which is always an issue on embedded. In my case, validation could be done elsewhere such as even statically.

PR #243, opened 2021-01-19
Forward parity-wasm's feature 'reduced-stack-buffer'

ithinuel:
This PR introduces a new feature as a forward of underlying feature introduced in https://github.com/paritytech/parity-wasm/pull/283.

PR #251, opened 2021-03-14
Update to parity-wasm 0.42.2

ithinuel:
Update to latest parity-wasm's release in preparation for https://github.com/paritytech/wasmi/pull/243.

Based on @jiayihu's proposal in https://github.com/paritytech/wasmi/pull/248/commits/f7235e13583d34749910abfd7a31fe9855228a33


athei:
@pepyakin Any idea why the CI is failing? The errors seem unrelated to the PR.

pepyakin:
Oh I think it's a recently [added](https://github.com/rust-lang/rust/pull/82770) assert_matches.

pepyakin (approved):

src/prepare/mod.rs
```diff
@@ -154,7 +154,7 @@ pub fn deny_floating_point(module: &Module) -> Result<(), Error> {
                         if func
                             .params()
                             .iter()
-                            .chain(func.return_type().as_ref())
+                            .chain(func.results().first())
```
pepyakin, line 157: Ah, missed this one. Here we check that none of the function parameters have a FP type, and here it checks only the first result. While it's ok-ish since wasmi will fail on validation of multivalue result types, it would be still good to fix.

ithinuel: Do you mean we should then chain the whole `results()` iterator?

ithinuel: Related to the same change for return types, shall the `Signature` be changed to reflect that? Instead of only picking the first type in `from_element` in `src/types.rs`.

pepyakin: Yes, we should go over all the result types, not just the first one.

The signature should be left as is since wasmi doesn't support the multivalue proposal as of now. As I mentioned above, multivalue code will be rejected anyway

ithinuel: Should now be fixed.

athei (approved):

PR #255, opened 2021-04-15
Minor fix to prevent a warning related to Rust 2021 in the tests

chevdor:
Prevent a warning while running the tests.

### DIFF
--- a/src/nan_preserving_float.rs
+++ b/src/nan_preserving_float.rs
@@ -96,6 +96,7 @@ macro_rules! float {
             }
         }
 
+        #[allow(clippy::cmp_owned)]
         impl<T: Into<$for> + Copy> PartialEq<T> for $for {
             fn eq(&self, other: &T) -> bool {
                 $is::from(*self) == $is::from((*other).into())
--- a/src/nan_preserving_float.rs
+++ b/src/nan_preserving_float.rs
@@ -96,6 +96,7 @@ macro_rules! float {
             }
         }
 
+        // clippy suggestion would fail some tests
         #[allow(clippy::cmp_owned)]
         impl<T: Into<$for> + Copy> PartialEq<T> for $for {
             fn eq(&self, other: &T) -> bool {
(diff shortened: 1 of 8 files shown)


tests/spec/run.rs
```diff
@@ -491,7 +491,7 @@ fn try_spec(name: &str) -> Result<(), Error> {
         for err in errors {
             write!(out, "{}", err).expect("Error formatting errors");
         }
-        panic!(out);
+        panic!("{}", out);
```
Robbepop, line 494: Maybe you could use Rust's newly stabilized `panic_any` here: https://doc.rust-lang.org/std/panic/fn.panic_any.html

chevdor: Sure, let me have a look

chevdor: `out` being a `String`, what benefit do you see using `panic_any` ?

Robbepop: ah then it probably won't make sense. please go forward. :)

chevdor:
The clippy issues seem to be related to an issue on the latest nightlies.

- rustc 1.51.0 (2fd73fabe 2021-03-23) => pass
- rustc 1.53.0-nightly (07e0e2ec2 2021-03-24) => pass
- rustc 1.53.0-nightly (d0695c908 2021-04-12) => fail

src/nan_preserving_float.rs
```diff
@@ -96,6 +96,7 @@ macro_rules! float {
             }
         }
 
+        #[allow(clippy::cmp_owned)]
```
pepyakin, line 99: When silencing clippy please leave a short comment on why you think it's fine here: is that a false positive? is it fine because of other reasons?

validation/src/func.rs
```diff
@@ -237,6 +237,7 @@ impl<'a> FunctionValidationContext<'a> {
                     top.block_type
                 };
 
+                #[allow(clippy::branches_sharing_code)]
```
pepyakin, line 240: ditto

pepyakin (approved):
Thanks! Sorry for missing this

src/nan_preserving_float.rs
chevdor: I actually tried the suggestion but it fails the tests then. Clippy seems to base the suggestion on *some* runs of the macro but does not pay attention to the whole.

PR #261, opened 2021-07-08
Add default rustfmt config

chevdor:
Allows specifying that the project uses the default config and prevents using the users global rustfmt config.

The file is left empty on purpose and it matches the current formatting of the code, thus generates no changes.
Having this empty file tells rustfmt to enforce the default and NOT look at the users' global config if there is one.


chevdor:
The reported clippy issues are fixed in #255 so #255 should be looked at first.

pepyakin:
Thanks, this makes sense!

PR #262, opened 2021-07-30
Loadable functions support

0x7CFE:
This PR is a proof-of-concept implementation of loadable functions. It is designed specifically with https://github.com/paritytech/substrate/issues/9354 in mind.

The main idea is to allow to instantiate a module where some (or all) of function bodies are absent. When VM hits such a function, it calls back to a provided implementation of a `Loader` trait to load the function body, and then resumes the execution of a function as if the function body was always there.

That way it would be possible to pass only the functions that are actually needed to dispatch a particular call, effectively reducing the size of PoV.

Further work may include a support for chunked function execution, where function body is loaded chunk by chunk depending on the offsets that were actually hit during the function execution.

### DIFF
--- a/src/runner.rs
+++ b/src/runner.rs
@@ -404,8 +404,8 @@ impl Interpreter {
                     break;
                 }
                 InstructionOutcome::LoadInstruction => {
-                    let (index, _body) = match function_context.function.as_internal() {
-                        FuncInstanceInternal::Internal { function_index, body, .. } => (*function_index, body),
+                    let index = match function_context.function.as_internal() {
+                        FuncInstanceInternal::Internal { function_index, .. } => *function_index,
                         FuncInstanceInternal::Host { .. } => unreachable!("must be internal"),
                     };
 
--- a/src/runner.rs
+++ b/src/runner.rs
@@ -1,8 +1,8 @@
 #![allow(clippy::unnecessary_wraps)]
 
-use crate::func::{FuncInstance, FuncInstanceInternal, FuncRef};
+use crate::func::{FuncBody, FuncInstance, FuncInstanceInternal, FuncRef};
 use crate::host::Externals;
-use crate::isa;
+use crate::{ModuleInstance, isa};
 use crate::memory::MemoryRef;
 use crate::memory_units::Pages;
 use crate::module::ModuleRef;
@@ -16,6 +16,7 @@ use alloc::{boxed::Box, vec::Vec};
 use core::fmt;
 use core::ops;
 use core::{u32, usize};
+use std::rc::Rc;
 use parity_wasm::elements::Local;
 use validation::{DEFAULT_MEMORY_INDEX, DEFAULT_TABLE_INDEX};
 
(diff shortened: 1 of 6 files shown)


src/runner.rs
```diff
@@ -161,6 +162,17 @@ enum RunResult {
     NestedCall(FuncRef),
 }
 
+pub trait Loader {
+    fn load_function_body(&self, index: usize) -> Option<FuncBody>;
```
athei, line 166: This is returning the wasmi `FuncBody` which is private to the crate. Any other crate cannot construct a `Some` value here. I think we should use the `parity_wasm::elements::FuncBody` here.

0x7CFE: > This is returning the wasmi FuncBody which is private to the crate.

I know. My original intention was to provide a deserializing type that would implement the interface. Something like this:

```rust
struct DeserializingLoader<'a>(Fn(usize) -> &'a [u8]);

impl<'a> Loader for DeserializingLoader<'a> {
    fn load_function_body(&self, index: usize) -> Option<FuncBody> {
        let buffer = self.0(index);
        let mut reader = Cursor::new(buffer);
        let deserialized_body: elements::FuncBody = elements::FuncBody::deserialize(&mut reader).ok()?;

        Some(FuncBody {
            locals: deserialized_body.locals().to_vec(),
            code: todo!() // Compile the function to `isa::Instructions`
        })
    }
}
```

athei: Yeah I am in favor of using `parity_wasm::elements::FuncBody`. As is the code cannot be used outside of the crate.

0x7CFE:
In the end, using `parity_wasm::elements::FuncBody` for `Loader` API resulted in some complications with internal instruction compiling.

I reverted that back to using `wasmi::FuncBody` which considerably simplifies the logic and moreover, allows dynamic instruction loading and hot-patching. This was already implemented in the latest commit.

P.S.: Haven't exposed private types into public yet, it's just a PoC.

0x7CFE:
Last version exposes all required types. Everything should be fine now. See [the test](https://github.com/paritytech/wasmi/pull/262/files#diff-ae6804909350d018e1baf3ab33bb18e56354b188134c15a8b873ebc690b95c65R119-R154) for reference.

src/tests/wasm.rs
```diff
@@ -88,6 +89,96 @@ fn load_from_file(filename: &str) -> Module {
...
+            // In a real setup these should be loaded from an external resource
+            let locals = self.bodies.get(index)?.locals().to_vec();
+            let code = self.code_map.get(index)?;
+
+            // Allocating body filled with placeholder instructions to be patched later
+            let len = code.instructions().len();
+            let mut stub = Instructions::with_capacity(len);
+            for _ in 0..len {
+                stub.push(crate::InstructionInternal::NotLoaded);
+            }
```
athei, line 135: Shouldn't the whole code be replaced with a single `NotLoaded` instruction? Otherwise what is the point of stubbing it when it is replaced with a code of the same size.

src/runner.rs
```diff
@@ -242,7 +287,7 @@ impl Interpreter {
                 .map_err(Trap::new)?;
         }
 
-        self.run_interpreter_loop(externals)?;
+        self.run_interpreter_loop(externals, &())?; // TODO loader
```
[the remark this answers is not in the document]
Robbepop: With nightly `rustfmt` we can protect against `TODO`s and `FIXME`s in the code.
Optionally filtered by those that have an associated issue to them

athei:
I will this because we plan to implement code merkelization without loadable function support: https://github.com/paritytech/substrate/issues/9431

Loadable function support would be very hard to do efficiently once the execution engine is inside the client (host function overhead) and would also need to be implemented inside a potential compiler (wasmer singlepass).

PR #271, opened 2021-12-08
Update num-rational 0.2.2 -> 0.4

Robbepop:


athei (approved):
Should be fine using the `std` version for both. Maybe we should add dependabot to this repo?

Issue #270, opened 2021-11-05
Remove `core` feature


athei:
A dependent needs to supply the `core` feature to wasmi in order to use it in a `no_std` build. This is against best practices in the cargo ecosystem because it makes using it awkward.

We should remove this feature and assume a "core build" whenever `std` is not supplied.

PR #272, opened 2021-12-08
Remove core crate feature

Robbepop:
Closes https://github.com/paritytech/wasmi/issues/270.

In summary this PR does the following:

- Removes the `core` crate feature. Explanation can be found in #270 .
- Replaces the `vec_memory` crate feature with the new `virtual_memory` crate feature.
    - The reason is that the new `virtual_memory` crate feature is additive unlike the old `vec_memory`.
    - Users of the crate are recommended to enable the `virtual_memory` crate feature if they are building `wasmi` for a platform that supports OS based virtual memory. Otherwise `wasmi` will fallback onto the vector based implementation.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,8 +27,8 @@ script:
 - if [ "$TRAVIS_RUST_VERSION" == "nightly" ]; then cargo check --benches --manifest-path=benches/Cargo.toml; fi
 # Make sure `no_std` version checks.
 - if [ "$TRAVIS_RUST_VERSION" == "nightly" ]; then cargo no-std-check --no-default-features --features core; fi
-# Check that `vec_memory` feature works.
-- cargo check --features vec_memory
+# Check that `virtual_memory` feature works.
+- cargo check --features virtual_memory
 - travis_wait 60 ./test.sh
 - TEST_NO_STD=1 travis_wait 60 ./test.sh
 - ./doc.sh
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,7 +43,6 @@ std = [
 core = [
     # `core` doesn't support mmaped memory
     "vec_memory",
-    "validation/core",
     "libm"
 ]
 # Enforce using the linear memory implementation based on `Vec` instead of
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,7 @@ memory_units = "0.3.0"
 libm = "0.2.1"
 num-rational = { version = "0.4", default-features = false, features = ["num-bigint"] }
 num-traits = { version = "0.2.8", default-features = false }
-libc = { version = "0.2.58",  optional = true }
+libc = { version = "0.2.58", optional = true }
 errno = { version = "0.2.4", optional = true }
 downcast-rs = { version = "1.2.0", default-features = false }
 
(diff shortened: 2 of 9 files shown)


src/memory/mod.rs
```diff
@@ -10,11 +10,11 @@ use core::{
 };
 use parity_wasm::elements::ResizableLimits;
 
-#[cfg(all(unix, not(feature = "vec_memory")))]
+#[cfg(all(unix, feature = "virtual_memory"))]
 #[path = "mmap_bytebuf.rs"]
 mod bytebuf;
 
-#[cfg(any(not(unix), feature = "vec_memory"))]
+#[cfg(not(all(unix, feature = "virtual_memory")))]
```
athei, line 17: I don't like the automatic fallback but I guess we need to live with it because you can't have target depended features and it would make it annoying for downstream crates if we don't had the fallback. Maybe we can have some compile time warning if this fallback is happening?

.travis.yml
```diff
@@ -27,8 +27,8 @@ script:
 - if [ "$TRAVIS_RUST_VERSION" == "nightly" ]; then cargo check --benches --manifest-path=benches/Cargo.toml; fi
 # Make sure `no_std` version checks.
 - if [ "$TRAVIS_RUST_VERSION" == "nightly" ]; then cargo no-std-check --no-default-features --features core; fi
-# Check that `vec_memory` feature works.
-- cargo check --features vec_memory
+# Check that `virtual_memory` feature works.
+- cargo check --features virtual_memory
```
athei, line 31: I think we can just delete this whole file (in another PR). We only use github actions. So no changes needed here.

src/memory/mmap_bytebuf.rs
```diff
@@ -5,8 +5,9 @@
 //! memory up to maximum. This might be a problem for systems that don't have a lot of virtual
 //! memory (i.e. 32-bit platforms).
 
-use std::ptr::{self, NonNull};
-use std::slice;
+use alloc::string::{String, ToString};
+use core::ptr::{self, NonNull};
+use core::slice;
```
athei, line 10: Strictly speaking we don't need this change because `virtual_memory` activates `std`.

.travis.yml
Robbepop: will do in a micro follow-up PR

src/memory/mod.rs
Robbepop: This should improve once we support more than just "unix" platforms for virtual memory.

Cargo.toml
```diff
@@ -15,10 +15,10 @@ exclude = [ "/res/*", "/tests/*", "/fuzz/*", "/benches/*" ]
 validation = { package = "wasmi-validation", version = "0.4", path = "validation", default-features = false }
 parity-wasm = { version = "0.42.0", default-features = false }
 memory_units = "0.3.0"
-libm = { version = "0.2.1", optional = true }
+libm = "0.2.1"
 num-rational = { version = "0.4", default-features = false, features = ["num-bigint"] }
 num-traits = { version = "0.2.8", default-features = false }
-libc = { version = "0.2.58", optional = true}
+libc = { version = "0.2.58",  optional = true }
```
athei, line 21: Double space. I think I have done that myself :D

athei (approved):

.travis.yml
kpp: Resolved in https://github.com/paritytech/wasmi/pull/273

src/memory/mod.rs
kpp: Resolved by https://github.com/paritytech/wasmi/pull/274

Issue #200, opened 2019-07-29
Use VirtualAlloc on Windows


pepyakin:
At the moment, wasmi's linear memory is backed by `mmap` on "unix". However, on Windows it falls back to the vec implementation. 

We could use `VirtualAlloc`, an equivalent of `mmap` for our purposes, to implement the backend for linear memories on Windows.

PR #274, opened 2021-12-08
Support virtual memory usage on windows

Robbepop:
Implements https://github.com/paritytech/wasmi/issues/200.

# Summary

- Enabling `virtual_memory` crate feature enables cross-platform virtual memory on 64-bit platforms.
- On 32-bit platforms we are falling back onto using the `Vec`-based implementation. This has several reasons:
    1) We want to keep the (initial) virtual memory implementation small and simple.
    2) What we did so far with respect to `mmap` usage in `wasmi` is not expected to yield performance improvements due to a huge overhead with re-allocations and copying over all data upon a `grow` operation.


src/memory/mmap_bytebuf.rs
```diff
@@ -5,181 +5,125 @@
...
-        // allocated by `mmap`.
-        // However, for the cases when it actually fails prefer to fail, in order to not leak
-        // and exhaust the virtual memory.
-        assert_eq!(ret_val, 0, "munmap failed");
+    /// Returns an exclusive slice over the bytes of the virtual memory allocation.
+    #[inline]
+    pub fn as_slice_mut(&mut self) -> &mut [u8] {
+        // # SAFETY
+        //
+        // See safety proof of `Mmap::as_slice`.
```
pepyakin, line 59:
```suggestion
        // See safety proof of `VirtualMemory::as_slice`.
```

or

```suggestion
        // See safety proof of `Self::as_slice`.
```

?

Robbepop: good catch. but none of the above since this is a pitfall for renaming the type.

src/memory/mmap_bytebuf.rs
```diff
@@ -5,181 +5,125 @@
...
 
 impl ByteBuf {
+    /// Determines the initial size of the virtual memory allocation.
+    ///
+    /// # Note
+    ///
+    /// In this implementation we won't reallocate the virtually allocated
+    /// buffer and instead simply adjust the `len` field of the `ByteBuf`
+    /// wrapper in order to efficiently grow the virtual memory.
+    const ALLOCATION_SIZE: usize = u32::MAX as usize;
```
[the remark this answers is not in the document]
Robbepop: what would you recommend then? 2GB? or a whole different approach? note that most platforms where `wasmi` runs on are 64-bit so we should keep our focus on those with respect to performance.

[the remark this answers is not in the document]
pepyakin: While I agree with you that 64-bit is our primary target, I think the compatibility is also important. I think a marginal win in performance does not outweigh the feature of running on 32-bit environments.

There are a couple of possible ways forward:

- Either return to the previous approach where we allocate as needed, or
- fallback in the case of 32-bit environments. We could, for example, fallback on the allocate as needed for them, or just plainly fallback to the vec impl.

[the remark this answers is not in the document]
Robbepop: Reallocating the virtual memory + copying over all its contents seems to be destroying the purpose of using virtual memory in the first place so I'd simply fallback on 32-bit platforms for now.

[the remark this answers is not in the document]
athei: I agree. We should just fall back to the `Vec` implementation on 32-bit platforms. Otherwise we would need to provide a configuration option for how much of the virtual memory space we are allowed to take per instance.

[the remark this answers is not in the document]
Robbepop: done

src/memory/mod.rs
```diff
@@ -10,11 +10,11 @@ use core::{
 };
 use parity_wasm::elements::ResizableLimits;
 
-#[cfg(feature = "virtual_memory")]
+#[cfg(all(feature = "virtual_memory", target_pointer_width = "64"))]
```
athei: Maybe you can make use of https://crates.io/crates/efg here?

Robbepop: i don't think it is worth it to put a whole new dependency in here for that tbh. I was also thinking about using `cfg_if` but same reason.

Robbepop: besides that: let's keep the PR small and focused and do stuff like this in follow-ups if we deem it useful.

src/memory/mmap_bytebuf.rs
```diff
@@ -5,181 +5,125 @@
...
 
-    fn as_slice_mut(&mut self) -> &mut [u8] {
-        unsafe {
-            // Safety Proof:
-            // - See the proof for `Self::as_slice`
-            // - Additionally, it is not possible to obtain two mutable references for `self.ptr`
-            slice::from_raw_parts_mut(self.ptr.as_ptr(), self.len)
-        }
+    /// Returns a shared slice over the bytes of the virtual memory allocation.
+    #[inline]
```
athei, line 42: Is this faster than whatever the compiler selects? Proof or remove.

src/memory/mmap_bytebuf.rs
```diff
@@ -5,181 +5,125 @@
...
-            libc::munmap(self.ptr.as_ptr() as *mut libc::c_void, self.len)
-        };
-
-        // There is no reason for `munmap` to fail to deallocate a private annonymous mapping
-        // allocated by `mmap`.
-        // However, for the cases when it actually fails prefer to fail, in order to not leak
-        // and exhaust the virtual memory.
-        assert_eq!(ret_val, 0, "munmap failed");
+    /// Returns an exclusive slice over the bytes of the virtual memory allocation.
+    #[inline]
```
athei, line 55: Is this faster than whatever the compiler selects? Proof or remove.

src/memory/mmap_bytebuf.rs
```diff
@@ -5,181 +5,125 @@
...
+        // This operation is only actually needed in order to make the
+        // Vec-based implementation less inefficient. In the case of a
+        // virtual memory with preallocated 4GB of virtual memory pages
+        // we only need to adjust the `len` field.
+        self.len = new_len;
         Ok(())
     }
 
+    /// Returns the current length of the virtual memory.
+    #[inline]
```
athei, line 103: Is this faster than whatever the compiler selects? Proof or remove.

src/memory/mmap_bytebuf.rs
```diff
@@ -5,181 +5,125 @@
...
 
+    /// Returns a shared slice over the bytes of the virtual memory allocation.
+    #[inline]
     pub fn as_slice(&self) -> &[u8] {
-        self.mmap.as_ref().map(|m| m.as_slice()).unwrap_or(&[])
+        &self.mem.as_slice()[..self.len]
     }
 
+    /// Returns an exclusive slice over the bytes of the virtual memory allocation.
+    #[inline]
```
athei, line 115: Is this faster than whatever the compiler selects? Proof or remove.

src/memory/mmap_bytebuf.rs
```diff
@@ -5,181 +5,125 @@
...
-            // The order is important.
-            //
-            // 1. First we clear, and thus drop, the current mmap if any.
-            // 2. And then we create a new one.
-            //
-            // Otherwise we double the peak memory consumption.
-            self.mmap = None;
-            self.mmap = Some(Mmap::new(len)?);
-        }
+        self.mem = VirtualMemory::new(Self::ALLOCATION_SIZE)?;
```
athei, line 126: Is there a test that verifies that the a new mapping is zeroed? I am asking because I am not so sure that this is guaranteed on windows.

Robbepop: I think it is safe to assume that even on windows it will be initialized to zeros:
https://docs.microsoft.com/en-us/windows/win32/api/memoryapi/nf-memoryapi-virtualalloc

athei: Okay.

athei (approved):

src/memory/mmap_bytebuf.rs
```diff
@@ -5,181 +5,121 @@
...
-                dst[..amount].copy_from_slice(&src[..amount]);
-            }
-            Some(new_mmap)
-        };
-
-        self.mmap = new_mmap;
+        // This operation is only actually needed in order to make the
+        // Vec-based implementation less inefficient. In the case of a
+        // virtual memory with preallocated 4GB of virtual memory pages
+        // we only need to adjust the `len` field.
```
kpp, line 95: We are in `target_pointer_width = "64"` so usize::BITS=64. Is there anything to stop us from `.realloc(2^32+1)`?

Robbepop: no.
we are aware about guard pages with additional page allocation but wanted to keep the implementation simple.

kpp: Wouldn't `assert!(new_len <= ALLOCATION_SIZE);` keep the impl simple?

athei: The `len` should always reflect the current size of the linear memory in wasm semantics. We need to bound check memory access.

kpp: I mean one can do `byte_buf.realloc(2^32+1)` and then call `byte_buf.as_slice()` which will give a slice that references to a out-of-bound mem

PR #275, opened 2021-12-09
GHA: Run tests on multiple platforms

Robbepop:
Carved out from #274.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -100,18 +100,6 @@ jobs:
         with:
           command: clippy
           args: --features virtual_memory -- -D warnings
-
-  clippy_no_std:
-    name: Clippy (no_std)
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v2
-      - uses: actions-rs/toolchain@v1
-        with:
-          profile: minimal
-          toolchain: nightly
-          override: true
-          components: clippy
       - uses: actions-rs/cargo@v1
         with:
           command: clippy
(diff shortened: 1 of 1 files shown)


Robbepop:
GHA CI passes fully. Note that some checkmarks are not there since our branch protection settings need to be updated in order to resemble the new CI passes.

athei:
Can you add `--test-threads 1` here? Adding this can also fail tests if they make use of thread local storage (some do).

Robbepop:
> Can you add `--test-threads 1` here? Adding this can also fail tests if they make use of thread local storage (some do).

You mean as an entire new test pass or add to existing test pipeline?

athei:
An extra pipeline which runs `windows + virtual_mem` with `--test-threads 1`.

Robbepop:
> An extra pipeline which runs `windows + virtual_mem` with `--test-threads 1`.

The `windows.latest` + `features virtual_memory` job now runs using `--test-threads 1` whereas everything else runs in parallel.

.github/workflows/rust.yml
```diff
@@ -58,9 +67,10 @@ jobs:
       - uses: actions-rs/cargo@v1
         env:
           RUSTFLAGS: '--cfg debug_assertions'
+          RUSTC_FLAGS: ${{ matrix.rustc-args }}
```
athei, line 70: The name is wrong. Those flags aren't passed to rustc. They are passed to the test binary.

Robbepop: is `TEST_FLAGS` better?

athei: Yes! But you also need to rename `rust-args` to `test-args`. :D

athei (approved):

PR #277, opened 2021-12-09
Unify & simplify GitHub Actions check jobs

Robbepop:


athei:
I think it is a good idea to merge it. However, maybe we should do `build` instead of `check`? This would also catch errors that happen at code generation time. Also, maybe add `wasm32-unknown-unknown` as a target platform. We care about this target.

Robbepop:
> I think it is a good idea to merge it. However, maybe we should do `build` instead of `check`? This would also catch errors that happen at code generation time. Also, maybe add `wasm32-unknown-unknown` as a target platform. We care about this target.

what kind of errors happen at `cargo build` but not at `cargo check`?
the purpose of this pipeline is to give a quick feedback so it ends quickly.
the `wasm32-unknown-unknown` target `cargo check` seems like a good idea though.

athei:
> what kind of errors happen at cargo build but not at cargo check?

I would assume anything that happens at the linker stage.

> the purpose of this pipeline is to give a quick feedback so it ends quickly.

Wouldn't you run `cargo check` locally for this? You need all pipelines to be green for merge anyway.

Robbepop:
> > what kind of errors happen at cargo build but not at cargo check?
> 
> I would assume anything that happens at the linker stage.
> 
> > the purpose of this pipeline is to give a quick feedback so it ends quickly.
> 
> Wouldn't you run `cargo check` locally for this? You need all pipelines to be green for merge anyways.

I sure do (+ test) but we have to organize our repo for external users, too.
I can change it to `build` since you are right about the link time errors.

athei (approved):

PR #279, opened 2021-12-09
Add GitHub Action CI job for coverage reporting

Robbepop:
Coverage reports are uploaded to codecov.io: https://app.codecov.io/gh/paritytech/wasmi


athei:
There is an error when clicking the link you posted:
Screenshot 2021-12-09 at 16 34 59

Robbepop:
> 

If you look through the logs of the CI job you will find the following link:
https://codecov.io/github/paritytech/wasmi/commit/c9008a9d9ab9127e8370c22d3d2d969243372977
There it says that the CI pipeline has not yet finished. I assumed they mean the CI pipeline on the codecov.io side because codecov has to generate all the visualizations but maybe that was wrong.
On the other handside you clearly see some coverage number (~79%) which more or less resembles what I get when I run `cargo tarpaulin` locally.
The logs only show one error that does not seem to cause major problems though.
Any ideas on your side?

athei:
I think the error is caused by
"Issue detecting commit SHA. Please run actions/checkout with fetch-depth > 1 or set to 0"

athei (approved):

PR #280, opened 2021-12-09
Deprecate MemoryInstance::get method for wasmi version 0.10

Robbepop:
The method stands in conflict with our API design guidelines:

> 1) Make it easy to use an API correctly and hard to use it incorrectly.

> 2) Good API design is achieved not when there is nothing more to add, but when there is nothing left to take away.

- The short name of the `get` API and the fact that it is slightly simpler to use than alternatives such as `get_into` makes it a preferred choice despite the fact that it is less efficient than the alternative `get_into` or `get_value` APIs. This violating API design rule 1.
- Besides that there is no scenario where a user cannot use one of the alternative APIs instead of `get`. Thus violating API design rule 2.


athei:
So maybe we should then remove the internal users of this deprecated function in this PR?

Robbepop:
> So maybe we should then remove the internal users of this deprecated function in this PR?

you got me there :D

athei (approved):

PR #286, opened 2021-12-14
Bump version 0.9.1 -> 0.10.0

Robbepop:
In preparation of the release of version `0.10.0`.


athei:
Shouldn't that be `0.9.2` because no breaking changes?

Robbepop:
> Shouldn't that be `0.9.2` because no breaking changes?

Introducing a new crate feature and removing another is definitely a breaking change imo.

athei (approved):
You are right. Forgot about that.

Issue #231, opened 2020-03-05
Consider implementing Send for ModuleInstance


That3Percent:
I am working on an application that exposes as a part of its host bindings functions which do asynchronous IO. It would be great to be able to suspend the execution of the module, have the current thread move on to other tasks, and resume execution on another thread when the IO is complete (basic async/await Rust stuff).

I'm not yet sure how big a lift that story would be, but the first problem that I'm running into is that `ModuleInstance` does not impl `Send` because `GlobalRef` contains an `Rc`.

pepyakin:
Hello! Sorry for the late reply.

Generally, I would be up to do that, however I am not yet sure how to achieve that cleanly without regressing the performance too much. I think it might be achievable by using `Arc` with careful caching of the pointers inside of the main interpreter loop (i.e. runner.rs), but I haven't really digged into that.

Robbepop:
The `v1` implementation of `wasmi` (https://github.com/paritytech/wasmi/pull/287) has a radically different architecture so `Instance` (formerly known as `ModuleInstance`) is trivially `Send` now.

Issue #177, opened 2019-04-18
Use strongly-typed errors in validation/compilation/execution


NikVolf:

Robbepop:
The `v1` implementation of `wasmi` supports strongly-typed errors: https://github.com/paritytech/wasmi/pull/287
Therefore this issue can now be closed.

Issue #145, opened 2018-11-15
Allocate all instructions in single block


Vurich:
My idea is to use a wrapper around a buffer with shared ownership, so each `Instructions` type has a pointer to a section of the buffer but the full buffer doesn't get dropped until all owners are dropped (similar to how `Bytes` works). Probably we would have to preallocate a buffer big enough to hold all instructions and then have some kind of `fn extend(iter: impl IntoIterator<Item = Instruction>) -> Option<BufferHandle>` (straw man naming, of course) that locks the shared buffer, writes the elements and then returns a handle that spans the newly-added elements. This avoids dealing with lifetimes and makes `Module::drop` much faster, at the cost of possibly making allocating a new module slower (we can either generate instructions in serial or generate them in parallel but use intermediate buffers that must then be copied from).

Vurich:
Here's a sketch of what that would look like if we keep the restriction of denying unsafe code:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2015&gist=223f92efd86137a3bd89770b045a11b8

pepyakin:
What, can you define `all` for me?

Vurich:
All within a single module, not all across all allocated modules.

Robbepop:
The `v1` implementation of `wasmi` (https://github.com/paritytech/wasmi/pull/287) does allocate all instructions of all functions of the same `Engine` in a single buffer directly implementation the proposal of this issue. Therefore this issue can be closed now.

Issue #136, opened 2018-10-29
Decrease pointer-chasing/improve cache efficiency


Vurich:
There are two problems, the way I see it:

1. Starting or resuming (after a nested call) a function call requires 4 derefs - instructions are a `Vec` in an `Rc` in an `Rc` in a `Vec`. This can blow the cache unless we're extremely lucky with where data gets allocated. Essentially, this means every function call requires up to 8 derefs which may be (and probably will be) on different cache lines. If we implemented TCO then tail calls would only require 4, but that's still bad.
2. The instructions are allocated in separate blocks, meaning more cache incoherency when changing function contexts (again, on calls and returns).

Since optimisers avoid function calls anyway and inline when possible (so the resultant wasm that we're executing should have minimal function calls) the impact of this is somewhat mitigated, but it's something that we can fix, so why not.

Vurich:
So absolute best-case (if everything is in L1), some light googling pegs the speed of a single pointer read as 4 cycles. This is pipelined but it doesn't matter since we're dereferencing pointers-in-pointers and so it's inherently serial. https://www.edn.com/design/systems-design/4399725/Memory-Hierarchy-Design---Part-6--The-Intel-Core-i7

Robbepop:
The `v1` implementation of `wasmi` (https://github.com/paritytech/wasmi/pull/287) now implements a Wasmtime inspired architecture and API and allocates all instructions of all functions in the same `Engine` instance in a single buffer. This heavily reduces the number of indirections from 4 down to 2 when given a `FuncBody` handle therefore closing this issue.

PR #287, opened 2021-12-20
[WIP] Mirror Wasmtime API in wasmi

Robbepop:
This PR builds a new API for `wasmi` inspired by Wasmtime: https://docs.rs/wasmtime/0.32.0/wasmtime/

## Preliminary Benchmarks

```
test bench_compile_and_validate    ... bench:   6,796,640 ns/iter (+/- 371,322)
test bench_compile_and_validate_v1 ... bench:   8,010,305 ns/iter (+/- 308,968)
test bench_instantiate_module      ... bench:     420,247 ns/iter (+/- 32,465)
test bench_instantiate_module_v1   ... bench:      61,995 ns/iter (+/- 3,467)
test bench_tiny_keccak             ... bench:   1,304,626 ns/iter (+/- 51,768)
test bench_tiny_keccak_v1          ... bench:   1,276,489 ns/iter (+/- 41,213)
test fac_opt                       ... bench:      27,856 ns/iter (+/- 4,120)
test fac_opt_v1                    ... bench:         800 ns/iter (+/- 69)
test fac_recursive                 ... bench:      29,064 ns/iter (+/- 4,505)
test fac_recursive_v1              ... bench:       1,907 ns/iter (+/- 248)
test host_calls                    ... bench:      80,189 ns/iter (+/- 7,856)
test host_calls_v1                 ... bench:      78,542 ns/iter (+/- 2,816)
test recursive_ok                  ... bench:     501,695 ns/iter (+/- 109,610)
test recursive_ok_v1               ... bench:     487,757 ns/iter (+/- 53,118)
```

The above benchmarks show that the general overhead of `v1` is greatly reduced compared to `v0` and for long running tasks for engines seem to operate with a comparable performance. Note though that `v1` has not yet been optimized.

Benchmark profile ran with `lto = "fat"` and `codegen-units = 1`, otherwise performance is going to be worse for `v1`.

## Expected Results

- Highly improved usability.
- Removal of tons of `RefCell` and `Rc` that are counter intuitive for Rust.
- Native implementation of the `multivalue` Wasm proposal. (partial)
- A more efficient and extensible runtime and interpreter.
- Ideally: Drop-in replacement between `wasmi` and Wasmtime.

## Follow-up PRs

- [ ] Implement `IntoFunc` for `Func::wrap` in a way that properly supports `Func::typed` later on.
    - The signature checks are currently embedded into the host function trampoline. The consequence of this is that it is impossible to separate signature checks and host function calls. However, in many cases signature checks happen before calling the host function eventually, e.g. in the `Engine` therefore doing duplicate checks at the moment. Also when supporting `Func::typed` API in the future the `TypedFunc` won't be more efficient since we cannot avoid the signature checks.
- [ ] Implement new bench tests based on Criterion.
    - Additionally we want more elaborate tests that test certain performance metrics in isolation.
- [ ] Clean-up: Shift compilation infrastructure from `v1::Module` to `v1::Engine`.
    - This allows us to eventually have multiple different `Engine` implementations, e.g. next to our currently existing `Engine` that has a simple stack based bytecode we could define an `Engine` that compiles to SSA based register bytecode for better performance of long running Wasm executions.
- [ ] Move old definitions into `v0` module under `crate` root.
- [ ] Introduce `v0` crate feature to optionally enable the entire `v0` module.
    - Initially `v0` feature is going to be enabled by default.
- [ ] Introduce `v1` crate feature to optionally enable the entire `v1` module.
    - Initially `v1` feature is _not_ enabled by default.
- [ ] Optimizations for the new `wasmi` bytecode and interpreter.
    - [ ] Implement more compactly encoded `wasmi` bytecode.
    - [x] Filter out no-op operations such as `reinterpret` instructions.
- [ ] Shift over to using `wasmparser` for parsing and validation from `parity_wasm` and `wasmi_validation`.
    - For this we are required to maintain a `no_std` fork of `wasmparser`.
- [ ] Implement resumable functions based on `async` similar to how it is done in Wasmtime.

## Usage

The typical usage of the new API is very similar to the usage of the Wasmtime API. In summary:

- Create an `Engine` type. This is the new `wasmi` interpreter. We call it `Engine` instead of `Interpreter` since it also hosts all the Wasm function bodies as the Wasmtime `Engine` does. This has two main reasons.

## Affected Issues

- https://github.com/paritytech/wasmi/issues/231
    - All user facing types now properly implement `Send` and `Sync` since most of them are just type safe references.
- https://github.com/paritytech/wasmi/issues/177 (partially)
    - The `v1` engine still uses `wasmi-validate` crate with its unstructured errors, however, the unstructured errors of the `prepare` module are gone.
- https://github.com/paritytech/wasmi/issues/145
    - The new `CodeMap` abstraction helps to reduce pointer chasing and stores all function bodies in a single allocation.
- https://github.com/paritytech/wasmi/issues/136 (2 indirections instead of 4)
    - The `FuncBody` reference type directly points into the `CodeMap` where its instructions can be found. So from a `FuncBody` there is exactly one indirection happening to resolve function instructions. Also local variable information now resides directly next to the instructions and therefore also no longer requires yet another pointer chasing.
- https://github.com/paritytech/wasmi/issues/125
    - API parity with Wasmtime is the whole point of the PR.
- https://github.com/paritytech/wasmi/issues/23
    - The entire `v1` uses structured error types and makes no use of `String` types internally.
- https://github.com/paritytech/wasmi/issues/100
    - The new `InstructionBuilder` and `VisitInstruction` abstractions greatly help in experimenting with new `wasmi` bytecode designs since decouple data from operations.

## What the WIP _won't_ do

- Change the parsing and validation infrastructure to Wasmtime's `wasm-tools` such as `wasmparser`, yet.
- Instructions won't be compacted as suggested by #100 but the codebase will be prepared for this step as a follow-up.
- The WIP tries to take as much of the existing `runner.rs`, `compile.rs` as possible but tries to make it simpler for follow-ups to make changes to those parts of the code.

## ToDo List

- [x] Properly implement `Caller` with respect to its internal `Instance` handle.
- [x] Extract maximum stack heights per Wasm function during compilation.
    - This information is required for efficient use of the dynamic value stack of the new `v1` interpreter.
- [x] Fix remaining Wasm spec tests for `v1` engine:
    - [x] `wasm_f32`
    - [x] `wasm_f32_bitwise`
    - [x] `wasm_f64`
    - [x] `wasm_f64_bitwise`
    - [x] `float_misc`
    - [x] `wasm_imports`
    - [x] `wasm_linking`
        - ~~`wasm_linking` test fails because instantiation of a module where the failure happens after element section and data sections are processed may leave the respective manipulated tables or linear memories in an invalid state. The fix is that instantiation should happen as if it was a transaction with this regard.~~ Since version 1.1 WebAssembly no longer demands this behavior and the current `v1` implementation seems to be fine. The only issue is that our currently embraced Wasm spec testsuite is quite outdated. Updating is not easy since many unimplemented Wasm proposals (such as `multivalue`) are included by default in later revisions.
    - [x] `wasm_skip_stack_guard_page`
        - `wasm_skip_guard_page` test fails because the value stack currently is missing adjustment with respect to function stack height requirements. This is planned to be implemented soon anyways. When artificially increasing the default value stack height, the test starts to succeed.
    - [x] `wasm_start`
- [x] Implement Wasm spec test suite tests for the new `v1` interpreter.
    - Use [`wast`](https://crates.io/crates/wast) instead of [`wabt`](https://crates.io/crates/wabt) for parsing of `.wast` files.
- [x] Implement `Instance::exports` method and make `Instance::get_export` public.
    - [`wasmtime::Linker::instance`](https://docs.rs/wasmtime/0.32.0/wasmtime/struct.Linker.html#method.instance)
- [x] Carve out new `LittleEndianConvert` trait into its own PR for `wasmi`.
- [x] Remove `Func::instance` and `Func::func_body` and use `Func::as_internal` instead.
    - Reason: `Func::as_internal` is more efficient and covers more use cases.
- [x] Implement Wasm module instantiation.
    - [x] Implement partial module instantiation where `start` function has not yet been executed.
    - [x] Implement `Linker::instantiate` for better user experience with respect to imports resolution.
    - [x] Implemented `InstancePre` to properly control execution of the `start` function.
- [x] Implement new `wasmi` interpreter.
    - [x] Properly advance the program counter.
    - [x] Lazily load default linear memory and table in `FunctionFrame`.
    - [x] Implement proper function dispatch.
         - [x] Wasm function dispatch.
         - [x] Host function dispatch.
- [x] Compile and validate function bodies into the new `wasmi` bytecode.
    - This is still using the `parity_wasm` `Module` and `wasmi-validator` as well as most of the existing compilation process.
    - The only difference is that the new `wasmi` bytecode is generated instead of the old one and that code is fed into the new `wasmi` `Engine` instead of being collected inside the `wasmi` `Module`.
- [x] Add unit tests for new `wasmi` codegen.
    - Coverage Report: https://github.com/paritytech/wasmi/pull/287#issuecomment-998355267

## Rendered Docs (WIP)

Rendered docs of what has been implemented so far (2021-12-27):

2021-12-27-234408_1299x1202_scrot


Robbepop:
We clarified internally that this PR is too big to be reviewed properly so we concentrate QA on follow-up PRs for extra fuzzing, and polishing.

PR #293, opened 2022-01-01
Fix Float::copysign impl

Robbepop:
I discovered today that the official WebAssembly spec test suite assertions and `wasmi` results diverge in a particular case with the `copysign` instruction.
For `F32::from_bits(0xFFFC_0000).copysign(F32::from_bits(0))` the expected result according to the WebAssembly Spec test suite is `0x7FFC_0000`, however, `wasmi` currently returns `0xFFFC_0000`.
The Rust documentation mandates the same behavior from `copysign` as WebAssembly does: https://doc.rust-lang.org/std/primitive.f32.html#method.copysign

The bug was not caught because the current `wasmi` Wasm spec test runner seems to improperly handle `NaN` values.

Fortunately the bug fix was easy, however, I still wonder why we are going to great lengths like this.

### DIFF
--- a/src/value.rs
+++ b/src/value.rs
@@ -767,11 +767,6 @@ macro_rules! impl_float {
             }
             fn copysign(self, other: $type) -> $type {
                 use core::mem::size_of;
-
-                if self.is_nan() {
-                    return self;
-                }
-
                 let sign_mask: $iXX = 1 << ((size_of::<$iXX>() << 3) - 1);
                 let self_int: $iXX = self.transmute_into();
                 let other_int: $iXX = other.transmute_into();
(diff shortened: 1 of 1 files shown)


athei:
> The bug was not caught because the current wasmi Wasm spec test runner seems to improperly handle NaN values.

Shouldn't we fix this first then? To make sure that this does not brake other things.

Robbepop:
> Shouldn't we fix this first then? To make sure that this does not brake other things.

The `v1` implementation caught this bug since `v1` properly tests this property.
The `v0` implementation of the Wasm spec test builds on `wabt` whereas `v1` builds on `wast` crate.
The fix is very isolated - we are just touching the one particular float operation.
I agree that ideally we should also fix the Wasm spec test suite for `v0` but this should best be done in another PR imo.

athei:
This code is used by `v0` and `v1`?

Robbepop:
> This code is used by `v0` and `v1`?

yes, `v1` already implements the fix. This PR backports it to `v0`.

athei (approved):

PR #299, opened 2022-01-03
Refactor and clean up benchmark code

Robbepop:
- Moves benchmark `.wat` files into their own files.
    - This is useful in cases where they are used by multiple files.
    - Also this separates data from operations.
- Cleans up some benchmark specific code.
- This PR is a good precursor to moving to Criterion.


benches/src/lib.rs
```diff
@@ -30,28 +30,39 @@ use wasmi::{
     Trap,
 };
 
-// Load a module from a file.
-fn load_from_file(filename: &str) -> Result<Module, Box<dyn error::Error>> {
+// Parses the Wasm binary at the given file name into a `wasmi` module.
+//
+// # Note
+//
+// This includes validation and compilation to `wasmi` bytecode.
```
athei, line 37: May be worth converting to rust doc while you at it.

athei (approved):

PR #312, opened 2022-01-07
Improve CI jobs mostly by adding '--workspace' where required

Robbepop:
When working on `wasmi_core` sub crate I found that our current CI did not check certain properties of `wasmi_core`.


athei (approved):

athei:
Would make sense to add all crates that should be build to the `default-members` instead of relying on `--workspace`, though.

PR #316, opened 2022-01-09
Rename RuntimeValue to Value internally

Robbepop:
The `RuntimeValue` type is one of the most commonly used types and therefore should have a short name, however, it has one of the longest type names throughout the `wasmi` codebase.
Other Wasm engines such as Wasmtime go even further and just call it `Val`.

Since the next update is going to be a breaking one (due to the introduction of `v1`) I thought it is okay to ship with some other breaking changes. We could technically provide a type alias
```rust
#[deprecated(note = "just use Value")]
pub type RuntimeValue = Value;
```
To signal users what to do in case they don't read the changelog.

To reviewers: Review commit by commit since I separated formatting from renaming.

### DIFF
--- a/benches/src/lib.rs
+++ b/benches/src/lib.rs
@@ -24,7 +24,7 @@ use wasmi::{
     ModuleInstance,
     NopExternals,
     RuntimeArgs,
-    RuntimeValue,
+    Value,
     Signature,
     TableDescriptor,
     TableRef,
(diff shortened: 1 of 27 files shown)


Robbepop:
Merged since this should be backwards compatible since we still re-export `Value` as `RuntimeValue` from the `wasmi` crate.

PR #351, opened 2022-02-01
Use wat instead of wabt crate where possible

Robbepop:
This will further improve Wasm spec compliance once https://github.com/paritytech/wasmi/pull/346 lands.
The `wat` crate is authored by the BytecodeAlliance and in a much better shape than the `wabt` crate.
Also it is stlil maintained and updated for new Wasm proposals and fixes.

Remaining wabt users are:

- The two outdated fuzz systems.
- The outdated v0 Wasm testsuite.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,6 +19,7 @@ parity-wasm = { version = "0.42.0", default-features = false }
 [dev-dependencies]
 assert_matches = "1.5"
 wabt = "0.9"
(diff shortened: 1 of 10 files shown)


athei (approved):
Yeah makes sense to slowly get rid of C dependencies.

Robbepop:
The CI test job failure seems to be disconnected from the changes introduced by this PR.

PR #352, opened 2022-02-01
CI: pin windows-latest to windows-2019

rcny:
This PR can be used as the possible temporary fix until the project's dependencies will be able to compile on `windows-2022`.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -43,17 +43,19 @@ jobs:
     name: Test
     strategy:
       matrix:
-        os: [ubuntu-latest, windows-latest, macos-latest]
+        # windows-latest was pinned to windows-2019
+        # because of https://github.com/paritytech/wasmi/runs/5021520759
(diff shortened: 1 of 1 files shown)


Robbepop:
Thank you for the PR. So unfortunately the `wabt` dependency causes the trouble.
The good thing is that we are already planning on phasing it out in `wasmi_v1`. The bad thing is that this process will probably take a while to complete.
Until then I guess we unfortunately need to deal with this heavily outdated windows experience.

rcny:
Not sure that `windows-2019` is already outdated. It has mainstream support until the beginning of 2024 and e.g. Rust team is [only planning](https://github.com/rust-lang/rust/issues/92948) migration to `windows-2022`.

Robbepop (approved):
Thanks a lot for fixing the CI.
We really should get `wabt` as dependency out of this library. While `windows-2019` will be supported for quite some time we want to always be working on the latest versions.

Issue #356, opened 2022-02-11
Implement Wasm proposal: `sign-extension-ops`


Robbepop:
This is about adding support for the [`sign-extension-ops` Wasm proposal](https://github.com/WebAssembly/sign-extension-ops).

The `sign-extension-ops` Wasm proposal has already been included into the Wasm standard and is fairly light weight, therefore implementing it won't be a lot of work.

We should generally keep up with supporting finished Wasm proposals as this allows us to keep the Wasm spec test suite up to date.

PR #360, opened 2022-02-13
Support `sign-extension-ops` Wasm proposal

Robbepop:
Closes https://github.com/paritytech/wasmi/issues/356.


Robbepop:
We might want to wait for https://github.com/bytecodealliance/wasm-tools/issues/481 to merge this so that we can safely guard against unsupported or forbidden Wasm proposal feature usage.

Issue #361, opened 2022-02-13
Move to register based bytecode internally


Robbepop:
Currently `wasmi`'s internal bytecode is very similar to WebAssembly.
There are a few Wasm operations, that are unfolded and where certain information, such as local depths or the `DropKeep` information, is precomputed.
Many papers about the topic indicate that register machine based bytecode is significantly more efficient than stack machine based bytecode with respect to runtime performance. However, machine based bytecode will probably consume more memory during runtime.

[This Intel paper](https://www.intel.com/content/www/us/en/developer/articles/technical/webassembly-interpreter-design-wasm-micro-runtime.html) describes an efficient way to translate WebAssembly stack machine based bytecode into a register machine based bytecode. The conversion is very efficient and produces good enough result bytecode. Their own benchmarks indicate performance improvements within the range of 50%-150% improvements.

For `wasmi` we should investigate this optimization as it bears significant wins over our current internal stack machine based bytecode.

Robbepop:
Research conducted in https://github.com/Robbepop/interpreter-dispatch-research indicates that it is best to continue using `switch` based instruction dispatch. Also benchmarks indicate that it is best not to fuse too many instructions together.

For example having `add r a0 a1` which computes `a0 + a1` and stores the result into `r` should only allow registers for `r` and `a0` and allow registers or constants (immediates) for `a2`.
A more flexible approach would be to allow registers, constants, globals and eventually linear memory regions for `r`, `a0` and `a1`. While this would further reduce the amount of instructions needed to represent the same input source code the overhead is too large for it to be an effective optimization.

Therefore looking at the simple `add` instruction above it is best to have a very simple model for the new register machine instructions.

However, this leads to another design choice: since we want to allow either a register or a constant (immediate) for the `a0` parameter of `add` we could either model this with 2 different instructions `add` and `add_imm` or use an `enum { Reg(Register), Imm(Constant) }` for the type of `a0`.
The research benchmarked both cases to be very similar with regards to performance. The approach with 2 instructions might be minimally superior with respect to performance whereas the approach with the `enum` type for `a0` is more flexible in case we want to change our model later on.

Robbepop:
For the sake of simplicity and because we haven't found significant performance different in either approaches we are going the `enum` approach described in the last posting. We can do an optimization follow-up to replace the `enum` approach with the variants approach to see if it actually improves performance.

Robbepop:
[...]

PR #367, opened 2022-02-19
Implement register machine based bytecode

Robbepop:
Closes https://github.com/paritytech/wasmi/issues/361.

## TODO

- [x] Define and document the new register machine based bytecode.
    - https://github.com/paritytech/wasmi/pull/367/commits/f28c58a10cbe5d6719d1d4af6e35edbbbc8e323d
- [x] Implement translation from Wasm to new `wasmi` bytecode.
    - [x] Commutative binary instructions (e.g. `i32.add`, `f32.max`, `i64.eq` etc.)
    - [x] Non-commutative binary instructions (e.g. `i32.sub`, `i64.div`, `f32.copysign` etc.)
    - [x] Order-comparison instructions (e.g. `i32.lt_s`, `f32.ge` etc.)
    - [x] Unary instructions (e.g. `i32.popcnt`, `i64.clz`, `f32.neg` etc.)
    - [x] Conversion instructions (e.g. `i32.wrap_i64`, `f32.reinterpret_i32` etc.)
    - [x] Load instructions (e.g. `i32.load`, `i64.load_8s` etc.)
    - [x] Store instructions (e.g. `i32.store`, `i64.store_8` etc.)
    - [x] Global variabe instructions (`global.get`, `global.set`)
    - [x] Memory instructions (`memory.size`, `memory.grow`)
    - [x] Stack manupulation instructions (`drop`, `select`)
    - [x] Local variable manipulation instructions (`local.set`, `local.tee`)
    - [x] Call instructions (`call` and `call_indirect`)
    - [x] Branching instructions (`return` `br`, `br_if` and `br_table`)
    - [x] Control flow structures (`if`, `else`, `loop` and `block`, `end`)
- [x] Implement new `Engine` that operates on the new `wasmi` bytecode.
    - [x] Implement execution stack data structure.
        - This handles both the function frame stack and their register spaces.
    - [x] Initialization and finalization routines for root function execution.
        - This handles initialization of the root function frame and its parameters
          the handling of execution of the root function outcomes and the final
          writing back the results of the computation to the callee.
        - [x] Initialization and finalization of root Wasm functions.
        - [x] Initialization and finalization of root Host functions.
            - Wasm Spec Test Suite requires this functionality.
        - [x] Wasm function calls.
        - [x] Host function calls.
    - [x] Instruction execution.
        - [x] Execution context definition and implementation.
        - [x] Instruction execution visitor.
        - [x] Unary instructions.
        - [x] Binary instructions.
        - [x] Comparison instructions.
        - [x] Conversion instructions.
        - [x] Call instructions.
            - [x] `call` instruction
            - [x] `call_indirect` instruction
        - [x] Control flow instructions.
            - [x] `br` instruction
            - [x] `br_eqz` instruction
            - [x] `br_nez` instruction
            - [x] `br_table` instruction
            - [x] `return` instruction
            - [x] `return_nez` instruction
            - [x] `trap` instruction
        - [x] Load instructions.
        - [x] Store instructions.
        - [x] Memory instructions
            - [x] `memory.size` instruction
            - [x] `memory.grow` instruction
            - [x] Memory store instructions.
            - [x] Memory load instructions.
        - [x] Misc. instructions (`select`, `copy`)
        - [x] Global variable instructions.
- [x] Debug the new execution engine implementation.
    - [x] Implement human readable debug printing of `wasmi` bytecode.
- [ ] Add engine execution tests.
    - [ ] `trap` test (function that always just traps)
    - [x] `add` test (simple)
    - [ ] `div_rem` test (taking `a` and `b` and returns `(a/b, a%b)`)
    - [x] `swap` test (multi-value returns)
    - [x] `count_until` test (simple loop)
    - [x] `factorial-iterative` test
    - [x] `factorial-recursive` test
    - [x] `fibonacci-iterative` test
    - [x] `fibonacci-recursive` test
    - [x] `deep-recursion` test (recursive calls to self)
    - [x] `is_even` and `is_odd` test (recursive functions that ping pong each other)
    - [x] `memory-sum` test (memory reads)
    - [x] `memory-fill` test (memory writes)
    - [x] `host-call-single-return` test (host calls with single return value)
    - [x] `host-call-multi-return` test (host calls with multiple return values)
    - [x] `host-call-single-param` test (host calls with single parameter)
    - [x] `host-call-multi-return` test (host calls with multiple parameters)
    - [ ] `host-call-trap` test (tests host functions that may trap)
- [x] Use the new `Engine` and the new bytecode in the `wasmi_v1` crate.
- [x] Remove the old execution engine implementation. (v1)
- [ ] Benchmark the new `Engine` against the old `Engine`.

## Known Bugs

- [ ] There currently is a `rustc` bug that pessimizes the performance of `wasmi`'s instruction scheduler by roughly 50-60%. [The issue](https://github.com/rust-lang/rust/issues/95409) has already been reported to the `rustc` developers but there was not yet a response or a fix. This is partly because a minified example is still missing (reporter's work).
    - [ ] Find a minified example that shows the regressed performance and report it to the `rustc` developers.
    - [x] Find and implement a work around for `wasmi` to not be affected by this performance bug.
- [x] The `wasmi` codegen does not properly store return values for branch instructions.
    - **Note:** This includes the `br` at the end of an `if` with an `else` block as well as all `br`, `br_if` and `br_table` constructs. There might be other cases that have yet to be discovered. For the sake of performance we should add a new `copy_many` `wasmi` instruction that copies a bunch of providers to their respective registers instead of generating many single `copy` instructions for each result. https://github.com/paritytech/wasmi/blob/f064a242ee69f8afc7ca8b87a8ddda9b4dea63d7/wasmi_v1/src/engine/func_builder/mod.rs#L523
    - **Proposal:** A possible solution to the problem is to add `results: ExecRegisterSlice` and `returns: ExecProviderSlice` fields to branch instructions such as `br`, `br_if` etc. to efficiently embed the value passing into the branch. This is especially useful for conditional branches such as `br_if` since this prevents unnecessary copying and also makes `br_table` in its current form work out of the box.

## Required Improvements

- [ ] Return `LimitsError` instead of panicking when encountering unsupported or out of bounds IDs.
    - **Example:**
        - Too many registers used by a function.
        - Too many instructions for a single function.
        - etc..

## Experimentation

- [ ] Add `CallSingle` and `CallIndirectSingle` `wasmi` instructions as optimizations for the common cases of calling a function using a single parameter.
- [ ] Improve quality of implementation of `DedupProviderSliceArena`.
- [ ] Optimize host function calls.
    - Currently host function calls allocate registers for their parameters and return values on the stack. This requires copying the parameters over to the newly allocated registers and also copying the return values back to the caller. This overhead can be avoided by using the registers of the caller directly.
- [ ] Use the same `ValueStack` architecture as in `v1` or `v0`.
- [ ] Fuse the `Bytecode::CallIndirect` `index` and `params` field into a single `index_and_params` field with type `ProviderSlice`. This will reduce the size of this large variant slightly.
- [x] Optimize `br_table` Wasm to `wasmi` translation so that not only `BranchMulti` and `ReturnMulti` arms are produced but also `BranchSingle` and `Return` as well as `ReturnImm`.
- [x] Add `ReturnSingle` and `ReturnNezSingle` `wasmi` instructions as optimizations for the common cases of returning a single value back to the caller.
- [x] Cache `instance`, `default_memory`, `default_table` (and `default_global`) outside of the call stack.
    - **Note:** This way we can cache the default instances easily and without call overhead across function calls. Also function frames on the call stack take less space and therefore the calling overhead should be reduced.
    - **Result:** This gave speed-ups for our real world test cases in the range of 5-8% but significantly regressed performance of our artificial `host_calls` test case by 22%. It is not yet clear why that is but we will leave this optimization in place until further notice.
- [x] Reuse the `pc` from the `StackFrameView` instead of having a local copy in the instruction executor.
    - **Note:** This actually regressed performance quite a bit so we did not push this change.
- [x] Introduce non-copying branch instructions `br`, `br_nez`, `br_eqz` and `return_nez`. Currently we only have copying branch instructions for the Wasm `multi-value` proposal. However, in many cases this optional copy overhead is not required and can be avoided at `wasmi` bytecode generation.
    - **Note:** Benchmarks shows improvements between 5-10% for most execution benchmarks. Not quite what we expected but still an improvement.
- [x] Experiment with `ExecOutcome::Continue { next_pc: usize }` as unification of `ExecOutcome::Continue` and `ExecOutcome::Branch { next_pc: usize }`.
    - **Note:** Experiment yieled major performance improvements from 10%-25% across the board. The experiment removed `ExecOutcome` entirely and inlined the visitor for executing the instructions into a giant `match` expression. This has had the benefit of returning immediately (without 2nd `match`) in case of `call`, `call_indirect`, `return` and `return_nez`.
- [x] Remove `LocalsRegistry` since its `ValueType` information store is no longer in use.
- [x] Currently the `wasmi` `br_table` instruction always executes in 2 cycles. First cycle is to select the correct `br` or `return` instruction that follows the `br_table` encoding scheme for `wasmi` bytecode. The 2nd cycle is the execution of the selected `br` or `return` instruction. This could be fused into a single cycle at the cost of having to include the `CodeMap` into the instruction execution context.
    - **Note:** Experiments showed that executing `br_table` in a single cycle created huge overhead for the rest of the execution engine which decreased efficiency of execution by roughly 10% as benchmarks indicated. So we reverted this change.
- [x] Use unchecked register reads and writes in `--release` mode.
    - **Note:** We can do this since all `wasmi` bytecode generated register accesses are valid by construction. This yielded roughly 5% improvements on some of our benchmarks.
- [x] Add more bytecode optimizations in order to simplify the bytecode itself.
    - **Example:** Replace `add(reg0, reg1, cref0)` with `sub(reg0, reg1, cref1)` where `cref1` is equal to `-cref0`. This way we can replace all `add` where the 2nd operand is a constant, thereby turning both inputs into the more efficient `ExecRegister`.
    - **Note**: This is no longer needed or useful due to the existence of the new `Add` and `AddImm` instructions.

## Follow-up PRs

- [x] Create `UntypedValue` in `wasmi_core` which provides an untyped API surface for all Wasm operations.
    - Note: This has become useful since now not only the engine executing the `wasmi` bytecode has to execute those operations but also parts of the compilation process when faced with constant values that allow for precomputation during compilation time. This way we can deduplicate code and also make it available in `wasmi_core` as a general API.
    - https://github.com/paritytech/wasmi/pull/369


paritytech-cicd-pr:
## CRITERION BENCHMARKS ## 

 |BENCHMARK|MASTER|PR|Diff|
 |---|---|---|---|
 | `compile_and_validate_v1`  | 5.4826 ms  | 6.9725 ms  | :red_circle:  +27.222% |
| `execute_count_until_v1`  | 2.0229 ms  | 1.4791 ms  | :green_circle:  -26.874% |
| `execute_factorial_iterative_v1`  | 919.43 ns  | 521.09 ns  | :green_circle:  -43.310% |
| `execute_factorial_recursive_v1`  |  1.2555 µs   |  1.2789 µs   | :red_circle:  +1.9233% |
| `execute_fib_iterative_v1`  | 4.8004 ms  | 3.0601 ms  | :green_circle:  -36.252% |
| `execute_fib_recursive_v1`  | 11.220 ms  | 11.390 ms  | :red_circle:  +1.3889% |
| `execute_global_bump_v1`  | 2.8150 ms  | 2.9336 ms  | :red_circle:  +4.1942% |
| `execute_host_calls_v1`  |  31.313 µs   |  40.956 µs   | :red_circle:  +33.612% |
| `execute_memory_fill_v1`  | 4.0526 ms  | 2.8865 ms  | :green_circle:  -28.772% |
| `execute_memory_sum_v1`  | 3.8005 ms  | 2.6501 ms  | :green_circle:  -30.250% |
| `execute_memory_vec_add_v1`  | 8.0472 ms  | 7.3445 ms  | :green_circle:  -8.5245% |
| `execute_recursive_is_even_v1`  | 2.2379 ms  | 2.3583 ms  | :red_circle:  +5.3187% |
| `execute_recursive_ok_v1`  |  295.68 µs   |  288.03 µs   | :green_circle:  -2.5966% |
| `execute_recursive_scan_v1`  |  357.60 µs   |  379.74 µs   | :red_circle:  +6.2695% |
| `execute_recursive_trap_v1`  |  25.343 µs   |  28.505 µs   | :red_circle:  +12.233% |
| `execute_regex_redux_v1`  | 1.3947 ms  | 1.3053 ms  | :green_circle:  -6.3149% |
| `execute_rev_complement_v1`  | 1.3813 ms  | 1.1137 ms  | :green_circle:  -19.482% |
| `execute_tiny_keccak_v1`  | 1.2305 ms  |  875.87 µs   | :green_circle:  -28.734% |
| `execute_trunc_f2i_v1`  | 1.8089 ms  | 1.2280 ms  | :green_circle:  -32.089% |
| `instantiate_v1`  |  74.136 µs   |  74.978 µs   | :white_circle:  +0.6958% |


 [Link to pipeline](https://gitlab.parity.io/parity/mirrors/wasmi/-/jobs/1853110)

PR #371, opened 2022-05-16
Make fuzz work

kpp:
For some reason fuzz didn't work:

```
WARNING: Failed to find function "__sanitizer_get_total_unique_coverage".
WARNING: Failed to find function "__sanitizer_reset_coverage".
INFO: Seed: 3860000992
ERROR: __sanitizer_reset_coverage is not defined. Exiting.
Did you use -fsanitize-coverage=... to build your code?
────────────────────────────────────────────────────────────────────────────────

Error: Fuzz target exited with exit status: 1
```

I fixed it and found bugs:

1)

```
thread '<unnamed>' panicked at 'assertion failed: `(left == right)`
  left: `true`,
 right: `false`', fuzz_targets/load_wabt.rs:10:5

Output of `std::fmt::Debug`:

	[0, 97, 115, 109, 1, 0, 0, 0, 12, 1, 11]
```	

Which means `wasmi` accepts a code which is not accepted by `wabt`. To try a .wasm file: `echo AGFzbQEAAAAMAQA= | base64 --decode > test.wasm`, this file is rejected with `000000a: error: invalid section code: 12\n` error. 

2)

```
thread '<unnamed>' panicked at 'assertion failed: `(left == right)`
  left: `false`,
 right: `true`', fuzz_targets/load_wasmparser.rs:16:5

Output of `std::fmt::Debug`:

	[0, 97, 115, 109, 1, 0, 0, 0, 12, 1, 8]
```

Which means `wasmi` accepts a code which is not accepted by `wasmparser`.

======

This divergence can be avoided with CI:

```
cargo +nightly fuzz run load_wabt -- -jobs=4 -max_total_time=10
cargo +nightly fuzz run load_wasmparser -- -jobs=4 -max_total_time=10
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,6 +103,14 @@ This tests both `wasmi` engines using all features available to them.
 It is recommended to test using `--release` since compiling and testing without optimizations
 usually is a lot slower compared to compiling and testing with optimizations.
 
+## Fuzzing
+
+Install `cargo-fuzz`: `$ cargo install cargo-fuzz`.
+
+List all the existing fuzz targets: `$ cargo +nightly fuzz list`.
+
+Run a fuzz target: `$ cargo +nightly run <target>`.
+
 ## Platforms
 
 Supported platforms are primarily Linux, MacOS, Windows and WebAssembly.
--- a/fuzz/Cargo.toml
+++ b/fuzz/Cargo.toml
@@ -9,14 +9,11 @@ publish = false
 cargo-fuzz = true
 
 [dependencies]
+libfuzzer-sys = "0.4"
 wasmi = { path = ".." }
 wabt = "0.9"
 wasmparser = "0.14.1"
 tempdir = "0.3.6"
-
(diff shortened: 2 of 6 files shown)


Robbepop:
Thank you @kpp  for fixing fuzz testing for `wasmi`.

I suppose this tests the only `wasmi` `v0` version and not the newer `v1` versions found under `wasmi_v1` which is an entirely new engine, an entirely new API and its parsing and validation infrastructure is based on `wasmparser`.
Still having a working fuzzing system is an improvement. In the long run we want to fuzz `wasmi_v1`.

Robbepop (approved):
LGTM

Robbepop:
Thank you for the fix. I will merge this PR now.

However, we are currently phasing out the old legacy `wasmi` engine for which these fuzz tests have been creating. We are going to implement new fuzz tests for the new engines using Wasmtime infrastructure.

PR #372, opened 2022-05-19
Return NotResumable from resume_execution

fabiosantoscode:
I was looking through this repository and how the resumable execution mechanism works, and I noticed that resume_execution was returning the wrong error.

If this is not the correct fix and indeed AlreadyStarted is correct, the documentation should be changed instead.

### DIFF
--- a/src/func.rs
+++ b/src/func.rs
@@ -340,7 +340,7 @@ impl<'args> FuncInvocation<'args> {
                 if interpreter.state().is_resumable() {
                     Ok(interpreter.resume_execution(return_val, externals)?)
                 } else {
-                    Err(ResumableError::AlreadyStarted)
+                    Err(ResumableError::NotResumable)
                 }
             }
             FuncInvocationKind::Host { .. } => Err(ResumableError::NotResumable),


Robbepop (approved):
LGTM

> If this is not the correct fix and indeed AlreadyStarted is correct, the documentation should be changed instead.

My own knowledge ends here but it really looks odd. Thanks for the fix.

fabiosantoscode:
By the way, I discovered this while trying to make this VM serializable.

I have a thing for VMs that can be frozen to disk and continued later. The use cases are niche but giving this capability to a WASM machine can enable lots of cool use cases like hibernating long running compute tasks.

If you'd like I can contribute a PR that adds serde serialization to the necessary structs. I've done a bit of this on my end already.

fabiosantoscode:
Oops, accidentally closed.

Robbepop:
Do you have an urgent use case for this? If so, what is it if I may ask? :)
You filed the PR so far on our old implementation of `wasmi`. We are currently working on the so-called `v1` implementation.
If we think that serializable `wasmi` is something we want to support, we should probably implement this feature on the `v1` implementation.

fabiosantoscode:
There's really no use case at all. I've just been experimenting for fun :) 

I can imagine use cases but that doesn't count.

Robbepop:
> There's really no use case at all. I've just been experimenting for fun :)
> 
> I can imagine use cases but that doesn't count.

I'd prefer to introduce features to `wasmi` (thereby extending its codebase that needs to be maintained) for features that actually have a strong use case.

fabiosantoscode:
Entirely understandable :)

Robbepop:
I am sorry for coming back to this so late but I am hesistant to merge this PR even though it is probably more correct behavior because I do not want to break any user code and also because we are currently phasing out this legacy `wasmi` engine anyways and replace it with the newly developed engines.

PR #384, opened 2022-07-18
feat(vm): avoid host error subtyping with generics

hussein-aitlahcen:
This PR get rid of the `HostError` subtyping. We simply introduce a new associated type `Error` in  `Externals` and build a hierarchy of errors. Instead of trying to embed the host error in the interpreter error, we inverse the dependency. The interpreter will yield a user error that can embed the interpreter error via `From`.

The code get simplified and [the user is now able to have `Copy, Clone, PartialEq, Eq` on the error type](https://github.com/paritytech/wasmi/pull/384/files#diff-db9430b169cf94ee9f6edc769280ae0d221f06f379421eee7ccf9256a2b2c2acL310-R310), hence also simplifying the code calling wasmi (no more ugly `Box<&dyn ...`).
By default, the user is able to use `wasmi::Error` as it can embed itself. A small drawback is that the `NopExternals` now require a `PhantomData` to track the actual error. Free from the perspective of performance though.

Tests/doctests have been updated accordingly.


hussein-aitlahcen:
> Thank you for the PR!
> 
> Can you please use proper Rust code formatting using `cargo +nightly fmt` in order to pass our CI and to make the diff readable for reviewing? :) Thanks! Seems like the `clippy` and `doc` CI passes are also unhappy atm.

Thanks :). Hopefully fixed everything!

hussein-aitlahcen:
> I have not grasped the entire PR scope but I was wondering how the new `v1` engine in `wasmi_v1` folder is now supposed to return a host error. For the `v0` engine in `src` you extended the `Externals` with an `Error` associated type but I am missing a replacement functionality for the non-legacy engines.

Oops didn't noticed I completely erased this part, we don't have any test with custom error for the `v1`? Let me push a solution for the `v1` as well.

Robbepop:
May I ask: what is the purpose or use case for which you require this change of `wasmi`?

Robbepop:
With `v1` we are trying to mirror [this Wasmtime API](https://docs.rs/wasmtime/0.38.1/wasmtime/) as good as we can.

Robbepop:
> 

But is there a use case on your side that requires this feature? Since the original `wasmi` package is considered a legacy software that we are no longer actively working on while at the same time having tons of dependencies we are kinda hesistant to merge breaking changes unless there is good reasoning or a critical use case for it.

src/host.rs
```diff
@@ -146,35 +148,40 @@ impl<'a> RuntimeArgs<'a> {
 /// }
 /// ```
 pub trait Externals {
+    /// Type of error the host can yield while executing a host function.
+    type Error: CanResume + From<Error>;
```
Robbepop, line 152: You might also want to have `Into<Error>` so that we kinda guarantee that we can carve out a `wasmi::Error` (trap) from a user error.

hussein-aitlahcen: Unless I am missing something, [it look like the current code is very permissive as any host error make the interpreter resumable?](https://github.com/paritytech/wasmi/pull/384/files#diff-68c6f449503a33c74a15800c0a9c8fd34028465a06edcac92b32bbb9db64bbceL328-R330)

hussein-aitlahcen: For the `Into<Error>`, it's impossible as the host error is variant containing the trap, i.e. you can inject the trap but the projection is not always possible. We could have a `TryInto` but it's not used anywhere in the code right now? Did I understood what you meant?

hussein-aitlahcen:
> Is there a use case on your side that requires this feature? I am asking since the original `wasmi` package is considered a legacy software that we are no longer actively working on while at the same time having tons of dependencies we are kinda hesistant to merge breaking changes unless there is good reasoning or a critical use case for it.

Alright, we are implementing a CosmWasm contract pallet at Composable and would like to use wasmi (as it's the only `no_std` friendly interpreter as of right now?) as underlying interpreter for the VM. The VM is abstract and we can plug almost any interpreter, so if you believe that it's better to leave this patch in our fork right now until `v1` is ready, we are fine with that :). Do you have an ETA for the `v1` release by the way? 

Other than that, the purpose of the patch is to simplify the embedding of the interpreter. It look very strange that a module of an application embed the application error itself. Pretty much like saying that a library hold your custom, specific error while it should in fact be as general as possible and agnostic to the application using the library.

dzmitry-lahoda:
@Robbepop there are several wasm runtimes. are they support compatiblity? example cosmwasm. also, is compatiblity is needed for what? stop development of wasmi and make use of other wasm? or making some performance measurments by temporal drop ins? 

if feels like making good runtime is hard task, making compabile api is easy. may be can do it when it needed and push best in class runtime with best possible hosting api?

src/host.rs
Robbepop: Yeah I see so the best we can have is an optional mapping from `HostError` to `wasmi::Error`, like `Into<Option<Error>>`. `TryInto` is not really good because we actually do not need an error variant there imo. `Into<Option<Error>>` looks very verbose on the other hand.

Robbepop:
> I just looked what the Wasmtime runtime uses for their host errors and it seems to be the very popular and simple [anyhow crate](https://crates.io/crates/anyhow/). So maybe we can improve the situation a lot for you guys by simply using the anyhow crate and [its Error type](https://docs.rs/anyhow/1.0.58/anyhow/struct.Error.html) as well. Could you test if that would suit you, too?

@hussein-aitlahcen Have you had time to check if `anyhow` could work for you, too? It is a very popular crate for handling Rust errors.

Robbepop:
Closing this due to inactivity of the PR author. @hussein-aitlahcen Please feel free to reopen once you got the time to come back to https://github.com/paritytech/wasmi/pull/384#issuecomment-1194642886.

Issue #374, opened 2022-06-21
Test for no_std availability in builds tests


chrysn:
The build tests that currently run as "no_std" do not check what they should check for -- that the project can actually be built without std: They do disable the std feature from the present crates by adding `--no-default-features`, but any crate that is no_std can just pull in std, and the CI would be none the wiser.

Unless there is a better way to detect (and fail on) the presence of std, a simple way to perform the test better is to run the no_std build with `--target thumbv7em-none-eabihf`, for which (unlike the wasm32-unknown-unknown) there is no std library available, and thus depending transitively on std does cause a hard build failure.

Fixing the test alone will likely uncover issues down the dependency chain -- my current attempts to build showed numerous errors in core/src/nan_preserving_float.rs (maybe also triggered by the presence of an own crate named "core"?), and the spin and wasmparser-nostd crates. (Currently you'd probably only ever want to look at the first error produced; the latest compilers regressed in the amount of follow-up errors they print).

hussein-aitlahcen:
There is a missing `#![no_std]` in `wasmi_core` :)

hussein-aitlahcen:
cc @Robbepop

Robbepop:
Oh wow!
I actually had to check the codebase locally to make sure that this is real.
Thanks for the fix!

hussein-aitlahcen:
@Robbepop we can close this :)

chrysn:
If you want to automate testing this, a line of `cargo +nightly build --target thumbv7em-none-eabi --no-default-features --lib` or similar (main point: the "none" is a pretty strong indication that there won't be a std for this target) helps enforcing this property in CI.

Robbepop:
@chrysn can you file a PR for this?

PR #386, opened 2022-07-19
missing `#![no_std]` in core leading to std

hussein-aitlahcen:


hussein-aitlahcen:
fixes #374

Robbepop:
Oh wow!
I actually had to check the codebase locally to make sure that this is real.
Thanks for the fix!

core/src/lib.rs
```diff
@@ -1,3 +1,5 @@
+#![no_std]
```
Robbepop, line 1:
```suggestion
#![cfg_attr(not(feature = "std"), no_std)]
```

Robbepop: We don't want to enable `no_std` when the `std` feature is enabled as with all crates.

hussein-aitlahcen: indeed

PR #387, opened 2022-07-21
CI: Check that the no_std build is built without std

chrysn:
... by placing it on a target without an operating system.

The test's scope is reduced from --workspace to --lib, as only that
component (and its transitive dependencies) are expected to be no_std.

Closes: https://github.com/paritytech/wasmi/issues/374

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -16,8 +16,12 @@ jobs:
         with:
           profile: minimal
           toolchain: stable
-          target: wasm32-unknown-unknown, thumbv7em-none-eabi
+          target: wasm32-unknown-unknown
           override: true
+      - name: Add extra targets
+        # Workaround for https://github.com/actions-rs/toolchain/issues/165
+        run: |
+          rustup target add thumbv7em-none-eabi
       - name: Build (default features)
         uses: actions-rs/cargo@v1
         with:
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -32,7 +32,7 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: build
-          args: --lib --no-default-features --target thumbv7em-none-eabi
+          args: --workspace --lib --no-default-features --target thumbv7em-none-eabi --exclude wasmi_cli
       - name: Build (wasm32)
         uses: actions-rs/cargo@v1
         with:
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -16,7 +16,7 @@ jobs:
         with:
           profile: minimal
           toolchain: stable
-          target: wasm32-unknown-unknown
+          target: wasm32-unknown-unknown, thumbv7em-none-eabi
           override: true
       - name: Build (default features)
         uses: actions-rs/cargo@v1
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -28,11 +28,11 @@ jobs:
         with:
           command: build
           args: --workspace --all-features
-      - name: Build (no_std)
+      - name: Build wasmi itself as no_std
         uses: actions-rs/cargo@v1
         with:
           command: build
-          args: --workspace --no-default-features
+          args: --lib --no-default-features --target thumbv7em-none-eabi
       - name: Build (wasm32)
         uses: actions-rs/cargo@v1
         with:
--- a/wasmi_v1/src/engine/func_builder/control_stack.rs
+++ b/wasmi_v1/src/engine/func_builder/control_stack.rs
@@ -1,5 +1,4 @@
 use alloc::vec::Vec;
-
 use super::ControlFrame;
 
 /// The stack of control flow frames.
--- a/wasmi_v1/src/engine/func_builder/control_stack.rs
+++ b/wasmi_v1/src/engine/func_builder/control_stack.rs
@@ -1,3 +1,5 @@
+use alloc::vec::Vec;
+
 use super::ControlFrame;
 
 /// The stack of control flow frames.
--- a/wasmi_v1/src/engine/func_builder/control_stack.rs
+++ b/wasmi_v1/src/engine/func_builder/control_stack.rs
@@ -1,5 +1,5 @@
-use alloc::vec::Vec;
 use super::ControlFrame;
+use alloc::vec::Vec;
 
 /// The stack of control flow frames.
 #[derive(Debug, Default)]
--- a/wasmi_v1/src/engine/func_builder/value_stack.rs
+++ b/wasmi_v1/src/engine/func_builder/value_stack.rs
@@ -1,6 +1,5 @@
 use alloc::vec::Vec;
 use core::cmp;
-
 use wasmi_core::ValueType;
 
 /// The value stack that is emulated during Wasm to `wasmi` bytecode translation.
--- a/wasmi_v1/src/engine/func_builder/value_stack.rs
+++ b/wasmi_v1/src/engine/func_builder/value_stack.rs
@@ -1,3 +1,4 @@
+use alloc::vec::Vec;
 use core::cmp;
 
 use wasmi_core::ValueType;
--- a/wasmi_v1/src/lib.rs
+++ b/wasmi_v1/src/lib.rs
@@ -3,6 +3,7 @@
 //! These closely mirror the WebAssembly specification definitions.
 //! The overall structure is heavily inspired by the `wasmtime` virtual
 //! machine architecture.
+#![cfg_attr(not(feature = "std"), no_std)]
 
 #[cfg(not(feature = "std"))]
 #[macro_use]
--- a/wasmi_v1/src/module/builder.rs
+++ b/wasmi_v1/src/module/builder.rs
@@ -1,4 +1,3 @@
-use alloc::vec::Vec;
 use super::{
     import::FuncTypeIdx,
     DataSegment,
@@ -22,6 +21,7 @@ use crate::{
     ModuleError,
     TableType,
 };
+use alloc::vec::Vec;
 
 /// A builder for a WebAssembly [`Module`].
 #[derive(Debug)]
--- a/wasmi_v1/src/module/builder.rs
+++ b/wasmi_v1/src/module/builder.rs
@@ -1,3 +1,5 @@
+use alloc::vec::Vec;
+
 use super::{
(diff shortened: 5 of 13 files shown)


.github/workflows/rust.yml
```diff
@@ -28,11 +28,11 @@ jobs:
         with:
           command: build
           args: --workspace --all-features
-      - name: Build (no_std)
+      - name: Build wasmi itself as no_std
         uses: actions-rs/cargo@v1
         with:
           command: build
-          args: --workspace --no-default-features
+          args: --lib --no-default-features --target thumbv7em-none-eabi
```
Robbepop, line 35: We cannot use `--lib` here because this would only build the legacy `wasmi` engine but not the newer `wasmi_v1`.

Robbepop (changes requested):

Robbepop:
> ... by placing it on a target without an operating system.
> 
> The test's scope is reduced from --workspace to --lib, as only that component (and its transitive dependencies) are expected to be no_std.
> 
> Closes: #374

Thanks a lot for opening the PR!

.github/workflows/rust.yml
Robbepop: `--workspace --exclude wasmi_cli` should work instead

Robbepop: However, it seems that the `thumbv7em-none-eabi` is not only missing the `std` library but also even the `alloc` and `core` libraries. I just ran this locally and saw some weird error messages such as:
```
error[E0405]: cannot find trait `Copy` in this scope
   --> /home/.cargo/registry/src/github.com-1ecc6299db9ec823/num-traits-0.2.15/src/lib.rs:617:31
    |
617 |     fn compute<T: NumAssign + Copy>(mut x: T, y: T) -> T {
    |                               ^^^^ not found in this scope
```

chrysn: It'll still need some limiting to libraries as the src/bin/instantiate.rs is clearly std dependent. As for the alloc and core, that I can probably add via the toolchain action.

chrysn: Looks all good now as far as I can tell, please clear for CI run again. (Adding two targets is a bit weird because the actions docs don't say it's an array, but it seems to work that way with other properties of the action, so meh, probably it works).

chrysn:
The target installation did *not* take as intended; added a workaround and fixed the other failing steps. If you're happy with the PR, I can squash it to get rid of the squashes / fixups.

wasmi_v1/src/engine/func_builder/control_stack.rs
```diff
@@ -1,3 +1,5 @@
+use alloc::vec::Vec;
+
 use super::ControlFrame;
```
Robbepop, line 3: Would you be so kind to merge all the `use` statements in all the changed files? This way `rustfmt` can automatically merge and sort them but not if there is whitespace in between.

Robbepop, line 3:
```suggestion
use alloc::vec::Vec;
use super::ControlFrame;
```

wasmi_v1/src/engine/func_builder/value_stack.rs
```diff
@@ -1,3 +1,4 @@
+use alloc::vec::Vec;
 use core::cmp;
 
 use wasmi_core::ValueType;
```
Robbepop, line 4:
```suggestion
use alloc::vec::Vec;
use core::cmp;
use wasmi_core::ValueType;
```

wasmi_v1/src/module/builder.rs
```diff
@@ -1,3 +1,5 @@
+use alloc::vec::Vec;
+
 use super::{
```
Robbepop, line 3:
```suggestion
use alloc::vec::Vec;
use super::{
```

wasmi_v1/src/module/element.rs
```diff
@@ -1,3 +1,5 @@
+use alloc::{boxed::Box, vec::Vec};
+
 use crate::ModuleError;
```
Robbepop, line 3:
```suggestion
use alloc::{boxed::Box, vec::Vec};
use crate::ModuleError;
```

wasmi_v1/src/module/error.rs
```diff
@@ -1,3 +1,5 @@
+use alloc::string::String;
+
 use super::ReadError;
```
Robbepop, line 3:
```suggestion
use alloc::string::String;
use super::ReadError;
```

wasmi_v1/src/module/export.rs
```diff
@@ -1,3 +1,5 @@
+use alloc::boxed::Box;
+
 use super::GlobalIdx;
```
Robbepop, line 3:
```suggestion
use alloc::boxed::Box;
use super::GlobalIdx;
```

wasmi_v1/src/module/import.rs
```diff
@@ -1,3 +1,4 @@
+use alloc::boxed::Box;
 use core::fmt::{self, Display};
 
 use crate::{GlobalType, MemoryType, ModuleError, TableType};
```
Robbepop, line 4:
```suggestion
use alloc::boxed::Box;
use core::fmt::{self, Display};
use crate::{GlobalType, MemoryType, ModuleError, TableType};
```

wasmi_v1/src/module/parser.rs
```diff
@@ -1,3 +1,5 @@
+use alloc::vec::Vec;
+
 use super::{
```
Robbepop, line 3:
```suggestion
use alloc::vec::Vec;
use super::{
```

.github/workflows/rust.yml
```diff
@@ -18,6 +18,10 @@ jobs:
           toolchain: stable
           target: wasm32-unknown-unknown
           override: true
+      - name: Add extra targets
+        # Workaround for https://github.com/actions-rs/toolchain/issues/165
+        run: |
+          rustup target add thumbv7em-none-eabi
```
Robbepop, line 24: Is there no better way to do this? I do not get a lot of info out of the linked issue tbh.

Robbepop:
> The target installation did _not_ take as intended; added a workaround and fixed the other failing steps. If you're happy with the PR, I can squash it to get rid of the squashes / fixups.

Thank you for all the work. Manual squashing won't be needed since we will squash this upon merge anyways.

wasmi_v1/src/lib.rs
```diff
@@ -3,6 +3,7 @@
 //! These closely mirror the WebAssembly specification definitions.
 //! The overall structure is heavily inspired by the `wasmtime` virtual
 //! machine architecture.
+#![cfg_attr(not(feature = "std"), no_std)]
```
Robbepop, line 6: I cannot believe we forgot about this all the time in `wasmi_v1` ...

chrysn:
>Is there no better way to do this? I do not get a lot of info out of the linked issue tbh.

The information on that is not so much in the issue's text but in its metadata: it's been around for a year, and nobody ever claimed that it's a non-issue. (I tested whether lists work in that place earlier -- they don't). Given that the whole thing appears to be unmaintained, it's either the simple workaround or larger CI overhauls.

wasmi_v1/src/engine/func_builder/control_stack.rs
chrysn: Done based on your suggestions, with a final cleanup `cargo fmt` on top.

The old sorting was based on some Python habits which I deem rustworthy: imports grouped into paragraphs of standard libraries dependencies and internals.

Robbepop (approved):
LGTM

wasmi_v1/src/lib.rs
chrysn: For something that was not tested on no_std it held up remarkably well :-)

(Then again, now that there is alloc, it has become easier ... an alloc-less version would be really great to have, but seems not to be trivial from what I've seen when inserting the `alloc::`s).

Robbepop: What use cases do you have for an alloc less version? The difference between alloc and no alloc is a global allocator which could be as simple as a trivial bump allocator such as bumpalo.

Robbepop:
Thanks a lot for this PR @chrysn !

wasmi_v1/src/lib.rs
chrysn: Better predictability on an embedded system -- plus it's hard to determine whether a plain bump allocator is really sufficient for a project like this or not, and a good allocator is costly on these systems.

Predictability is especially a concern when there are multiple components "like this" on a device. While it's only one (and if it's really "bump allocator friendly"), I can pick a simple allocator and worst that happens is that my WASM gets stuck out of memory. But once I have, say, a WASM and a Bluetooth stack that "only" need a simple allocator, suddenly a rogue WASM program can lock me out of Bluetooth. (Same can go for two WASMI instances, when one is performing administrative and another user provided tasks).

I think that for some uses of alloc in here it might even be simple to solve (eg. by compatibly turning `CallStack` into `CallStack<Frames: VecLikesh<FunctionFrame> = Vec<FunctionFrame>>`, which I'd then rather instanciate as `CallStack<heapless::Vec<FunctionFrame, 16>>` or so), but around the function maps or the WASM parser I have no clue whether the same would be practical, or (given there is a parser present) whether WASMI even makes sense on these platforms (on which I'd prefer to interpret WASM in-place or at least from something that gets built once from WASM input).

chrysn:
Thanks for reviewing and merging.

On a side note, re. squashing, even with the squash merge you're using, history would probably look better if I had squashed this in git (`git rebase --autosquash`) before, as that'd automatically have removed and reassigned all the fixups and squashes that are now listed long and detailed in the commit message.

wasmi_v1/src/lib.rs
Robbepop: Thanks for the clarification.
However, in all honesty I do not plan on supporting `wasmi` without `alloc` since that would be an absolut epic amount of additional work. Instead I'd really try to use a suiting heap allocator and maybe make use of Wasm `multiple-memories` once it has been standardized so that different Wasm instances do not compete for linear memory.

chrysn:
>Would you be so kind to merge all the `use` statements in all the changed files? This way `rustfmt` can automatically merge and sort them but not if there is whitespace in between.
Can do when back at PC. Sorted them that way from Python habits I deem rustworthy: imports grouped into paras of stdlib, dependencies and internals.

PR #393, opened 2022-07-24
Remove legacy wasmi and all of its components from the repo

Robbepop:
This also turns the workspace into a virtual workspace.


Robbepop:
@chrysn : by any chance do you see what is wrong with the `no_std` build according to the CI? I totally do not see where the `std` is indirectly used through the changes by this PR.

PR #407, opened 2022-08-13
Fix the read implementation on no-std

AldaronLau:
Currently behavior is panicking.

### DIFF
--- a/wasmi_v1/src/module/read.rs
+++ b/wasmi_v1/src/module/read.rs
@@ -1,8 +1,5 @@
 use core::{fmt, fmt::Display};
 
-#[cfg(not(feature = "std"))]
-use core::cmp;
-
 #[cfg(feature = "std")]
 use std::io;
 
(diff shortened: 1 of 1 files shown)


Robbepop (approved):
LGTM

Thanks a lot for the bug fix. How have you found this bu if I may ask?

AldaronLau:
@Robbepop I found the bug by trying to convert the example in lib.rs to work on no-std, as I was upgrading one of my projects from an older version of wasmi.  First, I had to get rid of the `&mut` from the call to `Module::new()` - then it compiled.  After that, I tried loading my WASM file, and I got an unexpected panic.

Robbepop:
@AldaronLau btw the new `wasmi` engine is still pretty new and therefore not yet battle hardened. We are currently in the process to change that. As of right now I would not recommend using the newest `wasmi` in production.

AldaronLau:
@Robbepop thanks for the warning.  I don't think my project counts a "production" (it's very experimental), so I'll keep testing it.

Robbepop:
@AldaronLau Do you have a link to the experimental project for me?

AldaronLau:
The project is https://github.com/ardaku/quantii, which pulls in https://github.com/ardaku/ardaku, which I was updating to the new version of wasmi.

Issue #435, opened 2022-09-12
Unexppected NaN


erxiaozhou:
# Build commands
I compile the code with commit id 3b29737f7aa5faaeb0c14cd38c4dd1339e72f597.
Platform: Ubuntu 20.04
CPU: amd64
compile:
```shell
cargo build
```
execute:

```shell
wasmi_cli cp907_b3.wasm _start
```
[cp907_b3.zip](https://github.com/paritytech/wasmi/files/9548016/cp907_b3.zip)

# Expected behavior:
It seems that the output should be -nan, with a value 0xffcefef4  left on the stack

# Actual behavior:
wasmi output:
```
NaN
```
and the value left on the stack is 0xff8efef4

Thank you.

Robbepop:
Thanks a lot for reporting and providing the failing test input. I will take a look. What version of wasmi did you use?

erxiaozhou:
> Thanks a lot for reporting and providing the failing test input. I will take a look. What version of wasmi did you use?

Thank you for your reply. The commit id of tested wasmi is 3b29737f7aa5faaeb0c14cd38c4dd1339e72f597.

Robbepop:
Minified `.wasm` test file:
```wat
(module
  (type (;0;) (func (result f32)))
  (func (;2;) (type 0) (result f32)
    f32.const -nan:0xefef4 (;=-nan;)
    f32.const -0x1.44ceeep+127 (;=-2.15872e+38;)
    f32.copysign
  )
  (export "_start" (func 0))
)
```

erxiaozhou:
> Minified `.wasm` test file:
> 
> ```webassembly
> (module
>   (func (export "_start") (param) (result f32)
>     (f32.copysign
>         (f32.const -nan:0xefef4)
>         (f32.const -1)
>     )
>   )
> )
> ```
> 
> Currently trying to understand what is going on. So far it seems to me that `wasmi`'s `copysign` implementation behaves correctly. Why are you expecting the whole run to return `0xFFCEFEF4` instead of `0xFF8EFEF4`?
> 
> Given that:
> 
> * `(f32.const -nan:0xefef4)` translates to `0xFF8EFEF4`
> * `(f32.const nan:0xefef4)` translates to `0x7F8EFEF4`
> 
> Note that the first bit is the sign bit.
> 
> What I found so far is that Rust prints `"NAN"` for both `f32::NAN` as well as `-f32::NAN` which is confusing.
> 
[...]

PR #436, opened 2022-09-12
Improve `F32` and `F64` `Display` impl for -NaN values

Robbepop:
Result of confusing in https://github.com/paritytech/wasmi/issues/435.

### DIFF
--- a/core/src/nan_preserving_float.rs
+++ b/core/src/nan_preserving_float.rs
@@ -153,10 +153,16 @@ macro_rules! float {
 
         impl ::core::fmt::Display for $for {
             fn fmt(&self, f: &mut ::core::fmt::Formatter) -> ::core::fmt::Result {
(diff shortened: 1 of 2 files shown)


Robbepop:
I asked in Rust Discord channel why Rust prints `"NaN"` for both `+NaN` and `-NaN`.
Someone answered that `-NaN` simply does not exist and has no semantic meaning and therefore Rust does not care to print this additional information. Information such as the sign-bit or the mantissa of a NaN can be extracted by converting the `f32` to its underlying `u32` bits.

Therefore I am sceptical if we want to merge this PR tbh ...

2022-09-12-162512_1502x713_scrot

Robbepop:
I decided for now that we do not want to go a special route with displaying `-NaN` unlike the rest of the ecosystem.

PR #449, opened 2022-09-18
Overhaul Pages and Bytes types

Robbepop:
Problems with previous types were found during fuzzy testing.
These types make it simpler to fix linear memory allocation problems in `wasmi` for 32-bit and 64-bit architectures for edge cases.

Also we get rid of the `memory_units` dependency which did a lot more things that we actually needed in `wasmi`.

PR #452, opened 2022-09-19
Support allocating 4GB of memory

bkchr:


core/src/vmem.rs
```diff
@@ -55,7 +55,7 @@ impl Debug for VirtualMemory {
 
 impl VirtualMemory {
     /// The maximum allocation size for a `wasmi` virtual memory.
-    const MAX_ALLOCATION_SIZE: usize = u32::MAX as usize;
+    const MAX_ALLOCATION_SIZE: usize = u32::MAX as usize + 1;
```
Robbepop, line 58: This breaks on 32-bit architectures since `usize == u32`. Therefore we need to use `u64` here to cover for this case OR make a distinction between 32-bit and 64-bit architectures were 32-bit architectures can only allocate at most 2GB.

For newer `wasmi` versions I already have a draft PR in the pipeline:
https://github.com/paritytech/wasmi/pull/449

bkchr: Good point :facepalm:

bkchr: Fixes

src/memory/mmap_bytebuf.rs
```diff
@@ -23,7 +23,8 @@ impl ByteBuf {
     /// In this implementation we won't reallocate the virtually allocated
     /// buffer and instead simply adjust the `len` field of the `ByteBuf`
     /// wrapper in order to efficiently grow the virtual memory.
-    const ALLOCATION_SIZE: usize = u32::MAX as usize;
+    const ALLOCATION_SIZE: usize =
+        validation::LINEAR_MEMORY_MAX_PAGES as usize * super::LINEAR_MEMORY_PAGE_SIZE.0;
```
bkchr, line 27: This is already guarded by `target-architecture = 64`

core/src/vmem.rs
```diff
@@ -55,7 +55,11 @@ impl Debug for VirtualMemory {
 
 impl VirtualMemory {
     /// The maximum allocation size for a `wasmi` virtual memory.
-    const MAX_ALLOCATION_SIZE: usize = u32::MAX as usize + 1;
+    #[cfg(target_pointer_width = "64")]
+    const MAX_ALLOCATION_SIZE: usize = 2 << 32;
```
Robbepop, line 59: `2 << 32` is not equal to `u32::MAX as usize + 1` but double that sice, therefore 8GB.
I'd propose to use `u32::MAX as usize + 1` instead since that's more clear.

How about this:

```rust
/// The maximum allocation size for a `wasmi` virtual memory on 16-bit.
#[cfg(target_pointer_width = "16")]
const MAX_ALLOCATION_SIZE: usize = compile_error!(
    "16-bit architectures are current unsupported by wasmi"
);

/// The maximum allocation size for a `wasmi` virtual memory on 32-bit.
#[cfg(target_pointer_width = "32")]
const MAX_ALLOCATION_SIZE: usize = i32::MAX as u32 as usize; // 2GB

/// The maximum allocation size for a `wasmi` virtual memory on 64-bit.
#[cfg(target_pointer_width = "64")]
const MAX_ALLOCATION_SIZE: usize = u32::MAX as usize + 1; // 4GB
```

With the current code and spec Wasm on 16-bit platforms is simply impossible to support properly.

Robbepop: It can actually be even `i32::MAX as usize + 1`. :)

I updated the code snippet above.

Robbepop (approved):
LGTM

core/src/vmem.rs
```diff
@@ -54,8 +54,18 @@ impl Debug for VirtualMemory {
...
-    /// The maximum allocation size for a `wasmi` virtual memory.
-    const MAX_ALLOCATION_SIZE: usize = u32::MAX as usize;
+    /// The maximum allocation size for a `wasmi` virtual memory on 16-bit.
+    #[cfg(target_pointer_width = "16")]
+    const MAX_ALLOCATION_SIZE: usize =
+        compile_error!("16-bit architectures are current unsupported by wasmi");
+
+    /// The maximum allocation size for a `wasmi` virtual memory on 32-bit.
+    #[cfg(target_pointer_width = "32")]
+    const MAX_ALLOCATION_SIZE: usize = i32::MAX as u32 as usize + 1; // 2GB
```
bkchr, line 64:
```suggestion
    const MAX_ALLOCATION_SIZE: usize = i32::MAX as usize + 1; // 2GB
```

PR #492, opened 2022-10-02
[WIP] Optimize jumps that keep zero elements on the stack

Robbepop:
This is a result of an observation that we could speed-up execution of branch and return instructions that keep no values upon jump.


Robbepop:
Closed in favor of https://github.com/paritytech/wasmi/pull/493.

PR #508, opened 2022-10-12
refactor: improve call wasm

yjhmelody:
- Make the logic more clean and clear.
<del>- check `recursion_limit` by `>=`</del>


crates/wasmi/src/engine/stack/frames.rs
```diff
@@ -72,20 +71,13 @@ impl CallStack {
...
-    pub(crate) fn push(
-        &mut self,
-        caller: &mut FuncFrame,
-        ip: InstructionPtr,
-        instance: Instance,
-    ) -> Result<FuncFrame, TrapCode> {
-        if self.len() == self.recursion_limit {
+    /// Pushes a Wasm caller function onto the [`CallStack`].
+    pub(crate) fn push(&mut self, caller: FuncFrame) -> Result<(), TrapCode> {
+        if self.len() >= self.recursion_limit {
```
Robbepop, line 76:
```suggestion
        if self.len() == self.recursion_limit {
```

Robbepop: Please revert this. The `>=` is less precise with respect to the implied invariants since this number cannot possibly be larger than `self.recursion_limit` given that the code is bug free. So testing for `>=` in this position no longer properly communicated this invariant. Also it may not uncover possible bugs.

Robbepop (changes requested):
Please revert `>=` change, rest of the PR looks fine to me. :)

crates/wasmi/src/engine/stack/frames.rs
yjhmelody: Ok

Robbepop:
I checked out your PR locally after seeing some regressions according to [our benchmarking CI](https://github.com/paritytech/wasmi/pull/508#issuecomment-1275727166) and unfortunately I can confirm locally that some benchmarks are slightly regressed with those changes.
I see that they are some kind of a clean up but I am hesistant merging a change if it introduces regressions that I do not understand.

yjhmelody:
@Robbepop I do not see regressions in my local

yjhmelody:
:( execute/memory_vec_add meet a lot regressions. Why?

Robbepop:
> :( execute/memory_vec_add meet a lot regressions. Why?

I don't know. Could be an artifact. Unfortunately `wasmi` these days is extremely sensible to performance regressions.

yjhmelody:
Why is red when improved? Such as 🔴 -3.70% 🔴 -0.35%

yjhmelody:
Strange, wasm_kernel is not stable, I see a big improvement last time.

yjhmelody:
@Robbepop  It now seems is a good improvement that could be merged now.

Robbepop:
> @Robbepop It now seems is a good improvement that could be merged now.

The problem with those benchmarks is that they are currently extremely flaky/noisy for this PR in particular. I don't know why that is but we cannot really trust those benchmarks fully with respect to this PR. I will take a look today on it and decide if we want to merge it.

Robbepop (approved):
I checked out benchmarks locally for this PR and it now seems to be fine. LGTM therefore.

PR #553, opened 2022-11-04
Inline conversion routines of the `UntypedValue` API

Robbepop:


Robbepop:
Closed since this PR introduces regressions that outweight the gains unfortunately.

Issue #572, opened 2022-11-21
Allow called host functions to execute Wasm functions via the `wasmi::Engine`


Robbepop:
As stated in https://github.com/paritytech/wasmi/issues/567 it is currently not possible in `wasmi` to call Wasm functions from within called host functions due to a deadlock in the `Engine`.
We designed it this way because it was a simple implementation and we thought it was done similarly in Wasmtime, however, we were wrong about that since in Wasmtime it is possible to call Wasm functions from within called host functions as stated by the issue author.

Since we want to ideally mirror the Wasmtime functionality we want to implement this feature as well. The only downside is that it is slightly harder to implement since `wasmi` is an interpreter that has to carry arround all the execution state with each call stack. Therefore we need to implement a way to allow for multiple executors for a single `wasmi::Engine`.

This issue tracks the progress of the feature.

mariaschett:
We also rely on calling Wasm functions from within host function. Any chance this issue is on the roadmap? Thank you!

tiziano88:
Another +1 to this issue, we also rely on a pattern similar to what described in https://github.com/paritytech/wasmi/issues/567#issue-1455408623 , in which the Wasm module invokes a host function, and during its execution the host calls back into the Wasm module to perform an allocation. This is not a super common use case, but I have already seen a similar pattern in other projects as well, so I hope wasmi can prioritize and implement this. Thanks in advance!

Robbepop:
I thought a bit about this issue and how we could implement it in `wasmi`. I came up with 2 competing designs:

1) We could do something very similar to Python's GIL where we'd not allow for actually concurrent Wasm executions but instead users could "queue" them up. This way we could reuse most engine resources since we would only need one Wasm executor. The neat side effect with this solution is that it is very easy to implement on top of the current architecture and things like stack overflow checking work out of the box.
2) We could do something very similar to how the old `wasmi` behaved. The old `wasmi` simply spawned an entirely new Wasm engine for every function invocation. This is extremely costly and didn't allow to reuse resources in the way the old `wasmi` was implemented. However, it is also the design that allows for most freedom for users and in theory users could even run multiple Wasm executions in parallel since they are not actually sharing resources. If we were to implement something like this in the new `wasmi` we would try to reuse as many resources as possible instead of simply spinning up an entirely new engine every time. A problem with this solution is that we need to take extra care to not violate stack overflow checks etc. I bet that this was not done properly in the old `wasmi` engine.

[...]

PR #590, opened 2022-12-07
Allow host functions to call Wasm functions

Robbepop:
Closes https://github.com/paritytech/wasmi/issues/572


Robbepop:
The PR is no longer in WIP phase and is ready to be tested.

One remaining thing left to do is to implement stack limits that are based on the offsets of parent call stacks to properly guard against requiring more stack space than has been preconfigured.
For example, with the current implementation it is possible to circumvent call and value space limitations set in `StackLimits` via `Config` by calling host and Wasm functions to reset the stacks in between.

When this issue is resolved this PR is good to go.

Robbepop:
> The PR is no longer in WIP phase and is ready to be tested.
> 
> One remaining thing left to do is to implement stack limits that are based on the offsets of parent call stacks to properly guard against requiring more stack space than has been preconfigured. For example, with the current implementation it is possible to circumvent call and value space limitations set in `StackLimits` via `Config` by calling host and Wasm functions to reset the stacks in between.
> 
> When this issue is resolved this PR is good to go.

Since this fix is non trivial I decided to do that in another PR and merge this PR as is.

Issue #728, opened 2023-05-29
Add `ResourceLimiter` to `wasmi::Store`


Robbepop:
As requested in [this comment](https://github.com/paritytech/wasmi/issues/720#issuecomment-1565809927) some `wasmi` users could profit from having access to a Wasmtime style [`ResourceLimiter` API](https://docs.rs/wasmtime/9.0.2/wasmtime/trait.ResourceLimiter.html).

Since `wasmi` tries to mirror the Wasmtime API we should and could mirror this API from Wasmtime as well. I think it would be a nice addition, however, special care has to be taken that translation and execution performance do not regress.

## TODOs

- [ ] Implementation
    - [ ] Implement the `ResourceLimiter` trait in `wasmi`.
    - [ ] Implement [`Store::limiter` API](https://docs.rs/wasmtime/9.0.2/wasmtime/struct.Store.html#method.limiter).
    - [ ] Implement [`StoreLimits` type](https://docs.rs/wasmtime/10.0.1/wasmtime/struct.StoreLimits.html) implementing `StoreLimiter` trait as convenience for users.
    - [ ] Implement [`StoreLimitsBuilder` type](https://docs.rs/wasmtime/10.0.1/wasmtime/struct.StoreLimitsBuilder.html) as convenience for constructing `StoreLimits`.
- [ ] Testing & QA
    - [ ] Add tests to verify that the implemented functionality works as intended.
    - [ ] Add proper docs to the feature as well as some doc example.
    - [ ] Validate with benchmarks that using this feature does not disrupt `wasmi` performance too much.
    - [ ] Validate with benchmarks that _not_ using this feature does not disrupt `wasmi` performance significantly.

**Note:** Wasmtime's [`epoch_interruption`](https://docs.rs/wasmtime/9.0.2/wasmtime/struct.Config.html#method.epoch_interruption) is not needed since `wasmi` does not (yet) support `async` execution.

Robbepop:
Sorry for my late response!

At the moment I am working on https://github.com/paritytech/wasmi/pull/729 which takes a big chunk of my time. I thought that this issue might be a perfect candidate for external contributions since it should be a rather local change. If it is extremely pressing we can probably work something out together. Shouldn't be a big task from what I see so far.

graydon:
@Robbepop ok if you're comfortable having us implement this, we're happy to direct our efforts to it. It looks like it ought to be relatively straightforward.

Robbepop:
[...]

PR #737, opened 2023-06-29
Support for `ResourceLimiter` API

graydon:
Closes https://github.com/paritytech/wasmi/issues/728.

This implements the `wasmtime` `ResourceLimiter` API, as discussed in bug https://github.com/paritytech/wasmi/issues/728 .

Current limitations, caveats and points to discuss/review:

- ~Missing tests~
- ~Missing docs~
- ~Missing convenience types `StoreLimits` and `StoreLimitsBuilder`~
- To keep `#[derive(Debug)]` working I had to add a couple wrapper types.
- ~As much as possible, I tried to track `wasmtime`'s implementation: the `{instance,memory,table}_{count,limit}` fields, the `Store::bump_resource_counts` method, the `limits.rs` file, the global defaults returned from the default methods, etc. etc. This all feels like a fairly weird design to me (information cached or returned from multiple places) but I figured you'd prefer to follow along their lead rather than make something gratuitously incompatible. Any deviations (see following points) were forced by differences between `wasmtime` and `wasmi`.~
- Awkwardly, because `wasmi` does not thread the full typed `Store<T>` into its executor context, only `StoreInner`, it's not possible to call the typed `FnMut(&mut T) -> &mut dyn ResourceLimiter` query API on `StoreInner` (it has no `data: T` element to pass as an argument) so I had to hoist the query up the call stack to caller sites that still have a `Store<T>` and then pass through a new `ResourceLimiterRef` type carrying the resulting ref. This seems to work but it's a bit ugly. The other options (refactoring `Store` and/or `Executor`) seemed worse.
- Unfortunately `wasmtime` just uses `anyhow::Error` everywhere, so `wasmi`'s specific error types all deviate. I did not do a lot of careful design work on the errors, I just reused those that already seemed to cover the cases and added new ones where necessary. They could theoretically be reorganized or refactored -- I don't care how they're represented at all, happy to do whatever you prefer.
- The `wasmtime` `ResourceLimiter` API actually seems .. fairly bad? I mean, it operates in terms of bytes (weird) and has the option of both returning an `Err(...)`-which-means-trap (which I didn't even try to support) _and_ returning `Ok(false)`-which-means-return-your-own-error. I am not sure how best to map its seemingly-bad-API-choices onto `wasmi`, I did my best.
- It also doesn't cap the number of globals, imports or exports, or any other sources of "too much stuff" in a wasm module. No explanation why. I think it probably should?
- Because `ResourceLimiter` wants to be called both before any inherent checks _and_ after any error occurs (including those resulting from inherent checks) I had to refactor the arithmetic in the memory and table growth functions, and repeat some code between initialization and growth. I believe I kept the logic the same. I might be able to factor some of this more, but it seems like diminishing returns.

Anyway, all that aside I _think_ this holds together. Feedback welcome!

### DIFF
--- a/crates/core/src/trap.rs
+++ b/crates/core/src/trap.rs
@@ -286,10 +286,10 @@ pub enum TrapCode {
     OutOfFuel,
 
     /// This trap is raised when a growth operation was attempted and an
-    /// installed [`ResourceLimiter`](crate::ResourceLimiter) returned
-    /// `Err(...)` from the associated `table_growing` or `memory_growing`
-    /// method, indicating a desire on the part of the embedder to trap the
-    /// interpreter rather than merely fail the growth operation.
+    /// installed `wasmi::ResourceLimiter` returned `Err(...)` from the
+    /// associated `table_growing` or `memory_growing` method, indicating a
+    /// desire on the part of the embedder to trap the interpreter rather than
+    /// merely fail the growth operation.
     GrowthOperationLimited,
 }
 
--- a/crates/core/src/trap.rs
+++ b/crates/core/src/trap.rs
@@ -286,10 +286,10 @@ pub enum TrapCode {
     OutOfFuel,
 
     /// This trap is raised when a growth operation was attempted and an
-    /// installed `ResourceLimiter` returned `Err(...)` from the associated
-    /// `table_growing` or `memory_growing` method, indicating a desire on the
-    /// part of the embedder to trap the interpreter rather than merely fail the
-    /// growth operation.
+    /// installed [`ResourceLimiter`](crate::ResourceLimiter) returned
+    /// `Err(...)` from the associated `table_growing` or `memory_growing`
+    /// method, indicating a desire on the part of the embedder to trap the
+    /// interpreter rather than merely fail the growth operation.
     GrowthOperationLimited,
 }
 
--- a/crates/core/src/trap.rs
+++ b/crates/core/src/trap.rs
@@ -284,6 +284,13 @@ pub enum TrapCode {
     /// internal bytecode so that fuel is consumed for each executed instruction.
     /// This is useful to deterministically halt or yield a WebAssembly execution.
     OutOfFuel,
+
+    /// This trap is raised when a growth operation was attempted and an
+    /// installed `ResourceLimiter` returned `Err(...)` from the associated
+    /// `table_growing` or `memory_growing` method, indicating a desire on the
+    /// part of the embedder to trap the interpreter rather than merely fail the
+    /// growth operation.
+    GrowthOperationLimited,
 }
 
 impl TrapCode {
@@ -305,6 +312,7 @@ impl TrapCode {
             Self::StackOverflow => "call stack exhausted",
             Self::BadSignature => "indirect call type mismatch",
             Self::OutOfFuel => "all fuel consumed by WebAssembly",
+            Self::GrowthOperationLimited => "growth operation limited",
         }
     }
 }
--- a/crates/wasmi/src/engine/executor.rs
+++ b/crates/wasmi/src/engine/executor.rs
@@ -24,6 +24,7 @@ use crate::{
         ValueStack,
     },
     func::FuncEntity,
+    store::ResourceLimiterRef,
     table::TableEntity,
     FuelConsumptionMode,
     Func,
@@ -92,15 +93,25 @@ pub enum ReturnOutcome {
 ///
 /// If the Wasm execution traps.
 #[inline(never)]
-pub fn execute_wasm<'engine>(
-    ctx: &mut StoreInner,
+pub fn execute_wasm<'ctx, 'engine>(
+    ctx: &'ctx mut StoreInner,
     cache: &'engine mut InstanceCache,
     value_stack: &'engine mut ValueStack,
     call_stack: &'engine mut CallStack,
     code_map: &'engine CodeMap,
     const_pool: ConstPoolView<'engine>,
+    resource_limiter: ResourceLimiterRef<'ctx>,
 ) -> Result<WasmOutcome, TrapCode> {
-    Executor::new(ctx, cache, value_stack, call_stack, code_map, const_pool).execute()
+    Executor::new(
+        ctx,
+        cache,
+        value_stack,
+        call_stack,
+        code_map,
+        const_pool,
+        resource_limiter,
+    )
+    .execute()
 }
 
 /// The function signature of Wasm load operations.
@@ -168,6 +179,7 @@ struct Executor<'ctx, 'engine> {
     code_map: &'engine CodeMap,
     /// A read-only view to a pool of constant values.
     const_pool: ConstPoolView<'engine>,
+    resource_limiter: ResourceLimiterRef<'ctx>,
 }
 
 macro_rules! forward_call {
@@ -195,6 +207,7 @@ impl<'ctx, 'engine> Executor<'ctx, 'engine> {
         call_stack: &'engine mut CallStack,
         code_map: &'engine CodeMap,
         const_pool: ConstPoolView<'engine>,
+        resource_limiter: ResourceLimiterRef<'ctx>,
     ) -> Self {
         let frame = call_stack.pop().expect("must have frame on the call stack");
         let sp = value_stack.stack_ptr();
@@ -208,6 +221,7 @@ impl<'ctx, 'engine> Executor<'ctx, 'engine> {
             call_stack,
             code_map,
             const_pool,
+            resource_limiter,
         }
     }
 
@@ -1097,7 +1111,7 @@ impl<'ctx, 'engine> Executor<'ctx, 'engine> {
                 let new_pages = this
                     .ctx
                     .resolve_memory_mut(memory)
-                    .grow(delta)
+                    .grow(delta, &mut this.resource_limiter)
                     .map(u32::from)
                     .map_err(|_| EntityGrowError::InvalidGrow)?;
                 // The `memory.grow` operation might have invalidated the cached
@@ -1219,7 +1233,7 @@ impl<'ctx, 'engine> Executor<'ctx, 'engine> {
                 let table = this.cache.get_table(this.ctx, table_index);
                 this.ctx
                     .resolve_table_mut(&table)
-                    .grow_untyped(delta, init)
+                    .grow_untyped(delta, init, &mut this.resource_limiter)
                     .map_err(|_| EntityGrowError::InvalidGrow)
             },
         );
--- a/crates/wasmi/src/engine/executor.rs
+++ b/crates/wasmi/src/engine/executor.rs
@@ -100,7 +100,7 @@ pub fn execute_wasm<'ctx, 'engine>(
     call_stack: &'engine mut CallStack,
     code_map: &'engine CodeMap,
     const_pool: ConstPoolView<'engine>,
-    resource_limiter: ResourceLimiterRef<'ctx>,
+    resource_limiter: &'ctx mut ResourceLimiterRef<'ctx>,
 ) -> Result<WasmOutcome, TrapCode> {
     Executor::new(
         ctx,
@@ -179,7 +179,7 @@ struct Executor<'ctx, 'engine> {
     code_map: &'engine CodeMap,
     /// A read-only view to a pool of constant values.
     const_pool: ConstPoolView<'engine>,
-    resource_limiter: ResourceLimiterRef<'ctx>,
+    resource_limiter: &'ctx mut ResourceLimiterRef<'ctx>,
 }
 
 macro_rules! forward_call {
@@ -207,7 +207,7 @@ impl<'ctx, 'engine> Executor<'ctx, 'engine> {
         call_stack: &'engine mut CallStack,
         code_map: &'engine CodeMap,
         const_pool: ConstPoolView<'engine>,
-        resource_limiter: ResourceLimiterRef<'ctx>,
+        resource_limiter: &'ctx mut ResourceLimiterRef<'ctx>,
     ) -> Self {
         let frame = call_stack.pop().expect("must have frame on the call stack");
         let sp = value_stack.stack_ptr();
--- a/crates/wasmi/src/engine/executor.rs
+++ b/crates/wasmi/src/engine/executor.rs
@@ -102,16 +102,8 @@ pub fn execute_wasm<'ctx, 'engine>(
     const_pool: ConstPoolView<'engine>,
     resource_limiter: &'ctx mut ResourceLimiterRef<'ctx>,
 ) -> Result<WasmOutcome, TrapCode> {
-    Executor::new(
-        ctx,
-        cache,
-        value_stack,
-        call_stack,
-        code_map,
-        const_pool,
-        resource_limiter,
-    )
-    .execute()
+    Executor::new(ctx, cache, value_stack, call_stack, code_map, const_pool)
+        .execute(resource_limiter)
 }
 
 /// The function signature of Wasm load operations.
@@ -179,7 +171,6 @@ struct Executor<'ctx, 'engine> {
     code_map: &'engine CodeMap,
     /// A read-only view to a pool of constant values.
     const_pool: ConstPoolView<'engine>,
-    resource_limiter: &'ctx mut ResourceLimiterRef<'ctx>,
 }
 
 macro_rules! forward_call {
@@ -207,7 +198,6 @@ impl<'ctx, 'engine> Executor<'ctx, 'engine> {
         call_stack: &'engine mut CallStack,
         code_map: &'engine CodeMap,
         const_pool: ConstPoolView<'engine>,
-        resource_limiter: &'ctx mut ResourceLimiterRef<'ctx>,
     ) -> Self {
         let frame = call_stack.pop().expect("must have frame on the call stack");
         let sp = value_stack.stack_ptr();
@@ -221,13 +211,15 @@ impl<'ctx, 'engine> Executor<'ctx, 'engine> {
             call_stack,
             code_map,
             const_pool,
-            resource_limiter,
         }
     }
 
     /// Executes the function frame until it returns or traps.
     #[inline(always)]
-    fn execute(mut self) -> Result<WasmOutcome, TrapCode> {
+    fn execute(
+        mut self,
+        resource_limiter: &'ctx mut ResourceLimiterRef<'ctx>,
+    ) -> Result<WasmOutcome, TrapCode> {
         use Instruction as Instr;
         loop {
             match *self.ip.get() {
@@ -294,13 +286,13 @@ impl<'ctx, 'engine> Executor<'ctx, 'engine> {
                 Instr::I64Store16(offset) => self.visit_i64_store_16(offset)?,
                 Instr::I64Store32(offset) => self.visit_i64_store_32(offset)?,
                 Instr::MemorySize => self.visit_memory_size(),
-                Instr::MemoryGrow => self.visit_memory_grow()?,
+                Instr::MemoryGrow => self.visit_memory_grow(&mut *resource_limiter)?,
                 Instr::MemoryFill => self.visit_memory_fill()?,
                 Instr::MemoryCopy => self.visit_memory_copy()?,
                 Instr::MemoryInit(segment) => self.visit_memory_init(segment)?,
                 Instr::DataDrop(segment) => self.visit_data_drop(segment),
                 Instr::TableSize(table) => self.visit_table_size(table),
-                Instr::TableGrow(table) => self.visit_table_grow(table)?,
+                Instr::TableGrow(table) => self.visit_table_grow(table, &mut *resource_limiter)?,
                 Instr::TableFill(table) => self.visit_table_fill(table)?,
                 Instr::TableGet(table) => self.visit_table_get(table)?,
                 Instr::TableSet(table) => self.visit_table_set(table)?,
@@ -1091,7 +1083,10 @@ impl<'ctx, 'engine> Executor<'ctx, 'engine> {
     }
 
     #[inline(always)]
-    fn visit_memory_grow(&mut self) -> Result<(), TrapCode> {
+    fn visit_memory_grow(
+        &mut self,
+        resource_limiter: &mut ResourceLimiterRef<'ctx>,
+    ) -> Result<(), TrapCode> {
         let delta: u32 = self.sp.pop_as();
         let delta = match Pages::new(delta) {
             Some(pages) => pages,
@@ -1111,7 +1106,7 @@ impl<'ctx, 'engine> Executor<'ctx, 'engine> {
                 let new_pages = this
                     .ctx
                     .resolve_memory_mut(memory)
-                    .grow(delta, &mut this.resource_limiter)
+                    .grow(delta, resource_limiter)
                     .map(u32::from)?;
                 // The `memory.grow` operation might have invalidated the cached
                 // linear memory so we need to reset it in order for the cache to
@@ -1223,18 +1218,20 @@ impl<'ctx, 'engine> Executor<'ctx, 'engine> {
     }
 
     #[inline(always)]
-    fn visit_table_grow(&mut self, table_index: TableIdx) -> Result<(), TrapCode> {
+    fn visit_table_grow(
+        &mut self,
+        table_index: TableIdx,
+        resource_limiter: &mut ResourceLimiterRef<'ctx>,
+    ) -> Result<(), TrapCode> {
         let (init, delta) = self.sp.pop2();
         let delta: u32 = delta.into();
         let result = self.consume_fuel_with(
             |costs| costs.fuel_for_elements(u64::from(delta)),
             |this| {
                 let table = this.cache.get_table(this.ctx, table_index);
-                this.ctx.resolve_table_mut(&table).grow_untyped(
-                    delta,
-                    init,
-                    &mut this.resource_limiter,
-                )
+                this.ctx
+                    .resolve_table_mut(&table)
+                    .grow_untyped(delta, init, resource_limiter)
             },
         );
         let result = match result {
--- a/crates/wasmi/src/engine/executor.rs
+++ b/crates/wasmi/src/engine/executor.rs
@@ -1112,8 +1112,7 @@ impl<'ctx, 'engine> Executor<'ctx, 'engine> {
                     .ctx
                     .resolve_memory_mut(memory)
                     .grow(delta, &mut this.resource_limiter)
-                    .map(u32::from)
-                    .map_err(|_| EntityGrowError::InvalidGrow)?;
+                    .map(u32::from)?;
                 // The `memory.grow` operation might have invalidated the cached
                 // linear memory so we need to reset it in order for the cache to
                 // reload in case it is used again.
@@ -1231,10 +1230,11 @@ impl<'ctx, 'engine> Executor<'ctx, 'engine> {
             |costs| costs.fuel_for_elements(u64::from(delta)),
             |this| {
                 let table = this.cache.get_table(this.ctx, table_index);
-                this.ctx
-                    .resolve_table_mut(&table)
-                    .grow_untyped(delta, init, &mut this.resource_limiter)
-                    .map_err(|_| EntityGrowError::InvalidGrow)
+                this.ctx.resolve_table_mut(&table).grow_untyped(
+                    delta,
+                    init,
+                    &mut this.resource_limiter,
+                )
             },
         );
         let result = match result {
--- a/crates/wasmi/src/limits.rs
+++ b/crates/wasmi/src/limits.rs
@@ -257,7 +257,7 @@ impl ResourceLimiter for StoreLimits {
 
     fn table_growing(
         &mut self,
-        _current: u32,
+        current: u32,
         desired: u32,
         maximum: Option<u32>,
     ) -> Result<bool, TableError> {
@@ -269,7 +269,11 @@ impl ResourceLimiter for StoreLimits {
             },
         };
         if !allow && self.trap_on_grow_failure {
-            Err(TableError::GrowOutOfBounds)
+            Err(TableError::GrowOutOfBounds {
+                maximum: maximum.unwrap_or(u32::MAX),
+                current: current,
+                delta: desired - current,
+            })
         } else {
             Ok(allow)
         }
--- a/crates/wasmi/src/limits.rs
+++ b/crates/wasmi/src/limits.rs
@@ -257,7 +257,7 @@ impl ResourceLimiter for StoreLimits {
 
     fn table_growing(
         &mut self,
-        current: u32,
+        _current: u32,
         desired: u32,
         maximum: Option<u32>,
     ) -> Result<bool, TableError> {
--- a/crates/wasmi/src/limits.rs
+++ b/crates/wasmi/src/limits.rs
@@ -51,6 +51,8 @@ pub trait ResourceLimiter {
     /// [`Memory::grow`](crate::Memory::grow) an error will be returned from
     /// those methods.
     ///
+    /// # Errors
+    ///
     /// If `Err(e)` is returned then the `memory.grow` function will behave
     /// as if a trap has been raised. Note that this is not necessarily
     /// compliant with the WebAssembly specification but it can be a handy and
@@ -71,8 +73,10 @@ pub trait ResourceLimiter {
     /// * `maximum` is either the table's maximum or a maximum from an instance
     ///   allocator.  A value of `None` indicates that the table is unbounded.
     ///
+    /// # Errors
+    ///
     /// See the details on the return values for `memory_growing` for what the
-    /// return value of this function indicates.
+    /// return values of this function indicates.
     fn table_growing(
         &mut self,
         current: u32,
@@ -203,6 +207,12 @@ impl StoreLimitsBuilder {
     }
 }
 
+impl Default for StoreLimitsBuilder {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
 /// Provides limits for a [`Store`](crate::Store).
 ///
 /// This type is created with a [`StoreLimitsBuilder`] and is typically used in
@@ -245,7 +255,7 @@ impl ResourceLimiter for StoreLimits {
             Some(limit) if desired > limit => false,
             _ => match maximum {
                 Some(max) if desired > max => false,
-                _ => true,
+                Some(_) | None => true,
             },
         };
         if !allow && self.trap_on_grow_failure {
@@ -265,13 +275,13 @@ impl ResourceLimiter for StoreLimits {
             Some(limit) if desired > limit => false,
             _ => match maximum {
                 Some(max) if desired > max => false,
-                _ => true,
+                Some(_) | None => true,
             },
         };
         if !allow && self.trap_on_grow_failure {
             Err(TableError::GrowOutOfBounds {
                 maximum: maximum.unwrap_or(u32::MAX),
-                current: current,
+                current,
                 delta: desired - current,
             })
         } else {
--- a/crates/wasmi/src/limits.rs
+++ b/crates/wasmi/src/limits.rs
@@ -1,0 +1,39 @@
+use crate::{memory::MemoryError, table::TableError};
+
+/// Value returned by [`ResourceLimiter::instances`] default method
+pub const DEFAULT_INSTANCE_LIMIT: usize = 10000;
+
+/// Value returned by [`ResourceLimiter::tables`] default method
+pub const DEFAULT_TABLE_LIMIT: usize = 10000;
+
+/// Value returned by [`ResourceLimiter::memories`] default method
+pub const DEFAULT_MEMORY_LIMIT: usize = 10000;
+
+pub trait ResourceLimiter {
+    fn memory_growing(
+        &mut self,
+        current: usize,
+        desired: usize,
+        maximum: Option<usize>,
+    ) -> Result<bool, MemoryError>;
+
+    fn table_growing(
+        &mut self,
+        current: u32,
+        desired: u32,
+        maximum: Option<u32>,
+    ) -> Result<bool, TableError>;
+
+    // Provided methods
+    fn memory_grow_failed(&mut self, _error: &MemoryError) {}
+    fn table_grow_failed(&mut self, _error: &TableError) {}
+    fn instances(&self) -> usize {
+        DEFAULT_INSTANCE_LIMIT
+    }
+    fn tables(&self) -> usize {
+        DEFAULT_TABLE_LIMIT
+    }
+    fn memories(&self) -> usize {
+        DEFAULT_MEMORY_LIMIT
+    }
+}
--- a/crates/wasmi/src/limits.rs
+++ b/crates/wasmi/src/limits.rs
@@ -9,7 +9,53 @@ pub const DEFAULT_TABLE_LIMIT: usize = 10000;
 /// Value returned by [`ResourceLimiter::memories`] default method
 pub const DEFAULT_MEMORY_LIMIT: usize = 10000;
 
+/// Used by hosts to limit resource consumption of instances.
+///
+/// This trait is used in conjunction with the
+/// [`Store::limiter`](crate::Store::limiter) to limit the
+/// allocation of resources within a store. As a store-level limit this means
+/// that all creation of instances, memories, and tables are limited within the
+/// store. Resources limited via this trait are primarily related to memory and
+/// limiting CPU resources needs to be done with something such as
+/// [`Config::consume_fuel`](crate::Config::consume_fuel).
+///
+/// Note that this trait does not limit 100% of memory allocated via a
+/// [`Store`](crate::Store). Wasmi will still allocate memory to track data
+/// structures and additionally embedder-specific memory allocations are not
+/// tracked via this trait. This trait only limits resources allocated by a
+/// WebAssembly instance itself.
 pub trait ResourceLimiter {
+    /// Notifies the resource limiter that an instance's linear memory has been
+    /// requested to grow.
+    ///
+    /// * `current` is the current size of the linear memory in bytes.
+    /// * `desired` is the desired size of the linear memory in bytes.
+    /// * `maximum` is either the linear memory's maximum or a maximum from an
+    ///   instance allocator, also in bytes. A value of `None`
+    ///   indicates that the linear memory is unbounded.
+    ///
+    /// The `current` and `desired` amounts are guaranteed to always be
+    /// multiples of the WebAssembly page size, 64KiB.
+    ///
+    /// ## Return Value
+    ///
+    /// If `Ok(true)` is returned from this function then the growth operation
+    /// is allowed. This means that the wasm `memory.grow` instruction will
+    /// return with the `desired` size, in wasm pages. Note that even if
+    /// `Ok(true)` is returned, though, if `desired` exceeds `maximum` then the
+    /// growth operation will still fail.
+    ///
+    /// If `Ok(false)` is returned then this will cause the `memory.grow`
+    /// instruction in a module to return -1 (failure), or in the case of an
+    /// embedder API calling [`Memory::new`](crate::Memory::new) or
+    /// [`Memory::grow`](crate::Memory::grow) an error will be returned from
+    /// those methods.
+    ///
+    /// If `Err(e)` is returned then the `memory.grow` function will behave
+    /// as if a trap has been raised. Note that this is not necessarily
+    /// compliant with the WebAssembly specification but it can be a handy and
+    /// useful tool to get a precise backtrace at "what requested so much memory
+    /// to cause a growth failure?".
     fn memory_growing(
         &mut self,
         current: usize,
@@ -17,6 +63,16 @@ pub trait ResourceLimiter {
         maximum: Option<usize>,
     ) -> Result<bool, MemoryError>;
 
+    /// Notifies the resource limiter that an instance's table has been
+    /// requested to grow.
+    ///
+    /// * `current` is the current number of elements in the table.
+    /// * `desired` is the desired number of elements in the table.
+    /// * `maximum` is either the table's maximum or a maximum from an instance
+    ///   allocator.  A value of `None` indicates that the table is unbounded.
+    ///
+    /// See the details on the return values for `memory_growing` for what the
+    /// return value of this function indicates.
     fn table_growing(
         &mut self,
         current: u32,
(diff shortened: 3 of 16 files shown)


Robbepop:
@graydon Just saw your PR and quickly scanned over it but what I saw and ! looked solid and the design decisions are reasonable. Going to take a more in-depth look now.

> Missing convenience types `StoreLimits` and `StoreLimitsBuilder`

Do you plan to add them to this PR or prefer to phase them out of this PR? 

> It also doesn't cap the number of globals, imports or exports, or any other sources of "too much stuff" in a wasm module. No explanation why. I think it probably should?

Sounds strange indeed. When I am confused about APIs or designs in Wasmtime I often simply ask the team. More often than not they have reasonable rationals behind those decisions.

> Awkwardly, because `wasmi` does not thread the full typed `Store<T>` into its executor context, only `StoreInner`, it's not possible to call the typed `FnMut(&mut T) -> &mut dyn ResourceLimiter` query API on `StoreInner` (it has no data: T element to pass as an argument) so I had to hoist the query up the call stack to caller sites that still have a `Store<T>` and then pass through a new `ResourceLimiterRef` type carrying the resulting ref. This seems to work but it's a bit ugly. The other options (refactoring Store and/or Executor) seemed worse.

This `wasmi` executor design was chosen because it really helps with certain optimizations that we were able to apply after splitting out the `StoreInner` with respect to function calls and also somehow `rustc` does a much better job optimizing non-generic code.
Your chosen design sounds fairly reasonable given the choices you had.

> The `wasmtime` `ResourceLimiter` API actually seems .. fairly bad? I mean, it operates in terms of bytes (weird) and has the option of both returning an `Err(...)`-which-means-trap (which I didn't even try to support) and returning `Ok(false)`-which-means-return-your-own-error. I am not sure how best to map its seemingly-bad-API-choices onto `wasmi`, I did my best.

I also wondered why it operates on bytes instead of pages. Oftentimes there is good reason behind such designs but then it would often be nice to have an accompanying comment to explain why this design was chosen.

> Unfortunately `wasmtime` just uses `anyhow::Error` everywhere, so `wasmi`'s specific error types all deviate.

Did you write "unfortunately" because you think `anyhow::Error` is not a good choice or because it deviates from `wasmi`'s errors? Errors in `wasmi` are a big mess unfortunately. I never got around to properly clean them up. We also cannot use `anyhow::Error` because it is incompatible with `no_std`.

crates/wasmi/src/store.rs
```diff
@@ -67,6 +68,22 @@ impl StoreIdx {
 /// A stored entity.
 pub type Stored<Idx> = GuardedEntity<StoreIdx, Idx>;
 
+pub struct ResourceLimiterRef<'a>(pub(crate) Option<&'a mut (dyn crate::ResourceLimiter)>);
```
Robbepop, line 71: After seeing many instances of
```rust
ResourceLimiterRef(match &mut store.limiter {
    Some(q) => Some(q.0(&mut store.data)),
    None => None,
})
```
I think it is fair to say that convenience getter would be nice. With the following 2 getters it is possible to get rid of all those `ResourceLimiterRef(match { .. })` constructs in the rest of the codebase and hide internals as well in `impl Store`:
```rust
pub(crate) fn store_inner_and_resource_limiter_ref(&mut self) -> (&mut StoreInner, ResourceLimiterRef) {
    let resource_limiter = ResourceLimiterRef(match &mut self.limiter {
        Some(q) => Some(q.0(&mut self.data)),
        None => None
    });
    (&mut self.inner, resource_limiter)
}

pub(crate) fn resource_limiter_ref(&mut self) -> ResourceLimiterRef {
    self.store_inner_and_resource_limiter_ref().1
}
```
This allows us to remove some `pub(crate)` annotations.

crates/wasmi/src/store.rs
```diff
@@ -80,7 +97,8 @@ pub struct Store<T> {
     /// Stored host function trampolines.
     trampolines: Arena<TrampolineIdx, TrampolineEntity<T>>,
     /// User provided host data owned by the [`Store`].
-    data: T,
+    pub(crate) data: T,
+    pub(crate) limiter: Option<ResourceLimiterQuery<T>>,
```
Robbepop, line 101: With the aforementioned getters for `ResourceLimiterRef` we can get rid of those `pub(crate)`.

crates/wasmi/src/store.rs
Robbepop, line 71: This is just personal preference but I would prefer to have no `pub(crate)` for the field in the `ResourceLimiterRef` type and instead a convenience getter:
```rust
impl<'a> ResourceLimiterRef<'a> {
    pub fn as_resource_limiter(&mut self) -> &mut Option<&'a mut dyn crate::ResourceLimiter> {
        &mut self.0
    }
}
```
Then all instances of `&mut limiter.0` can be replaced with `limiter.as_resource_limiter()`.

crates/wasmi/src/store.rs
```diff
@@ -67,6 +68,22 @@ impl StoreIdx {
...
 
+pub struct ResourceLimiterRef<'a>(pub(crate) Option<&'a mut (dyn crate::ResourceLimiter)>);
+impl<'a> core::fmt::Debug for ResourceLimiterRef<'a> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "ResourceLimiterRef(...)")
+    }
+}
+
+pub(crate) struct ResourceLimiterQuery<T>(
+    pub(crate) Box<dyn FnMut(&mut T) -> &mut (dyn crate::ResourceLimiter) + Send + Sync>,
```
Robbepop, line 79: From what I see the `pub(crate)` visibility modifier is not used and can be removed. Probably an oversight.

crates/wasmi/src/table/mod.rs
```diff
@@ -535,12 +586,13 @@ impl Table {
...
         delta: u32,
         init: Value,
+        limiter: &mut ResourceLimiterRef<'_>,
     ) -> Result<u32, TableError> {
         ctx.as_context_mut()
             .store
             .inner
             .resolve_table_mut(self)
-            .grow(delta, init)
+            .grow(delta, init, limiter)
```
Robbepop, line 595: Same as with `Memory::grow`: We shouldn't alter the API of user facing types.
Instead (with the addition of the aforementioned getters) we can do the following:
```rust
pub fn grow(
    &self,
    mut ctx: impl AsContextMut,
    delta: u32,
    init: Value,
) -> Result<u32, TableError> {
    let (store_inner, mut limiter) = ctx
        .as_context_mut()
        .store
        .store_inner_and_resource_limiter_ref();
    store_inner
        .resolve_table_mut(self)
        .grow(delta, init, &mut limiter)
}
```

Robbepop (commented):
Forgot to leave a review note:

Design looks good as is. I think this is good to go after incorporating the few review items and adding docs. However, there is still a need to run the benchmarks first. With the review items fixed the benchmarks should compile again.

crates/wasmi/src/store.rs
graydon: Ok.

graydon: Sure.

crates/wasmi/src/store.rs
graydon: Yeah I was calling it from outside the module at one point while I was moving stuff around. Looks superfluous now.

crates/wasmi/src/store.rs
graydon: Will do.

crates/wasmi/src/table/mod.rs
graydon: Ok (changed in my local branch).

I should mention that both `MemoryEntity::new` and `TableEntity::new` methods do change still. Not sure if those concern you (since they're not `wasmtime` types). They are public.

crates/wasmi/src/memory/mod.rs
```diff
@@ -318,12 +379,13 @@ impl Memory {
...
         mut ctx: impl AsContextMut,
         additional: Pages,
+        limiter: &mut ResourceLimiterRef<'_>,
     ) -> Result<Pages, MemoryError> {
         ctx.as_context_mut()
             .store
             .inner
             .resolve_memory_mut(self)
-            .grow(additional)
+            .grow(additional, limiter)
```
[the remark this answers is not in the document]
graydon: Done.

graydon:
> > Missing convenience types `StoreLimits` and `StoreLimitsBuilder`
> 
> Do you plan to add them to this PR or prefer to phase them out of this PR?

Oh, I'll put them in here, just wanted to get preliminary feedback. Glad this looks agreeable!

> > It also doesn't cap the number of globals, imports or exports, or any other sources of "too much stuff" in a wasm module. No explanation why. I think it probably should?
> 
> Sounds strange indeed. When I am confused about APIs or designs in Wasmtime I often simply ask the team. More often than not they have reasonable rationals behind those decisions.

I saw you open and then close a [wasmtime bug](https://github.com/bytecodealliance/wasmtime/issues/6668) about this, but I'm not sure what its conclusion means in terms of what you'd like me to do here. Should I put in more fixed limits (for globals, imports, exports, or other things) or leave it as is? I don't really expect people will be able to attack us too badly by defining a lot of these (especially since I think they're limited by the static byte-size of the wasm module input, and we bound that) but it might be nice to have a way to spell out specific limits too. What do you think?

> > Unfortunately `wasmtime` just uses `anyhow::Error` everywhere, so `wasmi`'s specific error types all deviate.
> 
> Did you write "unfortunately" because you think `anyhow::Error` is not a good choice or because it deviates from `wasmi`'s errors? Errors in `wasmi` are a big mess unfortunately. I never got around to properly clean them up. We also cannot use `anyhow::Error` because it is incompatible with `no_std`.

Oh, just "unfortunate" in the sense that it  means I can't easily make the interfaces _identical_, it's a source of divergence. I don't much care, Rust errors are always a little messy / often don't compose very well.

One other thing I wanted to ask: in terms of documentation, the wasmtime interfaces and implementations all have fairly long (well written, nice!) doc-comments. I can copy these over verbatim -- I think the licenses allow it -- or I can rewrite them myself, either equally verbose, or just with brief summaries ... do you have a preference?

Also, what level of testing will be adequate? Do you like to see doctests and such?

graydon:
Oh, also: the `ResourceLimiter` APIs around memory/table growth have this weird thing where they returns `Result<bool,...>` where

  - `Ok(true)` means "accept the growth"
  - `Ok(false)` means "reject the growth event" in a way that returns `-1` to the guest code executing the grow instruction _or_ if called from a non-guest context, return an error to the invoking API. 
  - `Err(...)` means "cause the guest to trap" which the docs state as non-wasm-compliant behaviour but sometimes useful.

I am .. not 100% clear on how best to structure this in terms of wasmi's APIs. I now _think_ that the sketch I posted here earlier is not quite right, and the `Ok(false)` path of `MemoryEntity::grow`should actually return `Ok(INVALID_GROWTH_ERRCODE)` (which is to say: `-1`) . In other words it should do:

```
            if !limiter.memory_growing(current_size, desired_size, maximum_size)? {
                // Limiter returning Ok(false) means let the guest continue
                // executing but return the official growth-failure 32-bit error code.
                return Ok(INVALID_GROWTH_ERRCODE);
            }
```

Does that seem right to you?

paritytech-cicd-pr:
## BENCHMARKS ## 

 <table> <thead> <tr> <th></th><th colspan=3>NATIVE</th><th colspan=3>WASMTIME</th><th></th> </tr> <tr> <th>BENCHMARK</th><th>MASTER</th><th>PR</th><th>DIFF</th><th>MASTER</th><th>PR</th><th>DIFF</th><th>WASMTIME OVERHEAD</th> </tr> </thead> <tbody> <tr><td><tt>execute/</tt><br><tt>bare_call_0</td> <td nowrap> 1.57ms</td> <td nowrap> 1.52ms</td> <td nowrap> :green_circle: -3.09%</td> <td nowrap> 1.02ms</td> <td nowrap> 1.09ms</td> <td nowrap> :red_circle: 7.04%</td> <td nowrap> :green_circle: -28%</td></tr><tr><td><tt>execute/</tt><br><tt>bare_call_0/typed</td> <td nowrap> 1.20ms</td> <td nowrap> 1.16ms</td> <td nowrap> :green_circle: -3.28%</td> <td nowrap> 656.95µs</td> <td nowrap> 732.96µs</td> <td nowrap> :red_circle: 11.55%</td> <td nowrap> :green_circle: -37%</td></tr><tr><td><tt>execute/</tt><br><tt>bare_call_1</td> <td nowrap> 1.61ms</td> <td nowrap> 1.59ms</td> <td nowrap> :red_circle: -1.74%</td> <td nowrap> 1.17ms</td> <td nowrap> 1.30ms</td> <td nowrap> :red_circle: 10.82%</td> <td nowrap> :green_circle: -18%</td></tr><tr><td><tt>execute/</tt><br><tt>bare_call_16</td> <td nowrap> 2.53ms</td> <td nowrap> 2.94ms</td> <td nowrap> :red_circle: 16.43%</td> <td nowrap> 3.34ms</td> <td nowrap> 4.31ms</td> <td nowrap> :red_circle: 28.80%</td> <td nowrap> :green_circle: 46%</td></tr><tr><td><tt>execute/</tt><br><tt>bare_call_16/typed</td> <td nowrap> 1.58ms</td> <td nowrap> 2.04ms</td> <td nowrap> :red_circle: 29.31%</td> <td nowrap> 1.76ms</td> <td nowrap> 2.40ms</td> <td nowrap> :red_circle: 36.52%</td> <td nowrap> :green_circle: 18%</td></tr><tr><td><tt>execute/</tt><br><tt>bare_call_1/typed</td> <td nowrap> 1.33ms</td> <td nowrap> 1.28ms</td> <td nowrap> :green_circle: -4.13%</td> <td nowrap> 928.28µs</td> <td nowrap> 1.08ms</td> <td nowrap> :red_circle: 16.46%</td> <td nowrap> :green_circle: -15%</td></tr><tr><td><tt>execute/</tt><br><tt>bare_call_4</td> <td nowrap> 1.74ms</td> <td nowrap> 1.89ms</td> <td nowrap> :red_circle: 6.51%</td> <td nowrap> 1.51ms</td> <td nowrap> 1.85ms</td> <td nowrap> :red_circle: 23.16%</td> <td nowrap> :green_circle: -2%</td></tr><tr><td><tt>execute/</tt><br><tt>bare_call_4/typed</td> <td nowrap> 1.21ms</td> <td nowrap> 1.31ms</td> <td nowrap> :red_circle: 8.36%</td> <td nowrap> 931.44µs</td> <td nowrap> 1.30ms</td> <td nowrap> :red_circle: 40.05%</td> <td nowrap> :green_circle: -0%</td></tr><tr><td><tt>execute/</tt><br><tt>br_table</td> <td nowrap> 1.30ms</td> <td nowrap> 1.47ms</td> <td nowrap> :red_circle: 12.90%</td> <td nowrap> 1.08ms</td> <td nowrap> 1.25ms</td> <td nowrap> :red_circle: 15.94%</td> <td nowrap> :green_circle: -15%</td></tr><tr><td><tt>execute/</tt><br><tt>count_until</td> <td nowrap> 591.63µs</td> <td nowrap> 1.78ms</td> <td nowrap> :red_circle: 201.42%</td> <td nowrap> 1.30ms</td> <td nowrap> 1.69ms</td> <td nowrap> :red_circle: 29.79%</td> <td nowrap> :green_circle: -5%</td></tr><tr><td><tt>execute/</tt><br><tt>factorial_iterative</td> <td nowrap> 319.60µs</td> <td nowrap> 734.84µs</td> <td nowrap> :red_circle: 129.36%</td> <td nowrap> 516.64µs</td> <td nowrap> 947.70µs</td> <td nowrap> :red_circle: 83.69%</td> <td nowrap> :green_circle: 29%</td></tr><tr><td><tt>execute/</tt><br><tt>factorial_recursive</td> <td nowrap> 486.91µs</td> <td nowrap> 876.58µs</td> <td nowrap> :red_circle: 79.96%</td> <td nowrap> 724.27µs</td> <td nowrap> 1.06ms</td> <td nowrap> :red_circle: 46.88%</td> <td nowrap> :green_circle: 21%</td></tr><tr><td><tt>execute/</tt><br><tt>fibonacci_iter</td> <td nowrap> 1.40ms</td> <td nowrap> 3.65ms</td> <td nowrap> :red_circle: 160.78%</td> <td nowrap> 2.83ms</td> <td nowrap> 3.99ms</td> <td nowrap> :red_circle: 41.24%</td> <td nowrap> :green_circle: 9%</td></tr><tr><td><tt>execute/</tt><br><tt>fibonacci_rec</td> <td nowrap> 3.93ms</td> <td nowrap> 7.39ms</td> <td nowrap> :red_circle: 87.98%</td> <td nowrap> 6.43ms</td> <td nowrap> 8.98ms</td> <td nowrap> :red_circle: 39.39%</td> <td nowrap> :green_circle: 22%</td></tr><tr><td><tt>execute/</tt><br><tt>fibonacci_tail</td> <td nowrap> 858.35µs</td> <td nowrap> 2.22ms</td> <td nowrap> :red_circle: 158.23%</td> <td nowrap> 1.45ms</td> <td nowrap> 2.18ms</td> <td nowrap> :red_circle: 49.76%</td> <td nowrap> :green_circle: -2%</td></tr><tr><td><tt>execute/</tt><br><tt>global_bump</td> <td nowrap> 750.74µs</td> <td nowrap> 1.97ms</td> <td nowrap> :red_circle: 161.03%</td> <td nowrap> 1.69ms</td> <td nowrap> 2.09ms</td> <td nowrap> :red_circle: 24.10%</td> <td nowrap> :green_circle: 6%</td></tr><tr><td><tt>execute/</tt><br><tt>global_const</td> <td nowrap> 681.54µs</td> <td nowrap> 2.03ms</td> <td nowrap> :red_circle: 197.38%</td> <td nowrap> 1.36ms</td> <td nowrap> 2.23ms</td> <td nowrap> :red_circle: 62.31%</td> <td nowrap> :green_circle: 10%</td></tr><tr><td><tt>execute/</tt><br><tt>host_calls</td> <td nowrap> 36.57µs</td> <td nowrap> 38.96µs</td> <td nowrap> :red_circle: 6.66%</td> <td nowrap> 40.51µs</td> <td nowrap> 50.80µs</td> <td nowrap> :red_circle: 24.25%</td> <td nowrap> :green_circle: 30%</td></tr><tr><td><tt>execute/</tt><br><tt>memory_fill</td> <td nowrap> 1.21ms</td> <td nowrap> 3.05ms</td> <td nowrap> :red_circle: 151.32%</td> <td nowrap> 2.28ms</td> <td nowrap> 3.21ms</td> <td nowrap> :red_circle: 41.12%</td> <td nowrap> :green_circle: 5%</td></tr><tr><td><tt>execute/</tt><br><tt>memory_sum</td> <td nowrap> 1.13ms</td> <td nowrap> 3.07ms</td> <td nowrap> :red_circle: 168.73%</td> <td nowrap> 2.31ms</td> <td nowrap> 3.35ms</td> <td nowrap> :red_circle: 45.08%</td> <td nowrap> :green_circle: 9%</td></tr><tr><td><tt>execute/</tt><br><tt>memory_vec_add</td> <td nowrap> 2.43ms</td> <td nowrap> 6.09ms</td> <td nowrap> :red_circle: 152.44%</td> <td nowrap> 4.58ms</td> <td nowrap> 7.64ms</td> <td nowrap> :red_circle: 66.96%</td> <td nowrap> :green_circle: 25%</td></tr><tr><td><tt>execute/</tt><br><tt>recursive_is_even</td> <td nowrap> 688.50µs</td> <td nowrap> 1.30ms</td> <td nowrap> :red_circle: 86.79%</td> <td nowrap> 1.05ms</td> <td nowrap> 1.46ms</td> <td nowrap> :red_circle: 39.39%</td> <td nowrap> :green_circle: 13%</td></tr><tr><td><tt>execute/</tt><br><tt>recursive_ok</td> <td nowrap> 96.99µs</td> <td nowrap> 172.94µs</td> <td nowrap> :red_circle: 79.60%</td> <td nowrap> 159.33µs</td> <td nowrap> 212.98µs</td> <td nowrap> :red_circle: 33.97%</td> <td nowrap> :green_circle: 23%</td></tr><tr><td><tt>execute/</tt><br><tt>recursive_scan</td> <td nowrap> 142.44µs</td> <td nowrap> 261.05µs</td> <td nowrap> :red_circle: 86.39%</td> <td nowrap> 202.06µs</td> <td nowrap> 292.35µs</td> <td nowrap> :red_circle: 44.78%</td> <td nowrap> :green_circle: 12%</td></tr><tr><td><tt>execute/</tt><br><tt>recursive_trap</td> <td nowrap> 9.16µs</td> <td nowrap> 17.79µs</td> <td nowrap> :red_circle: 95.38%</td> <td nowrap> 17.02µs</td> <td nowrap> 21.44µs</td> <td nowrap> :red_circle: 26.15%</td> <td nowrap> :green_circle: 21%</td></tr><tr><td><tt>execute/</tt><br><tt>regex_redux</td> <td nowrap> 457.41µs</td> <td nowrap> 976.63µs</td> <td nowrap> :red_circle: 113.52%</td> <td nowrap> 831.15µs</td> <td nowrap> 1.52ms</td> <td nowrap> :red_circle: 82.97%</td> <td nowrap> :yellow_circle: 56%</td></tr><tr><td><tt>execute/</tt><br><tt>rev_complement</td> <td nowrap> 422.46µs</td> <td nowrap> 1.01ms</td> <td nowrap> :red_circle: 138.07%</td> <td nowrap> 811.33µs</td> <td nowrap> 1.44ms</td> <td nowrap> :red_circle: 77.21%</td> <td nowrap> :green_circle: 43%</td></tr><tr><td><tt>execute/</tt><br><tt>tiny_keccak</td> <td nowrap> 355.59µs</td> <td nowrap> 911.98µs</td> <td nowrap> :red_circle: 155.66%</td> <td nowrap> 679.14µs</td> <td nowrap> 1.10ms</td> <td nowrap> :red_circle: 61.62%</td> <td nowrap> :green_circle: 21%</td></tr><tr><td><tt>execute/</tt><br><tt>trunc_f2i</td> <td nowrap> 732.37µs</td> <td nowrap> 1.55ms</td> <td nowrap> :red_circle: 111.93%</td> <td nowrap> 1.63ms</td> <td nowrap> 2.02ms</td> <td nowrap> :red_circle: 24.05%</td> <td nowrap> :green_circle: 30%</td></tr><tr><td><tt>instantiate/</tt><br><tt>wasm_kernel</td> <td nowrap> 51.72µs</td> <td nowrap> 54.58µs</td> <td nowrap> :red_circle: 4.85%</td> <td nowrap> 55.50µs</td> <td nowrap> 56.04µs</td> <td nowrap> :white_circle: -0.23%</td> <td nowrap> :green_circle: 3%</td></tr><tr><td><tt>translate/</tt><br><tt>erc1155</td> <td nowrap> 197.77µs</td> <td nowrap> 197.14µs</td> <td nowrap> :white_circle: 0.05%</td> <td nowrap> 332.29µs</td> <td nowrap> 334.87µs</td> <td nowrap> :white_circle: 0.49%</td> <td nowrap> :yellow_circle: 70%</td></tr><tr><td><tt>translate/</tt><br><tt>erc20</td> <td nowrap> 97.95µs</td> <td nowrap> 96.67µs</td> <td nowrap> :green_circle: -1.32%</td> <td nowrap> 161.31µs</td> <td nowrap> 161.42µs</td> <td nowrap> :white_circle: 0.13%</td> <td nowrap> :yellow_circle: 67%</td></tr><tr><td><tt>translate/</tt><br><tt>erc721</td> <td nowrap> 138.10µs</td> <td nowrap> 137.56µs</td> <td nowrap> :white_circle: -0.11%</td> <td nowrap> 231.76µs</td> <td nowrap> 232.72µs</td> <td nowrap> :white_circle: 0.45%</td> <td nowrap> :yellow_circle: 69%</td></tr><tr><td><tt>translate/</tt><br><tt>spidermonkey</td> <td nowrap> 61.25ms</td> <td nowrap> 61.40ms</td> <td nowrap> :white_circle: 0.12%</td> <td nowrap> 0.00ns</td> <td nowrap> 0.00ns</td> <td nowrap> :white_circle: -0.25%</td> <td nowrap> :green_circle: -100%</td></tr><tr><td><tt>translate/</tt><br><tt>wasm_kernel</td> <td nowrap> 3.96ms</td> <td nowrap> 3.97ms</td> <td nowrap> :white_circle: 0.34%</td> <td nowrap> 6.00ms</td> <td nowrap> 6.04ms</td> <td nowrap> :white_circle: 0.52%</td> <td nowrap> :yellow_circle: 52%</td></tr> </tbody> </table> 

 [Link to pipeline](https://gitlab.parity.io/parity/mirrors/wasmi/-/jobs/3095906)

Robbepop:
> I saw you open and then close a [wasmtime bug](https://github.com/bytecodealliance/wasmtime/issues/6668) about this, but I'm not sure what its conclusion means in terms of what you'd like me to do here. Should I put in more fixed limits (for globals, imports, exports, or other things) or leave it as is?

Let us leave the additions out. If it becomes obvious with evidence that we need these additional limitations we can add them at a later point, ideally together with Wasmtime. Wasm globals and functions are already kinda limited by the Wasm blob size which was the conclusion of the issue.

> One other thing I wanted to ask: in terms of documentation, the wasmtime interfaces and implementations all have fairly long (well written, nice!) doc-comments. I can copy these over verbatim -- I think the licenses allow it -- or I can rewrite them myself, either equally verbose, or just with brief summaries ... do you have a preference?

Great question! So far we avoided copying their docs and instead link to their docs for elaborated versions: for example in their `Func::wrap` API. (I hope at least haha.) So we kinda tend to be less verbose and also less complete. We usually shortly describe functionality of an API in a sentence, if needed a slight elaboration of it in a `# Note` section, potential panicking behavior in a `# Panic` section and errors that can be encountered in an `# Error` sections. But that's about it.

crates/wasmi/src/engine/executor.rs
```diff
@@ -92,15 +93,25 @@ pub enum ReturnOutcome {
...
-pub fn execute_wasm<'engine>(
-    ctx: &mut StoreInner,
+pub fn execute_wasm<'ctx, 'engine>(
+    ctx: &'ctx mut StoreInner,
     cache: &'engine mut InstanceCache,
     value_stack: &'engine mut ValueStack,
     call_stack: &'engine mut CallStack,
     code_map: &'engine CodeMap,
     const_pool: ConstPoolView<'engine>,
+    resource_limiter: ResourceLimiterRef<'ctx>,
```
Robbepop, line 103: For the sake of experimentation with benchmarks we could use `&mut ResourceLimiterRef<'ctx>` here instead since that's how it is used anyways.

crates/wasmi/src/store.rs
```diff
@@ -236,6 +268,17 @@ impl StoreInner {
             elems: Arena::new(),
             extern_objects: Arena::new(),
             fuel: Fuel::default(),
+
+            // Counts and limits of instances, memories and tables tracked by
+            // ResourceLimiter. Counts are kept separate from the sizes of arenas
+            // above in order to allow checking "up front" in a single early and
+            // fallible call to [Store::bump_resource_counts].
```
Robbepop, line 275: Cannot this check be performed uppon bumping as well with a check like

```rust
if store.memories.len() == store.memory_limit {
    return Ok(false);
}
```
?
It would be great if we could get rid of this redundant information. Always bug prone.

crates/wasmi/src/store.rs
```diff
@@ -714,6 +758,67 @@ impl<T> Store<T> {
...
+            let l = limiter(&mut self.data);
+            (l.instances(), l.tables(), l.memories())
+        };
+        inner.instance_limit = instance_limit;
+        inner.table_limit = table_limit;
+        inner.memory_limit = memory_limit;
+        self.limiter = Some(ResourceLimiterQuery(Box::new(limiter)))
+    }
+
+    pub fn bump_resource_counts(&mut self, module: &crate::Module) -> Result<(), Error> {
```
Robbepop, line 776: Instead of immediately bumping the counters we could make this method just check if creating a new instance for a Wasm module fits the limits as early check and potential bail out during the instantiation process.

This way we don't have the need for those redundant counters but also get our cheap checks. Or am I missing something?

crates/wasmi/src/store.rs
graydon: I think not "up front" / early when this method is called -- the memories haven't been allocated yet. I was trying to follow the structure of wasmtime in terms of how and when things get checked. I can try harder to get rid of this though.

crates/wasmi/src/engine/executor.rs
graydon: Will try.

crates/core/src/trap.rs
```diff
@@ -284,6 +284,13 @@ pub enum TrapCode {
     /// internal bytecode so that fuel is consumed for each executed instruction.
     /// This is useful to deterministically halt or yield a WebAssembly execution.
     OutOfFuel,
+
+    /// This trap is raised when a growth operation was attempted and an
+    /// installed `ResourceLimiter` returned `Err(...)` from the associated
+    /// `table_growing` or `memory_growing` method, indicating a desire on the
+    /// part of the embedder to trap the interpreter rather than merely fail the
+    /// growth operation.
+    GrowthOperationLimited,
```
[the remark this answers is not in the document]
graydon: Yeah. I don't love it but it seems like the structure of wasmi's errors is _just_ not-quite-fitting the pattern this interface wants that I have to adjust the errors a bit.

crates/wasmi/src/store.rs
graydon: No, I think that's possible. I can give it a try, I just was trying to follow how wasmtime did it / when it did the checks.

Robbepop: low priority issue though. highest priorities are getting CI happy and fixing those benchmarks. everything else can be done later.

crates/wasmi/src/store.rs
Robbepop: low priority issue though. highest priorities are getting CI happy and fixing those benchmarks. everything else can be done later or in a follow-up.

Robbepop:
@graydon To provide you with more information about the usual performance regressions we see when working on the engine executor of `wasmi` is that these massive slowdowns occur mostly because Rust/LLVM is somehow incapable to properly optimize the `loop+match` constructor that drives the execution.

The wanted codegen is the one that puts jumps into every match arm instead of having the jump be a central one. It is a very fragile optimization and the heuristics are frickle as you may imagine. One possible future proof solution to this miserable state is to get [explicit tail calls into the Rust language](https://github.com/rust-lang/rfcs/pull/3407) which would provide us with the control we need to properly guide the codegen towards our preferred optimized assembly.

For now the best we can do is to massage the executor code until it magically works again ... 🤦 It can be a very time consuming and frustrating experience at times.

graydon:
@Robbepop understood. I'll poke around with that. I just wanted to (a) finish what I think is functional completeness by making some tests that run and pass -- now done -- and (b) address your remaining design-choice questions. Happy to iterate on the performance regression until it's eliminated; we want it to go fast too!

crates/wasmi/src/module/instantiate/mod.rs
```diff
@@ -57,7 +57,10 @@ impl Module {
         let handle = context.as_context_mut().store.inner.alloc_instance();
         let mut builder = InstanceEntity::build(self);
 
-        context.as_context_mut().store.bump_resource_counts(self)?;
+        #[cfg(feature = "std")]
```
Robbepop, line 60: Why is this behind `std` feature?

graydon: It uses `Box` on the existential

crates/wasmi/src/store.rs
graydon: Update: this does seem to work ok if I move some stuff around, I've pushed a version that does so.

crates/wasmi/src/engine/executor.rs
graydon: Yes, this works. Pushed a version that does so. (Maybe that'll help with the benchmarks? Fingers crossed)

crates/wasmi/src/module/instantiate/mod.rs
Robbepop: Ah, for `no_std` builds in `wasmi` we simply make use of Rust's `core` and `alloc` libs. So when we are using something like `Box` we have to manually import it via `use alloc::boxed::Box;` since there is no `no_std` prelude. I know it is weird but that will fix the `no_std` build error. :)

crates/wasmi/src/store.rs
```diff
@@ -29,6 +28,8 @@ use crate::{
     TableEntity,
     TableIdx,
 };
+#[cfg(feature = "std")]
```
Robbepop, line 31: With `use alloc::boxed::Box` none of the `#[cfg(feature = "std")]` is needed since this import works both in `std` and `no_std` environments.

crates/wasmi/src/engine/executor.rs
Robbepop: Seems like it didn't help. Another thing I often succesfully try is to toy around with different kinds of `#[inline]`, `#[inline(always)]` and `#[inline(never)]` to guide the codegen towards hot patchs and cold paths. It is super ugly but sometimes it gets the job done. A perfect display of how `match+loop` isn't a good fit for efficient interpreters.

crates/wasmi/src/store.rs
graydon: got it. fixed.

graydon:
Hmm. So I took a look at the generated code and did a little profiling to see where the regression is. It doesn't look to my eyes like it's dispatch-related -- it looks like it's doing the correct tail-call / direct-threading thing in both cases. Here's a before (left) and after (right) of a single opcode block:

```
Fast                                                                                      Slow
                                                                                   |                                                                                    
  1ec280:       41 8b 46 04             mov    0x4(%r14),%eax                      |  1ec540:       8b 40 04                mov    0x4(%rax),%eax                       
  1ec284:       48 c1 e0 03             shl    $0x3,%rax                           |  1ec543:       48 8b 4c 24 18          mov    0x18(%rsp),%rcx                      
  1ec288:       48 f7 d8                neg    %rax                                |  1ec548:       48 8d 51 f8             lea    -0x8(%rcx),%rdx                      
  1ec28b:       49 8b 04 04             mov    (%r12,%rax,1),%rax                  |  1ec54c:       48 89 54 24 18          mov    %rdx,0x18(%rsp)                      
  1ec28f:       49 89 04 24             mov    %rax,(%r12)                         |  1ec551:       48 8b 51 f8             mov    -0x8(%rcx),%rdx                      
  1ec293:       49 83 c4 08             add    $0x8,%r12                           |  1ec555:       48 c1 e0 03             shl    $0x3,%rax                            
  1ec297:       49 83 c6 08             add    $0x8,%r14                           |  1ec559:       48 f7 d8                neg    %rax                                 
  1ec29b:       41 0f b6 06             movzbl (%r14),%eax                         |  1ec55c:       48 89 54 01 f8          mov    %rdx,-0x8(%rcx,%rax,1)               
  1ec29f:       49 63 04 82             movslq (%r10,%rax,4),%rax                  |  1ec561:       48 83 44 24 20 08       addq   $0x8,0x20(%rsp)                      
  1ec2a3:       4c 01 d0                add    %r10,%rax                           |  1ec567:       48 8b 44 24 20          mov    0x20(%rsp),%rax                      
  1ec2a6:       ff e0                   jmp    *%rax                               |  1ec56c:       0f b6 08                movzbl (%rax),%ecx                          
                                                                                   |  1ec56f:       48 63 4c 8d 00          movslq 0x0(%rbp,%rcx,4),%rcx                
                                                                                   |  1ec574:       48 01 e9                add    %rbp,%rcx                            
                                                                                   |  1ec577:       ff e1                   jmp    *%rcx                                
                                                                                   |                                                                                   
```

I _think_ it's just doing worse register allocation / doing more spilling and reloading, for some reason. I don't really get why, but this is at least plausibly consistent with the fact that we're passing in a new argument / changing the set of locals in this function. I will fiddle around a bit more, either this weekend or next week (it's a holiday weekend here).

Robbepop:
> Hmm. So I took a look at the generated code and did a little profiling to see where the regression is. It doesn't look to my eyes like it's dispatch-related -- it looks like it's doing the correct tail-call / direct-threading thing in both cases. Here's a before (left) and after (right) of a single opcode block:
> 
> ```
> Fast                                                                                      Slow
>                                                                                    |                                                                                    
>   1ec280:       41 8b 46 04             mov    0x4(%r14),%eax                      |  1ec540:       8b 40 04                mov    0x4(%rax),%eax                       
>   1ec284:       48 c1 e0 03             shl    $0x3,%rax                           |  1ec543:       48 8b 4c 24 18          mov    0x18(%rsp),%rcx                      
>   1ec288:       48 f7 d8                neg    %rax                                |  1ec548:       48 8d 51 f8             lea    -0x8(%rcx),%rdx                      
>   1ec28b:       49 8b 04 04             mov    (%r12,%rax,1),%rax                  |  1ec54c:       48 89 54 24 18          mov    %rdx,0x18(%rsp)                      
>   1ec28f:       49 89 04 24             mov    %rax,(%r12)                         |  1ec551:       48 8b 51 f8             mov    -0x8(%rcx),%rdx                      
>   1ec293:       49 83 c4 08             add    $0x8,%r12                           |  1ec555:       48 c1 e0 03             shl    $0x3,%rax                            
>   1ec297:       49 83 c6 08             add    $0x8,%r14                           |  1ec559:       48 f7 d8                neg    %rax                                 
>   1ec29b:       41 0f b6 06             movzbl (%r14),%eax                         |  1ec55c:       48 89 54 01 f8          mov    %rdx,-0x8(%rcx,%rax,1)               
>   1ec29f:       49 63 04 82             movslq (%r10,%rax,4),%rax                  |  1ec561:       48 83 44 24 20 08       addq   $0x8,0x20(%rsp)                      
>   1ec2a3:       4c 01 d0                add    %r10,%rax                           |  1ec567:       48 8b 44 24 20          mov    0x20(%rsp),%rax                      
>   1ec2a6:       ff e0                   jmp    *%rax                               |  1ec56c:       0f b6 08                movzbl (%rax),%ecx                          
>                                                                                    |  1ec56f:       48 63 4c 8d 00          movslq 0x0(%rbp,%rcx,4),%rcx                
>                                                                                    |  1ec574:       48 01 e9                add    %rbp,%rcx                            
>                                                                                    |  1ec577:       ff e1                   jmp    *%rcx                                
>                                                                                    |                                                                                   
> ```
> 
> I _think_ it's just doing worse register allocation / doing more spilling and reloading, for some reason. I don't really get why, but this is at least plausibly consistent with the fact that we're passing in a new argument / changing the set of locals in this function. I will fiddle around a bit more, either this weekend or next week (it's a holiday weekend here).

Maybe we could pack the `Executor` fields `value_stack`, `call_stack` and `code_map` under a common structure and only use a reference to such structure since all those three fields are only used for function calls and more or less used at roughly the same time during execution. So packing them could make sense.

graydon:
I tried something different. Just a hunch, but it seemed weird to me that it was doing all that spilling and reloading, when it should store the struct in registers -- one of LLVM's most important passes is "SROA" (scalar replacement of aggregates) where it "scalarizes" the individual fields of structs, breaking them into separate values -- which is how Executor's fields all wind up in registers. I hypothesized that the new reference was pushing the `Executor` over some kind of threshold that broke SROA. So I just moved the new reference out of the struct, as a separate argument to the function, and that seems to have helped a lot!

```
  1ec530:       41 8b 46 04             mov    0x4(%r14),%eax
  1ec534:       48 c1 e0 03             shl    $0x3,%rax
  1ec538:       48 f7 d8                neg    %rax
  1ec53b:       49 8b 04 04             mov    (%r12,%rax,1),%rax
  1ec53f:       49 89 04 24             mov    %rax,(%r12)
  1ec543:       49 83 c4 08             add    $0x8,%r12
  1ec547:       49 83 c6 08             add    $0x8,%r14
  1ec54b:       41 0f b6 06             movzbl (%r14),%eax
  1ec54f:       49 63 04 82             movslq (%r10,%rax,4),%rax
  1ec553:       4c 01 d0                add    %r10,%rax
  1ec556:       ff e0                   jmp    *%rax
```

I'll wait for the benchmark bot to run, but on my workstation it seems to solve the problem. We'll see!

Robbepop:
> I'll wait for the benchmark bot to run, but on my workstation it seems to solve the problem. We'll see!

Looking forward to the results. It is a bit unfortunate that I have to manually approve every single commit of yours before the CI will run ... cannot even change this in the GitHub settings. 🙃 This approval by maintainers seem to counter crypto miners according to GitHub. 😮‍💨

crates/wasmi/src/store.rs
```diff
@@ -26,6 +28,8 @@ use crate::{
     TableEntity,
     TableIdx,
 };
+#[cfg(not(feature = "std"))]
```
Robbepop, line 31: Due to https://github.com/paritytech/wasmi/blob/master/crates/wasmi/src/lib.rs#L85 we do not even need the `#[cfg(not(feature = "std"))]`. :) I know it isn't standard practice but it helps to get rid of tons of those `#[cfg]` attributes which are just noise at some point.

Robbepop:
Performance issue seems to have been resolved mostly. There is an outlier with `global_const`, however, it kinda is a flaky test for reasons that are not entirely clear to me unfortunately. (It doesn't do much at execution time since the `global.get` is optimized to a simple `i32.const x`.) Given that the Wasm side sees no change for `global_const` benchmark test I'd say it is an artifact.

Besides that it seems that performance slightly improved on Wasm targets which is always welcome.

Great job fixing the performance regressions @graydon ! :)

crates/wasmi/src/table/mod.rs
Robbepop: Both `MemoryEntity` and `TableEntity` are crate-private definitions therefore changing their APIs is not a concern to me.

crates/wasmi/src/store.rs
```diff
@@ -67,6 +71,28 @@ impl StoreIdx {
 /// A stored entity.
 pub type Stored<Idx> = GuardedEntity<StoreIdx, Idx>;
 
+pub struct ResourceLimiterRef<'a>(pub(crate) Option<&'a mut (dyn crate::ResourceLimiter)>);
```
Robbepop, line 74: Overlooked `pub(crate)` that could be removed from what I see.

Robbepop:
```suggestion
pub struct ResourceLimiterRef<'a>(Option<&'a mut (dyn crate::ResourceLimiter)>);
```

crates/wasmi/src/table/error.rs
```diff
@@ -38,21 +31,14 @@ pub enum TableError {
...
-                current,
-                delta,
-            } => {
-                write!(
-                    f,
-                    "tried to grow table with size of {current} and maximum of \
-                    {maximum} by {delta} out of bounds",
-                )
+            Self::GrowOutOfBounds => {
+                write!(f, "out of bounds table growth")
```
Robbepop, line 41: What is the reason this error information got removed?

graydon: There's a little bit of awkwardness in one of the call paths that returns these -- basically have to do the calculation of current and maximum in two places -- and I thought given that the `MemoryError::GrowOutOfBounds` case has no such fields, and you expressed dissatisfaction with the error structures in general, I'd just drop the extra fields. But I'm happy to revert it if you want to keep that resolution. I'll push a restoration + calculation change.

crates/wasmi/src/store.rs
graydon: Ok, ok. Removing.

crates/wasmi/src/store.rs
graydon: Sure. Added.

graydon:
I believe I've addressed all the remaining comments and CI failures and this is ready to go now.

crates/wasmi/src/table/error.rs
Robbepop: Okay, well, in all honesty I don't know if we gain too much from the lost information anyways. So do what suits you best. Fixing the error structure can be done in another PR.

crates/wasmi/src/store.rs
```diff
@@ -67,6 +70,28 @@ impl StoreIdx {
 /// A stored entity.
 pub type Stored<Idx> = GuardedEntity<StoreIdx, Idx>;
 
+pub struct ResourceLimiterRef<'a>(Option<&'a mut (dyn crate::ResourceLimiter)>);
```
Robbepop, line 73: I'd love to see a (short) doc comment explaining the rational behind the existence of this type for our future selves. :)

crates/wasmi/src/store.rs
```diff
@@ -67,6 +70,28 @@ impl StoreIdx {
...
+    }
+}
+
+impl<'a> ResourceLimiterRef<'a> {
+    pub fn as_resource_limiter(&mut self) -> &mut Option<&'a mut dyn crate::ResourceLimiter> {
+        &mut self.0
+    }
+}
+
+struct ResourceLimiterQuery<T>(
```
Robbepop, line 86: I'd love to see a (short) doc comment explaining the rational behind the existence of this type for our future selves. :)

crates/wasmi/src/store.rs
```diff
@@ -714,6 +742,62 @@ impl<T> Store<T> {
         self.data
     }
 
+    pub fn limiter(
```
Robbepop, line 745: I know it might seem annoying but for the sake of consistency I like to see doc comments even on non-public APIs. Same with the other methods below. Short description and maybe rational is enough.

crates/core/src/trap.rs
```diff
@@ -284,6 +284,13 @@ pub enum TrapCode {
     /// internal bytecode so that fuel is consumed for each executed instruction.
     /// This is useful to deterministically halt or yield a WebAssembly execution.
     OutOfFuel,
+
+    /// This trap is raised when a growth operation was attempted and an
+    /// installed `ResourceLimiter` returned `Err(...)` from the associated
```
Robbepop, line 289:
```suggestion
    /// installed [`ResourceLimiter`](crate::ResourceLimiter) returned `Err(...)` from the associated
```

Robbepop: We usually try to make use of intra-doc links in docs to improve UX for users when browsing the generated docs. Actually we even do this for crate-private docs (which is why our CI is checking it) because sometimes using `cargo doc --document-private-items` can be super helpful, too.

crates/wasmi/src/limits.rs
```diff
@@ -0,0 +1,303 @@
...
+        maximum: Option<u32>,
+    ) -> Result<bool, TableError> {
+        let allow = match self.table_elements {
+            Some(limit) if desired > limit => false,
+            _ => match maximum {
+                Some(max) if desired > max => false,
+                Some(_) | None => true,
+            },
+        };
+        if !allow && self.trap_on_grow_failure {
```
Robbepop, line 281: Technically we could check if `self.trap_on_grow_failure` was `false` and bail out early instead, right?

crates/wasmi/src/module/instantiate/mod.rs
```diff
@@ -204,17 +212,16 @@ impl Module {
...
+            .check_new_memories_limit(self.len_memories())?;
         for memory_type in self.internal_memories().copied() {
-            let memory =
-                Memory::new(context.as_context_mut(), memory_type).unwrap_or_else(|error| {
-                    panic!(
-                        "encountered unexpected invalid memory type \
-                        {memory_type:?} after Wasm validation: {error}",
-                    )
-                });
+            let memory = Memory::new(context.as_context_mut(), memory_type)?;
```
Robbepop, line 221: The `panic` was meaningful since it was a bug and not an error if this failed.

Robbepop:
Looking forward to having this merged! :)

crates/wasmi/src/store.rs
graydon: Done.

crates/wasmi/src/store.rs
graydon: Done.

crates/core/src/trap.rs
graydon: Done.

crates/wasmi/src/limits.rs
graydon: I'm not sure what you mean by early. The `allow` flag is a `bool` that has its full value -- either true or false -- returned in the false branch here. I don't think the conditional can be reorganized to be any simpler.

crates/wasmi/src/module/instantiate/mod.rs
graydon: Hah, this line is actually the entire point of the PR to me! Specifically being able to trap it nicely as an error.

I want to be able to catch (and prevent) instantiating a `Memory` that is too big according to the `ResourceLimiter`. Inside `Memory::new` (actually in its subsequent callee, `MemoryEntity::new`) the `ResourceLimiter` will return an error if it does not like the size of the memory, by policy. That's not a bug, that's a runtime limit-enforcement.

I understand that it subsumes the case that you were catching before -- if I'm reading correctly, the only bug it caught before was if you had mis-coded the validator and let through a page count that, translated to bytes, exceeds 4GiB  -- so I _think_ the thing for me to do here, if you want to preserve that cross-check, is to sink the panic down into the line in `MemoryEntity::new` where the page count is converted to bytes.

The other option I can think of is to add a special `MemoryError` that _just_ catches that case, and feed it back out to the call site in `instantiate`. This seems a little odd but it might be more appropriate if you can think of cases where `MemoryEntity::new` _is_ called (even indirectly) with a `MemoryType` that was not supposed to have already been validated.

I'm happy to go either way. Let me know!

Robbepop: Ah, I should not review PRs as late as 1AM. Forgive me, you are obviously totally right. The change is as intended by the PR.

crates/wasmi/src/limits.rs
Robbepop: Nevermind, shouldn't review when it is so late. 🙃

Robbepop:
The PR is ready to go once the remaining CI issues are fixed. Just a broken doc link and a new `clippy` warning from the newest clippy nightly that just got introduced. :)

crates/wasmi/src/store.rs
graydon: Oh yeah sorry missed this one! Fixed.

Robbepop: Pushed the fix already? :)
This PR still needs resolution for the CI.

graydon:
@Robbepop one more try! :)

Robbepop:
> @Robbepop one more try! :)

worked :)

Robbepop:
CI seems to be happy now so I guess we can merge this PR @graydon.
@graydon Great work and thanks a lot for the contribution. Very much appreciated!

PR #740, opened 2023-07-24
Publishing benchmarks for graphs

alvicsam:
PR adds gitlab and github jobs that run benchmarks for master and create nice graphs that can be found [here](https://paritytech.github.io/wasmi/benchmarks/)

close https://github.com/paritytech/ci_cd/issues/848

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -9,7 +9,7 @@ default:
     max: 2
 
 .docker-env:
-  image:                           "paritytech/ci-linux:staging"
+  image:                           "paritytech/ci-unified:bullseye-1.71.0"
   interruptible:                   true
   tags:
     - benchmark
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -27,6 +27,14 @@ default:
     paths:
       - target/ci/
 
+.common-before:
+  before_script:
+    # Show Versions
+    - cargo --version
+    - rustc --version
+    # Create dir for artifacts
+    - mkdir -p ./target/ci
+
 .pr-refs:
   rules:
     - if: $CI_COMMIT_REF_NAME =~ /^[0-9]+$/                   # PRs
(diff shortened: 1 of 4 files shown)


Robbepop:
@alvicsam thanks for implementing this so quickly and sorry for replying to late!

The graphs look pretty cool and the website display is clean.

I have a few extra requests if possible. 😇

- Can I haz dark mode eventually (if not much work)? 😅
- In most/all graphs the ns/iter column hosts a ton of zeros. Maybe if we use microseconds per iteration (us/iter) it would look a bit less overwhelming.
- Maybe it a local problem with my screen but only at 50% zoom level I have a decent level of zoom on the graphs. At 100% I am basically looking at a single graph per page and I think this is simply too big given that we have so many of those graphs. We definitely want a zoom, but default zoom should provide a nice overview as in this picture:

image

Robbepop:
> > Can I haz dark mode eventually (if not much work)? 😅
> 
> Unfortunately I couldn't find this option in the GHA

Would it maybe be possible to write feature request issues for this? Not that important though.

> > In most/all graphs the ns/iter column hosts a ton of zeros. Maybe if we use microseconds per iteration (us/iter) it would look a bit less overwhelming.
> 
> With this action it's not possible, it's [hardcoded](https://github.com/benchmark-action/github-action-benchmark/blob/master/src/extract.ts#L335) there

Hmm, again I think we could write feature request issues since the ns/iter is really not a great default for everything anyways imo. Sometimes for example you measure throughput or measure network stuff which usually goes with ms/iter.

.gitlab-ci.yml
```diff
@@ -8,40 +8,55 @@ default:
   retry:
     max: 2
 
-.docker-env:                       &docker-env
-  image:                           "paritytech/ci-linux:staging"
+.docker-env:
+  image:                           "paritytech/ci-unified:bullseye-1.71.0"
```
Robbepop, line 12: What are the effects of this and why was this changed/needed?

Robbepop (commented):
Thanks a lot for implementing this so quickly @alvicsam! Sorry for taking unusually long to reply and review.

.gitlab-ci.yml
sergejparity: We replaced `ci-linux` with `ci-unified`. It has transparent tags naming, flexible choice of 
Rust versions, and so on. More - https://hub.docker.com/repository/docker/paritytech/ci-unified

alvicsam: We are switching to our new `ci-unified` image. We decided to have "one for all" image and finally added versioning to it.

alvicsam:
> @alvicsam Is this ready for merge?

yes

Robbepop:
Thanks a lot @alvicsam for this! 🚀 
Looking forward to seeing the benchmark graphs over time.

PR #778, opened 2023-10-14
Update Wasm proposal support in README

Robbepop:


Robbepop:
All CI jobs but the benchmarks are fixed again. I have a hunch that the benchmarks CI jobs are only failing for `master` but not for the PR branch. So merging this PR might fix CI jobs, too.

PR #779, opened 2023-10-16
Fix typos

vuittont60:


Robbepop:
@vuittont60 Thanks a lot for the type fixes! Unfortunately the CI is currently broken so we might need to merge this without proper CI checks until the CI is fixed. However, this shouldn't be a problem for type fixes. :)

PR #938, opened 2024-02-19
Fix consume_fuel doc

pgherveou:
The doc mentions that `consume_fuel` would panic if it overflows the fuel counter, but we are using [checked_math under the hood ](https://github.com/wasmi-labs/wasmi/blob/e011f7d1070aa90c85fbcf0d6835a7104d1aba17/crates/wasmi/src/store.rs#L299-L305) so not sure if this is accurate.


Robbepop (approved):
@pgherveou Thank you for the doc fixes. Upon inspection seems to be right. :)

PR #939, opened 2024-02-22
feat: add static buffer as an alternative to allocating memory

### LINKED ISSUE #878: Support for Alternative Memory Backing Options
I'm currently using WASMI for a project on an embedded target and am exploring
ways to optimize memory management. The standard `Vec<u8>` [used](https://github.com/paritytech/wasmi/blob/d80c19fa51f34bcb7aa09a11c3f5017cb90fa2c6/crates/wasmi/src/memory/buffer.rs#L12)
for wasm memory in can be restrictive in embedded contexts, where heap
allocations are often less desirable. My aim is to have an option for using a
static array, or a similar lean memory structure, in place of `Vec<u8>`.

Would it be possible to consider an alternative to `Vec<u8>`, maybe through a
trait implementation that allows for different memory storage types? This
addition would not only help my case but could also open up new possibilities
for WASMI in various specialized environments.

Ddystopia:
Draft for static buffer as an alternative to allocating memory. Closes #878.

User can give `fn() -> Option<&'static mut [u8]>`, which is a factory for static buffers. I personally only need one, but there are problems if we want user to provide `&'static mut [u8]`. It is not `Copy` or `Clone` for obvious reasons, but `Config` is. It also cannot be wrapped in `RefCell` to be taken from config by shared reference, because there is a `Sync` assert. 

In my code, this feature could used in 100% safe way:
```rust
fn static_buffer_factory() -> Option<&'static mut [u8]> {
    #[repr(align(32))]
    struct Memory([u8; 64 * 1024]);

    // Now the linker will refuse to create the final binary
    // if there is not enough space in the target RAM
    static WASMI_LINEAR_MEMORY: StaticCell<Memory> = StaticCell::new();

    // Will only succeed the first time. Uses atomic CAS operations under the hood
    WASMI_LINEAR_MEMORY
        .try_init_with(|| Memory([0; 64 * 1024]))
        .map(|x| &mut x.0[..])
}

let mut config = Config::default();
let config = config.set_static_buffer_factory(static_buffer_factory);
let engine = Engine::new(config);

```

As can be seen from godbolt, rustc optimizes out enum access in the hot methods `data`, `data_mut` and `len`, so there is no need for unsafe code and enum is sufficient.

wasmi_godbolt

## Unresolved questions
- Are you okay with that kind of design?
- Should we strengthen/weaken the requirements? For example, add an alignment check?

### DIFF
--- a/crates/wasmi/src/memory/buffer.rs
+++ b/crates/wasmi/src/memory/buffer.rs
@@ -1,6 +1,7 @@
+use core::mem::ManuallyDrop;
 use std::{vec, vec::Vec};
 
-/// A `Vec`-based byte buffer implementation.
+/// A byte buffer implementation.
 ///
 /// # Note
 ///
@@ -9,8 +10,15 @@ use std::{vec, vec::Vec};
 /// solution fitting any platform.
 #[derive(Debug)]
 pub struct ByteBuffer {
-    bytes: Vec<u8>,
+    ptr: *mut u8,
+    len: usize,
+    capacity: usize,
+    is_static: bool,
 }
+
+// Safety: `ByteBuffer` is essentially an enum of `Vec<u8>` or `&'static mut [u8]`.
+// They both are `Send` so this is sound.
+unsafe impl Send for ByteBuffer {}
 
 impl ByteBuffer {
     /// Creates a new byte buffer with the given initial length.
@@ -20,8 +28,30 @@ impl ByteBuffer {
     /// - If the initial length is 0.
     /// - If the initial length exceeds the maximum supported limit.
     pub fn new(initial_len: usize) -> Self {
+        let mut vec = ManuallyDrop::new(vec![0x00_u8; initial_len]);
+        let (ptr, len, capacity) = (vec.as_mut_ptr(), vec.len(), vec.capacity());
         Self {
-            bytes: vec![0x00_u8; initial_len],
+            ptr,
+            len,
+            capacity,
+            is_static: false,
+        }
+    }
+
+    /// Creates a new byte buffer with the given initial length.
+    ///
+    /// # Errors
+    ///
+    /// - If the initial length is 0.
+    /// - If the initial length exceeds the maximum supported limit.
+    pub fn new_static(buf: &'static mut [u8], initial_len: usize) -> Self {
+        assert!(initial_len <= buf.len());
+        buf.fill(0);
+        Self {
+            ptr: buf.as_mut_ptr(),
+            len: initial_len,
+            capacity: buf.len(),
+            is_static: true,
         }
     }
 
@@ -29,24 +59,120 @@ impl ByteBuffer {
     ///
     /// # Panics
     ///
-    /// If the current size of the [`ByteBuffer`] is larger than `new_size`.
+    /// - If the current size of the [`ByteBuffer`] is larger than `new_size`.
+    /// - If backed by static buffer and `new_size` is larger than it's capacity.
     pub fn grow(&mut self, new_size: usize) {
         assert!(new_size >= self.len());
-        self.bytes.resize(new_size, 0x00_u8);
+        if self.is_static {
+            if self.capacity < new_size {
+                panic!("Cannot grow static byte buffer more then it's capacity")
+            }
+            self.len = new_size;
+        } else {
+            // Safety: those parts have been obtained from `Vec`.
+            let vec = unsafe { Vec::from_raw_parts(self.ptr, self.len, self.capacity) };
+            let mut vec = ManuallyDrop::new(vec);
+            vec.resize(new_size, 0x00_u8);
+            let (ptr, len, capacity) = (vec.as_mut_ptr(), vec.len(), vec.capacity());
+            self.ptr = ptr;
+            self.len = len;
+            self.capacity = capacity;
+        }
     }
 
     /// Returns the length of the byte buffer in bytes.
     pub fn len(&self) -> usize {
-        self.bytes.len()
+        self.len
     }
 
     /// Returns a shared slice to the bytes underlying to the byte buffer.
     pub fn data(&self) -> &[u8] {
-        &self.bytes[..]
+        // Safety: either is backed by a `Vec` or a static buffer, ptr[0..len] is valid.
+        unsafe { core::slice::from_raw_parts(self.ptr, self.len) }
     }
 
     /// Returns an exclusive slice to the bytes underlying to the byte buffer.
     pub fn data_mut(&mut self) -> &mut [u8] {
-        &mut self.bytes[..]
+        // Safety: either is backed by a `Vec` or a static buffer, ptr[0..len] is valid.
+        unsafe { core::slice::from_raw_parts_mut(self.ptr, self.len) }
+    }
+}
+
+impl Drop for ByteBuffer {
+    fn drop(&mut self) {
+        if !self.is_static {
+            // Safety: those parts have been obtained from `Vec`.
+            unsafe { Vec::from_raw_parts(self.ptr, self.len, self.capacity) };
+        }
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    #[test]
+    fn test_basic_allocation_deallocation() {
+        let buffer = ByteBuffer::new(10);
+        assert_eq!(buffer.len(), 10);
+        // Dropping the buffer should not cause UB.
+    }
+
+    #[test]
+    fn test_basic_data_manipulation() {
+        let mut buffer = ByteBuffer::new(10);
+        assert_eq!(buffer.len(), 10);
+        let data = buffer.data(); // test we can read the data
+        assert_eq!(data, &[0; 10]);
+        let data = buffer.data_mut(); // test we can take a mutable reference to the data
+        data[4] = 4; // test we can write to the data and it is not UB
+        let data = buffer.data(); // test we can take a new reference to the data
+        assert_eq!(data, &[0, 0, 0, 0, 4, 0, 0, 0, 0, 0]); // test we can read the data
+                                                           // test drop is okay
+    }
+
+    #[test]
+    fn test_static_buffer_initialization() {
+        static mut BUF: [u8; 10] = [7; 10];
+        let buf = unsafe { &mut *core::ptr::addr_of_mut!(BUF) };
+        let mut buffer = ByteBuffer::new_static(buf, 5);
+        assert_eq!(buffer.len(), 5);
+        // Modifying the static buffer through ByteBuffer and checking its content.
+        let data = buffer.data_mut();
+        data[0] = 1;
+        unsafe {
+            assert_eq!(BUF[0], 1);
+        }
+    }
+
+    #[test]
+    fn test_growing_buffer() {
+        let mut buffer = ByteBuffer::new(5);
+        buffer.grow(10);
+        assert_eq!(buffer.len(), 10);
+        assert_eq!(buffer.data(), &[0; 10]);
+    }
+
+    #[test]
+    fn test_growing_static() {
+        static mut BUF: [u8; 10] = [7; 10];
+        let buf = unsafe { &mut *core::ptr::addr_of_mut!(BUF) };
+        let mut buffer = ByteBuffer::new_static(buf, 5);
+        assert_eq!(buffer.len(), 5);
+        assert_eq!(buffer.data(), &[0; 5]);
+        buffer.grow(8);
+        assert_eq!(buffer.len(), 8);
+        assert_eq!(buffer.data(), &[0; 8]);
+        buffer.grow(10);
+        assert_eq!(buffer.len(), 10);
+        assert_eq!(buffer.data(), &[0; 10]);
+    }
+
+    #[test]
+    #[should_panic]
+    fn test_static_buffer_overflow() {
+        static mut BUF: [u8; 5] = [7; 5];
+        let buf = unsafe { &mut *core::ptr::addr_of_mut!(BUF) };
+        let mut buffer = ByteBuffer::new_static(buf, 5);
+        buffer.grow(10); // This should panic.
     }
 }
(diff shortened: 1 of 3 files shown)


Ddystopia:
fixes #878

Robbepop:
@Ddystopia Thank you a lot for the PR!

I have to make sure (at least locally) that these changes indeed won't have an effect on the performance of Wasmi for inputs that should have a large impact on linear memory accesses.

I am not yet sold on the config addition but it seems to be necessary to allow for non-imported linear memory backed by static buffers. Is this a critical feature to you? Could we maybe separate this out to make this feature a bit smaller and only support imported linear memory for now?

Ddystopia:
Thank you for the response! I don't actually know, would imported memory be sufficient or not. Is there a way to compile Rust to wasm so that it would only import memory and not export any, do that wasmi would not create them on it's own?

Ddystopia:
Force-pushed to remove config changes. This is now only available for imported memories.

Ddystopia:
Force pushed to fix formatting and clippy jobs

Robbepop:
> I'm not sure what should I do with that clippy job - it would not compile if I remove those 2 imports. They're probably unused under some combination of features. Should I gate those imports?

I think it is a `clippy` or even Rust compiler bug. Nothing to do here for us at the moment.

crates/wasmi/src/memory/mod.rs
```diff
@@ -165,6 +165,49 @@ impl MemoryEntity {
         }
     }
 
+    /// Creates a new memory entity with the given memory type.
+    pub fn new_static(
```
Robbepop, line 169: The main design question I have is the following: Shall it be possible for a static memory based `Memory` instance to be growable? Technically this does not make sense since the buffer is static anyways. Also for certain use cases such as when fuel metering is enabled the advantage of ungrowable static memory is that `memory.grow` would cost O(1) fuel guaranteed.
This requires `initial_pages == current_pages == static_buffer.len() * BYTES_PER_PAGE` always for static memory based `Memory` instances.

Is there a use case that would suffer from this design decision?

With the current design the static memory buffer already is _not_ growable (obviously), however `Memory::grow` still performs (hopeful) work to do so.

Robbepop: In case we want to change growability of static memory based `Memory` instances we should reflect that in the documentation of the `new_static` constructor.

Ddystopia: Some WASM allocators are requesting pages (minimum 1), but I'm not sure would it get one if memory is already at max capacity, do you know? Maybe if memory would be allowed to grow, they can get their page. That is the only possible problem I can think of, I don't know is it true though

Robbepop: The `memory.grow` instruction cannot grow linear memory if it is already at its maximum size. However, failing to grow linear memory is something that should be expected hence failure won't trap but return a magic value which can (and should) be handled by a memory allocator. I am not sure how much memory allocators rely on growing linear memory but it would be very good to know for sure.

Robbepop: I found an answer to our problem. For static memories we need to zero all the bytes of the new pages whenever we grow it. This is done implicitly with the non-static buffer backend (via `Vec`) but needs to be done with the static buffer upon grow because we are not guaranteed that the provided static buffer is already zeroed. And then we resolved this question because now it makes sense to do linear work for growing memories backed by a static buffer.

Robbepop (changes requested):

crates/wasmi/src/memory/mod.rs
Robbepop: This needs to be done for both, growing memory and when creating new linear memory from a static buffer.
When growing memory bytes for the new pages need to be zeroed and when creating a new linear memory from a static buffer the initial pages need to be zeroed.

Robbepop: More info: https://webassembly.github.io/spec/core/exec/modules.html#growing-memories

Ddystopia: If wasmi will zero those bytes, maybe request `&'static mut [MaybeUninit<u8>]`?

Ddystopia: I guess keeping everything safe and simple would be better, because zeroing would probably be very fast and quite infrequent, but it will make *some* difference because static memory is not required to be initialized.

Robbepop: Okay then please go ahead and implement the changes with `&'mut static [u8]` and we can always change it later to `&'static mut [MaybeUninit<u8>]` if we see a good enough reason to do so. :)

Ddystopia:
Unfortunately, compiler did not manage to optimize that time when I've added a len to the static buffer, so I went with your initial idea to use vec's raw parts. There are also some tests that I tried with miri and they do pass.

Additionally rebased to main.

crates/wasmi/src/memory/buffer.rs
```diff
@@ -20,33 +27,157 @@ impl ByteBuffer {
     /// - If the initial length is 0.
     /// - If the initial length exceeds the maximum supported limit.
     pub fn new(initial_len: usize) -> Self {
+        let mut vec = vec![0x00_u8; initial_len];
+        let (ptr, len, capacity) = (vec.as_mut_ptr(), vec.len(), vec.capacity());
+        core::mem::forget(vec);
```
Robbepop, line 32:
```suggestion
        let mut vec = ManuallyDrop::new(vec![0x00_u8; initial_len]);
        let (ptr, len, capacity) = (vec.as_mut_ptr(), vec.len(), vec.capacity());
```

Robbepop: This is how it is implemented in https://doc.rust-lang.org/src/alloc/vec/mod.rs.html#822 and probably a bit safer. We might need to import `ManuallyDrop` at the top though.

crates/wasmi/src/memory/buffer.rs
```diff
@@ -20,33 +27,157 @@ impl ByteBuffer {
...
+    /// - If the initial length exceeds the maximum supported limit.
+    pub fn new_static(buf: &'static mut [u8], initial_len: usize) -> Self {
+        assert!(initial_len <= buf.len());
+        Self {
+            ptr: buf.as_mut_ptr(),
+            len: initial_len,
+            capacity: buf.len(),
+            is_static: true,
         }
     }
```
Robbepop, line 55: We want to zero the bytes of the static buffer in this method for `0..initial_len` as the same behavior applies to its sibling `ByteBuffer::new` constructor in the `vec!` macro invocation.

crates/wasmi/src/memory/buffer.rs
```diff
@@ -20,33 +27,157 @@ impl ByteBuffer {
...
+            self.len = new_size;
+        } else {
+            // Safety: those parts have been obtained from `Vec`.
+            let mut vec = unsafe { Vec::from_raw_parts(self.ptr, self.len, self.capacity) };
+            vec.resize(new_size, 0x00_u8);
+            let (ptr, len, capacity) = (vec.as_mut_ptr(), vec.len(), vec.capacity());
+            self.ptr = ptr;
+            self.len = len;
+            self.capacity = capacity;
+            core::mem::forget(vec);
```
Robbepop, line 84:
```suggestion
            // Safety: those parts have been obtained from `Vec`.
            let mut vec = ManuallyDrop::new(unsafe { Vec::from_raw_parts(self.ptr, self.len, self.capacity) });
            vec.resize(new_size, 0x00_u8);
            let (ptr, len, capacity) = (vec.as_mut_ptr(), vec.len(), vec.capacity());
            self.ptr = ptr;
            self.len = len;
            self.capacity = capacity;
```

Robbepop: At this point I think we should simply add a `vec_into_raw_parts` method and reuse it in both locations since the one in the Rust standard library is unfortunately not yet stable.

crates/wasmi/src/memory/buffer.rs
```diff
@@ -20,33 +27,157 @@ impl ByteBuffer {
...
+        let buf = unsafe { &mut *core::ptr::addr_of_mut!(BUF) };
+        let mut buffer = ByteBuffer::new_static(buf, 5);
+        assert_eq!(buffer.len(), 5);
+        // Modifying the static buffer through ByteBuffer and checking its content.
+        let data = buffer.data_mut();
+        data[0] = 1;
+        unsafe {
+            assert_eq!(BUF[0], 1);
+        }
+    }
```
Robbepop, line 150: This test should probably initialize the static buffer with `[0x1; 10]` and check if `ByteBuffer::new_static` successfully initializes all the bytes to `0x0`.

Robbepop: Could also be yet another test, too.

crates/wasmi/src/memory/buffer.rs
```diff
@@ -20,33 +27,157 @@ impl ByteBuffer {
...
+            if self.capacity < new_size {
+                panic!("Cannot grow static byte buffer more then it's capacity")
+            }
+            // Safety:
+            // - self.ptr.add(self.len) is still in bounds of the buffer.
+            // - we are allowed to write to the buffer anywhere between `len` and `capacity`.
+            unsafe {
+                core::ptr::write_bytes(self.ptr.add(self.len), 0x00_u8, new_size - self.len);
+            }
+            self.len = new_size;
```
Robbepop, line 75: We can remove the `unsafe` code usage here by simply reusing `data_mut` here.

Robbepop, line 75:
```suggestion
            let old_size = self.len;
            self.len = new_size;
            let data = self.data_mut()[old_size..].fill(0x0);
```

Robbepop: I have not checked if the above code actually works. Please check yourself when you apply the changes. :)

crates/wasmi/src/memory/buffer.rs
```diff
@@ -20,33 +27,157 @@ impl ByteBuffer {
...
+        let mut buffer = ByteBuffer::new_static(buf, 5);
+        assert_eq!(buffer.len(), 5);
+        assert_eq!(buffer.data(), &[0; 5]);
+        buffer.grow(8);
+        assert_eq!(buffer.len(), 8);
+        assert_eq!(buffer.data(), &[0; 8]);
+        buffer.grow(10);
+        assert_eq!(buffer.len(), 10);
+        assert_eq!(buffer.data(), &[0; 10]);
+    }
```
Robbepop, line 173: Grow tests should probably have a non-zero static buffer and check if growing zeros the correct regions.

crates/wasmi/src/memory/buffer.rs
Ddystopia: Thank you!

crates/wasmi/src/memory/buffer.rs
```diff
@@ -20,33 +28,151 @@ impl ByteBuffer {
...
+
+    /// Creates a new byte buffer with the given initial length.
+    ///
+    /// # Errors
+    ///
+    /// - If the initial length is 0.
+    /// - If the initial length exceeds the maximum supported limit.
+    pub fn new_static(buf: &'static mut [u8], initial_len: usize) -> Self {
+        assert!(initial_len <= buf.len());
+        buf.fill(0);
```
Robbepop, line 49: Only fill `buf[0..initial_len]`

crates/wasmi/src/memory/buffer.rs
```diff
@@ -20,33 +28,151 @@ impl ByteBuffer {
...
+    /// - If the current size of the [`ByteBuffer`] is larger than `new_size`.
+    /// - If backed by static buffer and `new_size` is larger than it's capacity.
     pub fn grow(&mut self, new_size: usize) {
         assert!(new_size >= self.len());
-        self.bytes.resize(new_size, 0x00_u8);
+        if self.is_static {
+            if self.capacity < new_size {
+                panic!("Cannot grow static byte buffer more then it's capacity")
+            }
+            self.len = new_size;
```
Robbepop, line 70: zeroing is missing here

Robbepop (commented):
From what I can see now only the proper zeroing of bytes on `new_static` and `grow` is missing. We want to zero only the `initial_len` for `new_static` and only the new parts in `grow` to limit the amount of work needed per operation to the minimum.

Also please do not rebase this PR since this kinda destroys my previous reviews and makes it really hard for me to keep track of what happened.

crates/wasmi/src/memory/buffer.rs
Ddystopia: The whole buffer is now zeroed in new_static

crates/wasmi/src/memory/buffer.rs
Ddystopia: Is there any benefit to increase complexity?

crates/wasmi/src/memory/buffer.rs
Robbepop: Yes but this is not ideal imo: https://github.com/wasmi-labs/wasmi/pull/939#pullrequestreview-1935266425

crates/wasmi/src/memory/buffer.rs
Robbepop: Yes, for the sake of fuel metering we charge linearly for `memory.grow` anyways, so it is wise to actually distribute the work to zero the memory instead of doing it up front which might be not desirable for large static buffers in all cases.

Ddystopia:
Okay, I'll implement it tomorrow as a separate commit.

Robbepop (approved):
Looks good to me now! :)
The new tests pass and the CI is happy.

Thanks a lot for the PR and the new feature with static buffer backed linear memories in Wasmi. I am sure this feature is going to be very helpful for the embedded community. :)

🚀

PR #947, opened 2024-03-05
Update link to discussions board to avoid redirect

lygstate:


Robbepop:
Thanks a lot for the fix @lygstate !

PR #964, opened 2024-03-24
Reorder Wasmi instruction bytecodes

Robbepop:
This experiment is to see how reordering affects performance for Wasmi.

On my local machine I mostly see regressions and a few minor improvements for call intense workloads. Maybe this could be fixed with some inline annotation adjustments.


Robbepop:
My findings so far:

Rust 1.77 generally seems to regress performance by roughly 15% for Wasmi across the board.

- On Rust 1.77 this PR does not really affect performance significantly.
- On Rust nightly (two stable releases from now) this PR improves performance for some cases a bit.

Additionally this PR reorders bytecodes in a way that is more logical for general use cases.
All in all I think this PR is a win and thus we continue to merge.

PR #965, opened 2024-03-24
Export ReadError

orsinium:
Currently, `Read` cannot be implemented outside of the crate because the `read` method signature uses `ReadError` which is public but not exported. The PR exports it.


Robbepop:
@orsinium good catch! thanks for the fix. :)

Robbepop:
the `clippy` CI error is not connected to this PR

PR #969, opened 2024-03-25
Merge copy instructions if possible

Robbepop:
Closes https://github.com/wasmi-labs/wasmi/issues/946.

This merges two `copy` instructions to a `copy2` instruction if possible.

Locally I saw ~20% perf improvement for the `fibonacci/iter` benchmark test case which actually profits from this optimization since it features 2 consecutive `(local.set n (local.get m))` in its `.wat` code:

image


Robbepop:
The `miri` CI job fails because of this issue: https://github.com/rust-lang/miri/issues/3404

PR #989, opened 2024-04-20
Optimize `wasmi::Linker` host function setup

Robbepop:
Closes https://github.com/wasmi-labs/wasmi/issues/979.

cc @athei 

Most inefficiencies are in `StringInterner::get_or_insert` and in `memcmp` calls to compare definition names.

image

## Findings

- By using an empty `""` name for the `module` name for all host functions we can improve performance by ~15%.
    - This is likely due to a fast check for empty-string to avoid having to use `memcmp`.
- Introduction of `struct LenOrder(Arc<str>)` and `struct LenOrderStr(str)` wrapper types with new `Ord` implementation to avoid `memcmp` are very effective, improving performance by roughly 50%. Unfortunately this requires a single line of `unsafe` Rust code since Rust's `BTreeMap` does not support custom comparators.

image

### DIFF
--- a/crates/wasmi/benches/benches.rs
+++ b/crates/wasmi/benches/benches.rs
@@ -72,7 +72,11 @@ criterion_group!(
         .warm_up_time(Duration::from_millis(1000));
     targets =
         bench_linker_setup_same,
+        bench_linker_build_finish_same,
+        bench_linker_build_construct_same,
         bench_linker_setup_unique,
+        bench_linker_build_finish_unique,
+        bench_linker_build_construct_unique,
 );
 criterion_group! {
     name = bench_execute;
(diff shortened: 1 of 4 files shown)


Robbepop:
The new experimental `LinkerBuilder` API benchmarks show that `LinkerBuilder::finish` is up to 6 times faster than re-constructing the `Linker` which is a great speed-up for certain use cases.
The downside is that this requires to build up the `LinkerBuilder` beforehand which is roughly as slow as building up a normal `Linker`. Thus if a user now builds all `Linker` using the new `LinkerBuilder` but only actually builds up a single `Linker` this is going to be roughly 20% slower.

PR #1028, opened 2024-05-11
Split up `miri` CI job into `miri` and `miri-spec`

Robbepop:
The `miri` CI job was taking too much time before. The 2 jobs should take roughly similar time and thus "half" the overall time it takes to check Wasmi via `miri`.


Robbepop:
Conclusion:

- The single `miri` job took roughly 9 min before this PR
- The two `miri` and `miri-spec` jobs each take 4-5 min with this PR.

Thus we saved roughly half of the time as expected.

PR #1042, opened 2024-05-22
engine/config: fix wasm_tail_call and wasm_extended_const docs

airwoodix:
The defaults for `Config::tail_call` and `Config::extended_const` were changed in #1031 but the change wasn't reflected in the documentation.

Also fix the markdown reference for the `extended-const` proposal link.


Robbepop:
@airwoodix great catch! Thanks for the fix. :) Will merge once CI is green.

PR #1046, opened 2024-05-28
Prepare release for Wasmi `v0.32.0` stable

Robbepop:
Closes https://github.com/wasmi-labs/wasmi/issues/999.


Robbepop:
Somehow the `Miri (Spec)` job fails even though there was just a timeout which is fine and to be expected. Need to investigate later why the job now started to fail for timeouts.

PR #1062, opened 2024-06-11
Optimize memory access in the Wasmi executor

Robbepop:
This has fewer indirections when using the default linear memory. Local perf measurements indicated up to 20% improvements for memory-intense workloads. Unfortunately it seems that call-intense workloads regress.


Robbepop:
With Rust 1.79 this PR no longer has great effects on the benchmarks so I guess it is fine to merge it and have it as baseline for both https://github.com/wasmi-labs/wasmi/pull/1060 and https://github.com/wasmi-labs/wasmi/pull/1065.

PR #1067, opened 2024-06-14
Define `From` impls for index type conversion to `u32`

Robbepop:


Robbepop:
The failing `clippy` job triggers for something that has already been fixed on `master` branch so it should be fine to merge if everything else is fine.

PR #1075, opened 2024-06-17
Use `CachedMemory` in bulk-ops executors

Robbepop:
Locally this yielded a performance improvement for our bulk-ops benchmark test case by roughly 3.5%.
The local test uses `memory.copy` and `memory.fill` with 5000 bytes per operation.
Note that performance wins for bulk-ops with smaller lengths are going to be way more significant.


Robbepop:
Unfortunately, despite the triviality of the changes introduced by this PR local benchmarks indicate extreme performance regressions of 15-20% across the board. Probably the Rust/LLVM optimization heuristics for optimizing the loop+match construct in the interpreter's hot path got confused by this ... again.

Robbepop:
https://github.com/wasmi-labs/wasmi/pull/1075/commits/afc3c5f82630346c9845a1cb8667fb580c684d00 helped to eliminate all locally observed performance regressions.

PR #1098, opened 2024-07-01
Use `rust-cache` GitHub action in Wasmi's CI

Robbepop:
The problem with the current `cargo-cache` setup is that it still re-compiles all crates each run and only avoids re-downloading all the dependencies.
The `rust-cache` GitHub action promises to also avoid re-compiling all dependencies.

Link: https://github.com/Swatinem/rust-cache


Robbepop:
Results: we dropped CI times from sometimes well over 15 min to roughly 4-5 minutes.
Especially the test jobs improved substantially and are now at around 1-2 minutes due to non-optimizing test builds.

PR #1116, opened 2024-07-07
Allow Wasm module instantiation in host functions called from Wasmi's executor

Robbepop:
So far it was not possible to instantiate a Wasm module in a host function that has been called from Wasmi's executor.
This was because of a dead lock that occurred due to a `RwLock<EngineResources>` which contained the `FuncTypeRegistry` which was (in the end) mainly used for resolving the number of parameter and result types of a called host function in the Wasmi executor.

There are two ways of solving this with the following change:

Remove `EngineResources` and instead `RwLock` both `CodeMap` and `FuncTypeRegistry` themselves.

1. Continue to use `&CodeMap` in the Wasmi executor but now also a `RwLock<FuncTypeRegistry>`. Unlock the registry very shortly to resolve the number of parameters and results whenever a host function is called which should occur kind of rarely. The downside is that this probably regresses performance of host function calls due to the additional locking per call.
2. Populate `HostFuncEntity` with inline information about the number of its parameter types and result types. This acts as chace. Since `HostFuncEntity` has to be resolved anyway for a host function call no performance regressions are expected with this solution. The downside is increased space requirements for the `FuncEntity` buffer.

This PR chose 2. for its implementation. Before merging we might want to experiment with another PR that tried out 1. Local benchmarks confirmed that there were no significant changes.

## ToDo

- [ ] Try to reduce space regression by using `u16` instead of `usize` to store the inline number of parameters and results in `HostFuncEntity`. This is safely possible since we limit the number of both values between `0..=1000` similar to Wasmtime.

### DIFF
--- a/crates/wasmi/src/engine/executor/instrs.rs
+++ b/crates/wasmi/src/engine/executor/instrs.rs
@@ -20,10 +20,9 @@ use crate::{
             TableIdx,
             UnaryInstr,
         },
-        code_map::InstructionPtr,
+        code_map::{CodeMap, InstructionPtr},
         executor::stack::{CallFrame, FrameRegisters, ValueStack},
         DedupFuncType,
-        EngineResources,
     },
     memory::DataSegment,
     module::DEFAULT_MEMORY_INDEX,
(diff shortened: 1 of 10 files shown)


Robbepop:
I have now refactored `FuncType` to enforce maximum 1000 parameters and results.

This allows us to use `u16` values internally to represents the lengths and thus we can easily use `u16` instead of `usize` in the `HostFuncEntity` to inline-cache the number of parameters and results to save lots of space.

Besides that this PR has the added side benefit of improving the performance of host function calls significantly:
image

The reason is because we no longer need to `RwLock` to resolve a deduplicated function type with these changes.

Issue #631, opened 2023-01-23
Cannot compile a Wasm module from a called host function using the same `Engine`


lrubasze:
The issue is quite similar to #567 
But instead of calling a WASM function we try to
- create another WASM module using the same `Engine`
or
- wrap `Func::wrap` host function for while preinstantiating already created WASM module
```
         lock                                 deadlock
host ------------> WASM #1 ------------> host ------------> create WASM module or wrap `Func::wrap` host function for already created WASM module
```

Would it be possible to fix this?

Robbepop:
This is due to the fact that in `wasmi` the `Engine` is used to allocate function signatures and constants as `EngineResources` instead of the Wasm `Store` for performance reasons. It is simply more efficient for the `Engine` executor to have the resources right away. However, this means that we need to lock the `Engine` while compiling a Wasm module for a bit. And while executing a Wasm function we need to lock it for readability.
We could fix this in several different ways:

1) Move all the stuff into the `Store` which will cause overhead when executing Wasm functions.
2) Make the allocations during Wasm module compilation atomic which would cause overhead for both compilation and execution.
3) Unlock the `EngineResources` before making a call to the host function. This would certainly add overhead for host function calls and maybe even for Wasm function calls.

There might be other solutions.

lrubasze:
Hi @Robbepop!
thanks for quick reply.
The main scenario for us us is the second one in fact (sorry, should have mentioned it in a first place).
Here are more details.
In order to save time we cache compiled WASM modules to be reused later when instantiating another instance. And one of our use cases assumes instantiating from host function invoked from WASM instance.

Robbepop:
As a work around what you could do is the following:

You could resumably call the Wasm functions and instead of compiling a new Wasm module in a host function you instead return an error. Then the resumable function will return and unlock everything while doing so. This allows you to compile the required Wasm module. The only thing left to do is feed back the interrupted Wasm call with the results that the interrupted host function would have returned and resume the Wasm function call.

iamyulong:
Thank you @Robbepop for your swift reply. 

Per your comment, I'd like to add more context for our case.

A little bit background on us, we're from RadixDLT and we've integrated `wasmi v11.0` into [RadixEngine/Scrypto](https://github.com/radixdlt/radixdlt-scrypto) about one year ago, and we've got this problem when attempting to upgrade to `v23.0`.

[...]

PR #1122, opened 2024-07-09
Allow to compile Wasm modules in host functions called from the Wasmi executor (take 2)

Robbepop:
Closes https://github.com/wasmi-labs/wasmi/issues/631.

This is another relatively naive fix for #631. It heavily refactors the `CodeMap` to no longer use atomics for synchronization but a global `Mutex`. The idea behind this is that this is the lowest we can get with simple synchronization locks. If this technique does not suffice the performance needs, we need to look elsewhere.

Benchmarks comparing this PR with `main`:

Rust-sourced benchmarks do not significantly regress:

image

Call-intense benchmark regress as expected:

image

Other benchmarks are also not signficantly affected:

image

### DIFF
--- a/crates/wasmi/src/engine/code_map.rs
+++ b/crates/wasmi/src/engine/code_map.rs
@@ -563,19 +563,35 @@ impl CodeMap {
         fuel: Option<&mut Fuel>,
         func: CompiledFunc,
     ) -> Result<CompiledFuncRef<'a>, Error> {
-        let mut funcs = self.funcs.lock();
-        let Some(entity) = funcs.get_mut(func) else {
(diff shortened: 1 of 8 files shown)


Robbepop:
Due to the progress in optimizing the `Mutex`-based `CodeMap` implementation provided by this PR and only seeing performance regressions mostly for call-intense workloads (as expected) with mostly single digit percentage numbers, I am confident enough in this implementation to merge it for now with the outlook of finding a better, ideally lock-free and wait-free `CodeMap` data structure as a follow-up PR to hash out the performance regressions of this PR.

PR #1129, opened 2024-07-13
Add uniarg for new feature

lanbones:
UniArg helps extending argument from pop() to pop()/stack()/const(), which is useful in wasm instruction extension.


lanbones:
Wrong PR target, sorry for the mistack.

PR #1141, opened 2024-07-26
Replace `wasmparser-nostd` fork with upstream `wasmparser`

Robbepop:
Thread discussing the performance regressions in `wasmparser`:
https://github.com/bytecodealliance/wasm-tools/issues/1701

---

Unfortunately benchmarks conducted locally show significant performance regressions for Wasm validation:

We can see this clearly in the posted screenshot since `translate/tiny_keccak/checked/eager` regresses massively while `translate/tiny_keccak/unchecked/eager` only regresses by 3% which could be noise.

The only difference between these two cases is that `translate/tiny_keccak/checked/eager` validates the Wasm via `wasmparser` while `translate/tiny_keccak/unchecked/eager` skips Wasm validation and only translates the Wasm via Wasmi.

The huge 48% performance regression for the `checked/lazy` test case indicates that Wasm validation has a much higher constant cost bump for validation of non-function bodies, too. OR it indicates that the setup of the function validators became a lot more expensive.

Reproduce benchmarks locally:

1. On `main` branch: `cargo bench --bench benches -- --save-baseline main`
2. On PR branch: `cargo bench --bench benches -- --baseline main`

image


Robbepop:
I just found one reason why it regressed: I benchmarked without setting the `no-hash-maps` feature. Generally `BTreeMap` is faster than `HashMap` for Wasmi uses and the old `wasmparser-nostd` fork alsways uses `BTreeMap`s. With this feature set local benchmarks still indicate regressions but they are less severe, ranging between 10-25% instead of 20-35%. The unchecked (no validation) case now even has slightly improved performance which indicates very strongly to me that the regressions that we see come from the Wasm validation.

image

Robbepop:
Running yet another benchmark because I forgot to use `no-hash-maps` on `main`. 🤦 

This is the slowdown from `main` to this PR:
image

Robbepop:
After improving translation benchmarks and updating `wasmparser` to `v0.218.0` I reran benchmarks:

Small execution-heavy benchmarks such as `tiny_keccak`:
image

Large benchmarks such as `spidermonkey`:
image

Typical Wasmi users oriented benchmarks such as `erc20`:
image

## Conclusions

- Benchmarks with validation (`checked`) are affected a lot worse.
- Benchmarks that only perform validation on non-function body parts are also heavily affected. (`lazy/checked`)
- Even benchmarks that perform no Wasm validation whatsoever are affected. (`lazy/unchecked`)

Robbepop:
With `wasmparser v0.219.0` there is a new `component-model` crate feature which disables Wasm component-model support. Wasmi can disable this for some nice compile time improvements: 

**v0.218.0:**
```
cargo build -p wasmi --profile bench  22.52s user 1.16s system 201% cpu 11.778 total
```
**v0.219.0:**
```
cargo build -p wasmi --profile bench  18.95s user 1.18s system 178% cpu 11.248 total
```

Robbepop:
I just found out that `wasmparser` v0.219.0 has this API: https://docs.rs/wasmparser/0.219.1/wasmparser/struct.FunctionBody.html#method.as_bytes
This will likely fix a workaround in Wasmi:
- https://github.com/wasmi-labs/wasmi/blob/02621ad7a7f769dc97524075a693cc96e2049cb5/crates/wasmi/src/module/parser/buffered.rs#L159-L170
- https://github.com/wasmi-labs/wasmi/blob/02621ad7a7f769dc97524075a693cc96e2049cb5/crates/wasmi/src/engine/translator/driver.rs#L19-L31

danielstuart14:
[This will allow us to implement ](https://github.com/wasmi-labs/wasmi/issues/1197) requires this change.

Issue #1083, opened 2024-06-24
Add `Store::call_hook` API


Robbepop:
This is about adding a `Store::call_hook` API similar to Wasmtime's [`Store::call_hook`](https://docs.rs/wasmtime/22.0.0/wasmtime/struct.Store.html#method.call_hook) in order to improve Wasmi's Wasmtime API mirroring.}

This should not regress performance of host function calls or the Wasmi executor generally.

emiltayl:
I have a use case for call_hook, so I figured I could take a shot at this. I now have a working version of call_hook, but had one question.

Should call_hook be behind a feature flag? It seems like a change is coming to wasmtime to put it behind a feature flag, which should make it easier to make sure that this has no/minimal performance impact where it is not needed. I have tried benchmarking with my changes locally and could see some regressions, but I also saw 20% speedups that I can't explain, so I don't know how much I can trust the benchmark numbers on my computer.

Robbepop:
Hi @emiltayl ,

thank you for working on this feature.

Can you please share a link to the discussion for putting this behind a feature flag? I would like to know more about the reasoning.
From what I can see with the linked PR the feature has not yet been implemented. Is that correct?
What performance regressions are you already seeing?
I imagined that this feature would not have significant performance regressions for Wasmi, since Wasmi being an interpreter will be slowed down but not as harshly as a JIT runtime. Maybe I am wrong about this.

emiltayl:
The pull request for creating the feature flag is at https://github.com/bytecodealliance/wasmtime/pull/8795, and at https://github.com/bytecodealliance/wasmtime/pull/8808 they disable it by default. It seems to be merged to main, but not put in a release yet if I understand correctly.

I agree that wasmi's performance shouldn't be impacted to a large degree by this change in theory. I don't have the code in front of me right now, but I will be back with the regressions from the benchmarks tomorrow.

emiltayl:
I've included the benchmarks that reported regressions sorted by % change. Judging by the changes I've made, I would have expected to see most change in `execute/call/host/*` as there is an extra function call on host->wasm and wasm->host boundaries, but I did not get any reported regressions from those benchmarks. 

In the instantiate benchmarks an extra `None` is added when creating `Store`, but I don't think that should cause a 32% increase in run time. I have not touch linking or module parsing (I'm pretty sure of this at least).

```
instantiate/reverse_complement
                        time:   [14.146 µs 14.256 µs 14.354 µs]
[...]

PR #1144, opened 2024-07-29
Add `Store::call_hook` API

emiltayl:
My attempt at resolving #1083. I have tried to keep the public API similar to Wasmtime, and everything else as simple as possible.

This adds the function `Store::call_hook` that accepts a `FnMut(&mut T, CallHook) -> Result<(), Error>`, which is then called whenever a wasm function is called from the host, or a host function is called from wasm.

Let me know if you have any questions or you need me to explain or change something. Benchmarking on my own computer indicated a performance regresssion in the execute/br_table and execute/tiny_keccak benchmarks. If these regressions are unacceptable, a possible solution could be to add this behind a feature flag. This should be quite simple, so let me know if you would prefer that.

I did add a `#[allow(clippy::type_complexity)]` to please clippy to avoid having to create a type alias that is not used anywhere else, but it should be quite easy to fix that as well.


Robbepop:
@emiltayl Thanks a lot for this PR! Code looks pretty clean, too. :) Also great to have tests!

I will check it out locally and see how it performs and see if we can improve it. If we cannot fix the performance regressions we might have to put this feature behind a feature gate just like Wasmtime before merging.

Robbepop:
I toyed a bit around and the fastest codegen I received was with this particular implementation:
```rust
/// Executes the callback set by [`Store::call_hook`] if any has been set.
///
/// # Note
///
/// - Returns the value returned by the call hook.
/// - Returns `Ok(())` if no call hook exists.
#[inline]
pub(crate) fn invoke_call_hook(&mut self, call_type: CallHook) -> Result<(), Error> {
    match self.call_hook.as_mut() {
        None => Ok(()),
        Some(call_hook) => Self::invoke_call_hook_impl(&mut self.data, call_type, call_hook),
    }
}

/// Invokes the [`Store::call_hook`] that is asserted to be available in this case.
#[cold]
fn invoke_call_hook_impl(
    data: &mut T,
    call_type: CallHook,
    call_hook: &mut CallHookWrapper<T>,
) -> Result<(), Error> {
    call_hook.0(data, call_type)
}
```

crates/wasmi/src/store.rs
```diff
@@ -165,6 +183,20 @@ fn test_store_is_send_sync() {
     };
 }
 
+#[derive(Debug)]
+/// Argument to the callback set by [`Store::call_hook`] to indicate why the
+/// callback was invoked.
+pub enum CallHook {
```
Robbepop, line 189: Not important but it itches me a bit to see `#[derive(Debug)]` above doc comments. 🙃 Rest of the codebase has doc comments first.

crates/wasmi/src/store.rs
```diff
@@ -977,6 +1011,36 @@ impl<T> Store<T> {
...
+
+    /// Executes the callback set by [`Store::call_hook`] if any has been set
+    /// and returns the value returned by the callback. If no callback has been
+    /// set, `Ok(())` is returned.
+    pub(crate) fn invoke_call_hook(&mut self, call_type: CallHook) -> Result<(), Error> {
+        match &mut self.call_hook {
+            None => Ok(()),
+            Some(call_hook) => call_hook.0(&mut self.data, call_type),
+        }
+    }
```
Robbepop, line 1043: Please see my non-review comment for improvements (optimizations) here. Also could you please see how the suggested code performs on your machine? I only have an M2 macbook to run benchmarks on.

crates/wasmi/tests/e2e/v1/call_hook.rs
```diff
@@ -0,0 +1,237 @@
+//! Tests to check if `Store::call_hook` works as intended.
+
+use wasmi::{core::TrapCode, CallHook, Caller, Error, Extern, Func, Linker, Module, Store};
+
+#[derive(Default)]
+/// Number of times different callback events have fired.
+struct TimesCallbacksFired {
```
Robbepop, line 7: 🙃

emiltayl:
Thanks for the review!

I did not realize that clippy didn't check the tests when I executed it locally. I will fix the clippy errors and reposition the attributes appearing above doc comments. I will also try out your suggested changes and be back with benchmark results.

I do not have the time today, but should get this done before the weekend.

emiltayl:
I tried your suggested changes. My benchmark numbers were quite similar, I even had a slight reported regression in `execute/tiny_keccak` still, but I think I should disable SMT on my 3700x to get reliable benchmark numbers.

I'm not too happy with the fix I did in the tests as `WasmTy` is not implemented for `!`, but it does work and I guess the tests convey what should (not) happen still.

crates/wasmi/src/store.rs
```diff
@@ -978,6 +1011,54 @@ impl<T> Store<T> {
...
+
+    /// Executes the callback set by [`Store::call_hook`] if any has been set.
+    ///
+    /// # Note
+    ///
+    /// - Returns the value returned by the call hook.
+    /// - Returns `Ok(())` if no call hook exists.
+    #[inline]
+    pub(crate) fn invoke_call_hook(&mut self, call_type: CallHook) -> Result<(), Error> {
+        match &mut self.call_hook {
```
Robbepop, line 1043:
```suggestion
        match self.call_hook.as_mut() {
```

Robbepop (approved):
All in all LGTM! Just one super minor nit.
Thanks for the tons of tests.

> I'm not too happy with the fix I did in the tests to work around WasmTy not being implemented for ! (and could therefore not be "returned" from functions), but it does work and I do not think it is too bad.

Can you please point me to the code that you mean? I seem to fail to find it.

I'd be fine with merging your PR.

emiltayl:
Great! Thank you for the help and reviews.

As for the changes I did that I am not that satisfied with myself (I still think the code is good to go, though), it is related to https://github.com/wasmi-labs/wasmi/actions/runs/10148134369/job/28090510038#step:8:84. `should_not_run` in the tests was implemented like this:

```rust
let should_not_run = Func::wrap(&mut store, |_: Caller<TimesCallbacksFired>| {
    panic!("Host function that should not run due to trap from call hook executed");
});
```

From what I understand (which may be flawed/incomplete), with the upcoming changes in Rust 2024, this would fail to compile. This is because the closure's return type would be `!` instead of `()`, and `WasmRet` / `WasmTy` is not implemented for `!`, which is required for the return types of the closures passed to `Func::wrap`. I do not think it neccessarily makes sense to implement either `WasmRet` or `WasmTy` for `!` though, I imagine it would only be useful for tests like these.

The new `should_not_run` can be seen at https://github.com/wasmi-labs/wasmi/pull/1144/files#diff-93c6a63dd2cba7296e30181521de10f97f315df932f3beb82c90d8879dac56deR175. This requires an additional `assert!` to check whether `should_not_run` has been executed or not, but I think it is a decent solution even though I think the old `should_not_run` is a tad more neat.

Robbepop:
@emiltayl Couldn't the function return an error instead of panicking to provide the same functionality of not continuing the execution?

PR #1154, opened 2024-08-27
Move `TypedVal` to `wasmi_core` crate

Robbepop:


Robbepop:
The `Miri (spec)` job fails because of a new regression in the `wasmi_c_api` crate that has nothing to do with this PR.

PR #1155, opened 2024-08-27
Fix weird `clippy` warning

Robbepop:


Robbepop:
The failing `Miri (spec)` job has nothing to do with this PR and it simply time outs. I checked in other CI pipelines and the timeout was very close for them so it is just unlucky that it timed out here. We probably need to increase the time out limit for this particular job.

PR #1163, opened 2024-09-04
Use named fields for Wasmi `load` instructions

Robbepop:
This will help simplify the integration of https://github.com/wasmi-labs/wasmi/pull/1152.


Robbepop:
The `Miri (spec)` CI job fails because it timed out. This is a known issue and we need to either change what the job tests via Miri or increase the timeout limit to fix the flakiness.

PR #1173, opened 2024-09-13
Clean up and refactor `RegSpan` type and usage

Robbepop:


Robbepop:
The `Mini (spec)` job found UB in some `unsafe` code during this run. However, this PR does not change this particular part of the code base and has no influence on it. Also I am not yet sure if this is a false-positive of `miri` (nightly) since this has been tested for a long time and Miri has never reported issues with it. Maybe a change to semantics with new Rust compiler versions caused this.

PR #1174, opened 2024-09-13
Fix `br_table_many` execution bug

Robbepop:
Fix for bug introduced in https://github.com/wasmi-labs/wasmi/pull/1158 post latest v0.36.0 release.


Robbepop:
Concerning the Miri jobs, read here: https://github.com/wasmi-labs/wasmi/pull/1173

PR #1186, opened 2024-09-18
Remove old `HostVisitor` impl code remnants

Robbepop:
oups 🙃


Robbepop:
The `udeps` job is failing because this PR does not yet include the most recent changes on `main` which fixes this particular issue.

PR #1210, opened 2024-09-30
Prepare release for Wasmi `v0.37.0`

Robbepop:


Robbepop:
The failing `miri` job is because of the usual timeout. We should fix this CI job some time soon.

PR #1233, opened 2024-10-08
WIP: add Wasmi translation invariants checker

Robbepop:
Closes https://github.com/wasmi-labs/wasmi/issues/1209.


Robbepop:
The idea of a separate invariants checker pass did not really shape to what was hoped for.
Instead of a separate pass we should invest more into local `debug_assert`s and an more involved `extra-checks` feature. Keeping a separate invariants pass in sync with everything else is a mundane and error prone work.
This branch is very out of date and won't be merged, thus closed.

PR #1234, opened 2024-10-09
Use Honggfuzz for fuzzing targets

gtrepta:
This switches over the fuzz targets from using libfuzzer to using honggfuzz, which has a couple of nice features:

Honggfuzz detects how many cores are available on the host and launches as many processes as it can to maximize execution rate, no manual intervention is necessary here.

It also has support for hardware based feedback, like instruction/branch count, and on Intel hardware it can do edge coverage. This removes the need for compiling with instrumentation which creates a large speedup. It isn't enabled in this PR, because it looks unlikely that the github runners enable access to the perf counters for this feature, but it's available should these targets ever be run on hardware that supports it.

https://honggfuzz.dev/

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -221,7 +221,7 @@ jobs:
       - name: Install cargo-honggfuzz and dependencies
         run: |
           cargo install honggfuzz || true
-          apt-get update && apt-get install --yes binutils-dev libunwind-dev
+          sudo apt-get update && sudo apt-get install --yes binutils-dev libunwind-dev
       - name: Run HFuzz
         env:
           HFUZZ_RUN_ARGS: "--run_time 120 --verbose --quiet" # 2 minutes of fuzzing
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -198,9 +198,12 @@ jobs:
       - name: Check uDeps
         run: cargo udeps --all-targets
 
-  fuzz-translate:
-    name: Fuzz (Translation)
+  fuzz:
+    name: Fuzz
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        fuzz_target: ['translate', 'translate_metered', 'execute', 'differential']
     steps:
       - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332
         with:
@@ -213,71 +216,16 @@ jobs:
         uses: Swatinem/rust-cache@v2
         with:
           cache-directories: |
-            ~/fuzz/corpus/translate/
-            ~/fuzz/corpus/translate_metered/
-      - name: Install cargo-fuzz
+            ${{ github.workspace }}/hfuzz_target
+            ${{ github.workspace }}/hfuzz_workspace
+      - name: Install cargo-honggfuzz and dependencies
         run: |
-          # Note: We use `|| true` because cargo install returns an error
-          #       if cargo-udeps was already installed on the CI runner.
-          cargo install cargo-fuzz || true
-      - name: Build Fuzzing
-        run: cargo fuzz build translate
-      - name: Fuzz (Translation)
-        run: cargo fuzz run translate -j 2 --verbose -- -max_total_time=60 # 1 minute of fuzzing
-      - name: Fuzz (Translation) + fuel
-        run: cargo fuzz run translate_metered -j 2 --verbose -- -max_total_time=60 # 1 minute of fuzzing
-
-  fuzz-execute:
-    name: Fuzz (Execution)
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332
-        with:
-          submodules: true
-      - uses: dtolnay/rust-toolchain@master
-        with:
-          # Pin nightly so that it does not invalidate GitHub Actions cache too frequently.
-          toolchain: nightly-2024-09-24
-      - name: Set up Rust cache
-        uses: Swatinem/rust-cache@v2
-        with:
-          cache-directories: |
-            ~/fuzz/corpus/execute/
-      - name: Install cargo-fuzz
-        run: |
-          # Note: We use `|| true` because cargo install returns an error
-          #       if cargo-udeps was already installed on the CI runner.
-          cargo install cargo-fuzz || true
-      - name: Build Fuzzing
-        run: cargo fuzz build execute
-      - name: Fuzz (Execution)
-        run: cargo fuzz run execute -j 2 --verbose -- -max_total_time=120 # 2 minutes of fuzzing
-
-  fuzz-differential:
-    name: Fuzz (Differential)
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332
-        with:
-          submodules: true
-      - uses: dtolnay/rust-toolchain@master
-        with:
-          # Pin nightly so that it does not invalidate GitHub Actions cache too frequently.
-          toolchain: nightly-2024-09-24
-      - name: Set up Rust cache
-        uses: Swatinem/rust-cache@v2
-        with:
-          cache-directories: |
-            ~/fuzz/corpus/differential/
-      - name: Install cargo-fuzz
-        run: |
-          # Note: We use `|| true` because cargo install returns an error
-          #       if cargo-udeps was already installed on the CI runner.
-          cargo install cargo-fuzz || true
-      - name: Build Fuzzing
-        run: cargo fuzz build differential
-      - name: Fuzz (Differential)
-        run: cargo fuzz run differential -j 2 --verbose -- -max_total_time=120 # 2 minutes of fuzzing
+          cargo install honggfuzz || true
+          apt-get update && apt-get install --yes binutils-dev libunwind-dev
+      - name: Run HFuzz
+        env:
+          HFUZZ_RUN_ARGS: "--run_time 120 --verbose --quiet" # 2 minutes of fuzzing
+        run: cargo hfuzz run ${{ matrix.fuzz_target }}
 
   miri:
     name: Miri
--- a/fuzz/fuzz_targets/differential.rs
+++ b/fuzz/fuzz_targets/differential.rs
@@ -3,7 +3,7 @@ mod utils;
 use arbitrary::Unstructured;
 use honggfuzz::fuzz;
 use std::{collections::hash_map::RandomState, mem};
-use utils::{arbitrary_config, ty_to_val};
+use utils::arbitrary_config;
 use wasmi as wasmi_reg;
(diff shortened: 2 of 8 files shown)


fuzz/fuzz_targets/execute.rs
```diff
@@ -1,57 +1,68 @@
...
-    let module = Module::new(store.engine(), wasm.as_slice()).unwrap();
-    let Ok(preinstance) = linker.instantiate(&mut store, &module) else {
-        return;
-    };
-    let Ok(instance) = preinstance.ensure_no_start(&mut store) else {
-        return;
-    };
+            let mut config = Config::default();
+            config.consume_fuel(true);
+            config.compilation_mode(wasmi::CompilationMode::Lazy);
```
Robbepop, line 19: Why was lazy compilation mode chosen? It feels like this will worsen the fuzz coverage a lot since code is only going to be compiled when it is actually used but don't we ideally want to always compile and run everything?

.github/workflows/rust.yml
```diff
@@ -198,64 +198,12 @@ jobs:
...
-          #       if cargo-udeps was already installed on the CI runner.
-          cargo install cargo-fuzz || true
-      - name: Build Fuzzing
-        run: cargo fuzz build execute
-      - name: Fuzz (Execution)
-        run: cargo fuzz run execute -j 2 --verbose -- -max_total_time=120 # 2 minutes of fuzzing
-
-  fuzz-differential:
-    name: Fuzz (Differential)
+  fuzz:
```
Robbepop, line 201: Do we have just one `fuzz` job because fuzzing now runs everything multi-threaded? I split up the fuzzing jobs in the past because that allowed us to use more CI core at the same time since each CI machine comes with 2 cores. So 3 jobs -> 6 CPU cores for fuzzing. With this it seems we are back to just 2 cores. Is that correct? If yes, could we fix this in an elegant way? Obviously using 3 jobs was kind of a "hack".

fuzz/fuzz_targets/differential.rs
```diff
@@ -613,31 +620,43 @@ impl FuzzContext {
...
-    context.run();
-});
+fn main() {
+    loop {
+        fuzz!(|seed: &[u8]| {
+            let mut unstructured = Unstructured::new(seed);
+            let Ok(mut smith_module) =
+                arbitrary_config(&mut unstructured).and_then(|mut config| {
+                    config.reference_types_enabled = false;
+                    config.tail_call_enabled = false;
```
Robbepop, line 630: I wonder why we do not use `tail-calls` proposal in differential fuzzing. 🤔 Do you have an idea?

fuzz/fuzz_targets/differential.rs
```diff
@@ -613,31 +620,43 @@ impl FuzzContext {
...
-        exports,
-    };
-    context.run();
-});
+fn main() {
+    loop {
+        fuzz!(|seed: &[u8]| {
+            let mut unstructured = Unstructured::new(seed);
+            let Ok(mut smith_module) =
+                arbitrary_config(&mut unstructured).and_then(|mut config| {
```
Robbepop, line 628: Could this be dis-entangled by removing the `and_then` and simply pulling out the `let else` here?

fuzz/fuzz_targets/execute.rs
```diff
@@ -1,57 +1,68 @@
...
-    store.limiter(|lim| lim);
-    let module = Module::new(store.engine(), wasm.as_slice()).unwrap();
-    let Ok(preinstance) = linker.instantiate(&mut store, &module) else {
-        return;
-    };
-    let Ok(instance) = preinstance.ensure_no_start(&mut store) else {
-        return;
-    };
+            let mut config = Config::default();
+            config.consume_fuel(true);
```
Robbepop, line 18: With the above `smith_module.ensure_termination(1000 /* fuel */)` we now have a double fuel-based protection. Not sure if worth it. 🤔

fuzz/fuzz_targets/utils.rs
```diff
@@ -1,71 +1,48 @@
...
-        gc_enabled: false,
-        tail_call_enabled: false,
-        ..Default::default()
-    }
+use wasmi::{core::ValType, ExternRef, FuncRef, Val};
+
+pub fn disable_unsupported_config(config: &mut wasm_smith::Config) {
+    config.gc_enabled = false;
+    config.exceptions_enabled = false;
+    config.max_memories = 1;
```
Robbepop, line 9: Wasmi supports `multi-memory` since v0.37.0. :) Can be enabled.

fuzz/fuzz_targets/utils.rs
```diff
@@ -1,71 +1,48 @@
...
-
-    config.map(|mut config| {
-        config.gc_enabled = false;
-        config.exceptions_enabled = false;
-        config.simd_enabled = false;
-        config.threads_enabled = false;
-
-        wasm_smith::Module::new(config, &mut unstructured)
-    })?
+pub fn arbitrary_swarm_config_module(
```
Robbepop, line 22: Sorry for the confusion, but what exactly is a swarm config module?

fuzz/fuzz_targets/utils.rs
```diff
@@ -1,71 +1,48 @@
...
+                .into(),
+        ),
+        ValType::F64 => Val::F64(
+            u.int_in_range::<u64>(u64::MIN..=u64::MAX)
+                .map(f64::from_bits)
+                .unwrap_or(1.0)
+                .into(),
         ),
+        ValType::FuncRef => Val::FuncRef(FuncRef::null()),
+        ValType::ExternRef => Val::ExternRef(ExternRef::null()),
```
Robbepop, line 46: Generally this is way nicer than before with the constant magic value. Though I am a bit astounded that there are no convenience functions to create simple Rust primitives (e.g. `i32`) from `&mut Unstructured`.

Hmmm, I think this is it: https://docs.rs/arbitrary/1.3.2/arbitrary/trait.Arbitrary.html#impl-Arbitrary%3C'a%3E-for-i32
`Arbitrary` impl for Rust primitives should do the job, right?

Robbepop (changes requested):

Robbepop:
Thanks a lot for the improvements to Wasmi's fuzzing system: One thing I wondered was the strength and weaknesses of different fuzzing engines. For this I found this paper which concludes that libfuzzer is better at finding memory issues while hongfuzz is better at coverage. Further it continues that it is probably best to combine different fuzzers to use their different strengths. What is your opinion on this?

fuzz/fuzz_targets/differential.rs
gtrepta: A TranslationError gets thrown by WasmiStack when trying to make the module

https://github.com/wasmi-labs/wasmi/blob/708bef07206c8805db92e7f4513cb5b2c729d51e/fuzz/fuzz_targets/differential.rs#L216

``called `Result::unwrap()` on an `Err` value: Module(Translation(TranslationError { inner: Validate(BinaryReaderError { inner: BinaryReaderErrorInner { message: "tail calls support is not enabled", offset: 157, needed_hint: None } })
 }))``

.github/workflows/rust.yml
gtrepta: This is still 4 jobs for the 4 fuzzing targets. A job runs for each variable in `strategy.matrix`.

https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategymatrix

fuzz/fuzz_targets/utils.rs
gtrepta: The WasmiStack DifferentialTarget also has issues creating a module when this is enabled.

fuzz/fuzz_targets/differential.rs
Robbepop: This is probably because `tail-calls` (and `extended-const`) are disabled by default in Wasmi v0.31 (`wasmi_stack`).

fuzz/fuzz_targets/utils.rs
Robbepop: true.
Wasmtime supports it though.
Ideally our fuzzing strategy was a bit more flexible and depended on `Unstructured` input what other Wasm runtime to compare with `wasmi_reg`.

fuzz/fuzz_targets/utils.rs
gtrepta: Swarm testing is a term we've lifted off of this paper

https://users.cs.utah.edu/~regehr/papers/swarm12.pdf

Using random wasm_smith configurations helps with coverage.

fuzz/fuzz_targets/execute.rs
Scott-Guest: Ah apologies, I agree for coverage it's be better to do the eager compilation - I'll change this.

I was using lazy when I was fuzzing for execution time and only wanted to consider the actual opcodes that got executed before a set fuel limit ran out.

fuzz/fuzz_targets/execute.rs
Scott-Guest: Good catch - just a copy-and-paste error from moving changes to this branch.

gtrepta:
> Further it continues that it is probably best to combine different fuzzers to use their different strengths. What is your opinion on this?

It would require extra effort maintaining different compilation targets and runner scripts/jobs. It might be a good idea, but there's plenty of scaling that can be done with a single target first.

fuzz/fuzz_targets/differential.rs
gtrepta: The `let else` is here to silently fail in case wasm_smith encounters any errors which we don't want to show up during fuzzing. Not sure what you mean about removing `and_then`, splitting this statement into two `let`s?

fuzz/fuzz_targets/execute.rs
Scott-Guest: Actually, on second glance, not sure which part you're referring to. In this harness, the `smith_module.ensure_termination` call has been removed and replaced with the built-in fuel metering.

Robbepop: You are right, for the `execution` fuzz job that should be fine and maybe even better because the fuzzer covers more code this way. For differential fuzzing it would be bad.

fuzz/fuzz_targets/differential.rs
Robbepop: I meant something like this:
```rust
let Ok(config) = arbitrary_config(&mut unstructured) else {
    return;
}
config.reference_types_enabled = false;
config.tail_call_enabled = false;
let Ok(mut smith_module) = wasm_smith::Module::new(config, &mut unstructured) else {
    return;
};
```
However, this is a bit nit-picky tbh.

Robbepop (approved):
I think this is ready. The only review items left are nit-picky. Thanks a lot for the great improvements. 🚀
More improvements can be done in follow-up PRs if needed.

PR #1243, opened 2024-10-21
Reduce `cmp` and `cmp+br` instructions

Robbepop:
This reduces `i{32,64}.{gt,ge}_{s,u}` and `f{32,64}.{gt,ge}` instructions into respective `{lt,le}` variants.
Fused `cmp+br` instructions are also affected by the above change.

To support this, new `{lt,le}` variants with 16-bit immediate `lhs` values have been added.

Overall this reduces the total amount of Wasmi instructions by 24 which is quite significant.

Local benchmarks show significant improvements across the board:
image


dkcumming (approved):
I had a bit of trouble with the executor, as I am less experienced at that part, but it I couldn't see anything that looked wrong. The rest of the code `ir`, `executor`, `instrs`, `tests` I was more thorough and understood better and all looks good to me!

Robbepop:
> I had a bit of trouble with the executor, as I am less experienced at that part, but it I couldn't see anything that looked wrong. The rest of the code `ir`, `executor`, `instrs`, `tests` I was more thorough and understood better and all looks good to me!

Thank you for the code review @dkcumming ! 🚀

PR #1278, opened 2024-11-02
Simplify error handling and restructure `runner.rs`

Robbepop:


Robbepop:
The Miri (spec) job experienced a time-out despite the recent increase in time.

PR #1287, opened 2024-11-04
Add missing import

siefkenj:
The current example provided in the documentation fails to build because it uses the `wat` without a reference. This PR adds the necessary use statement.

### DIFF
--- a/crates/wasmi/src/lib.rs
+++ b/crates/wasmi/src/lib.rs
@@ -16,6 +16,7 @@
 //! ```
 //! use anyhow::Result;
 //! use wasmi::*;
+//! use wat;
 //!
 //! fn main() -> Result<()> {
 //!     // First step is to create the Wasm execution engine with some config.


Robbepop:
@siefkenj thank you for the bug report and fix. How did you discover this bug? Unfortunately the CI did not catch it. When using `cargo test --doc` locally it somehow also passes the test for me. 🤔

Robbepop:
@siefkenj I am going to merge this out of caution but would be very happy about a response later.

siefkenj:
> @siefkenj thank you for the bug report and fix. How did you discover this bug?

I did `cargo add wasmi` to another project, then I copy-and-pasted the example into it, and it failed to build.

Robbepop:
> > @siefkenj thank you for the bug report and fix. How did you discover this bug?
> 
> I did `cargo add wasmi` to another project, then I copy-and-pasted the example into it, and it failed to build.

Ah okay, that is to be expected then.

PR #1298, opened 2024-11-07
C API: Fix function name of compilation mode setter

turbolent:
`config.h` defines `WASMI_CONFIG_PROP` to define setters for configuration options as `wasmi_config_##name##_set`. Just like for all other configuration options, this macro is used for `compilation_mode`.  However, the function is exported from the `c_api` crate with a different name.

Fix the exported name and make it consistent with the other configuration options.

### DIFF
--- a/crates/c_api/src/config.rs
+++ b/crates/c_api/src/config.rs
@@ -141,7 +141,7 @@ pub enum wasmi_compilation_mode_t {
 ///
 /// Wraps [`wasmi::Config::compilation_mode`]
 #[no_mangle]
-pub extern "C" fn wasmi_config_set_compilation_mode(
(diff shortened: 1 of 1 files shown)


turbolent:
It might make sense to add some tests for the C API to ensure the headers and the library can be used successfully.

Robbepop:
> It might make sense to add some tests for the C API to ensure the headers and the library can be used successfully.

Yes, totally! I wanted to do this a long time ago but there are so many things with a higher priority right now. Missing automated tests is why I flagged the C-API as experimental so far. Though, I am very happy about help and PRs. :)

Robbepop:
@turbolent Thanks a lot for the fix!

PR #1312, opened 2024-11-21
feat: Add the `mangle-symbols` feature

xdoardo:
As per title.


Robbepop:
@xdoardo Thank you!

I made a small mistake when I told you that it would be best to enable this feature by default. I think it is better to disable it by default.

Also could you add some documentation to the `mangle-symbols` crate feature so that users know why it exists and what it is useful for?

Robbepop (approved):
Thanks a lot for this work! LGTM

PR #1314, opened 2024-11-21
Use the config in the wasmtime oracle

gtrepta:
The config in the wasmtime oracle for differential fuzzing wasn't being used, this fixes that.

It also disables POSIX signals in the config because some fuzzers (ie. honggfuzz) will catch them and report false positives. A couple examples of these signals are SIGILL for bad instructions, and SIGSEGV when there's an out of bounds load in the webassembly. [More info.](https://docs.rs/wasmtime/latest/wasmtime/struct.Config.html#method.signals_based_traps)


Robbepop (approved):

Robbepop:
@gtrepta great catch! thank you.

PR #1315, opened 2024-11-25
Add `prefix-symbols` feature

xdoardo:
Hello again! 

This small patch adds a `prefix-symbols` feature that adds a `wasmi_` prefix to all the public symbols that did not have it before, keeping the symbols not mangled. This way, users (that need to use `wasmi`'s c_api implementation with other backends) need not worry about `Option<Box<_>>`es but can simply use `bindgen` to take care of it. 

I think that this method makes more sense than the other one we ended up merging.

Again, thank you in advance for taking the time to review the PR.

### DIFF
--- a/crates/c_api/macro/lib.rs
+++ b/crates/c_api/macro/lib.rs
@@ -22,7 +22,7 @@ pub fn declare_own(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
     let ty = extract_ident(input);
     let name = ty.to_string();
     let delete = quote::format_ident!("{}_delete", &name[..name.len() - 2]);
-    let prefixed_delete = format!("wasmi_{}_delete", &name[..name.len() - 2]);
+    let prefixed_delete = format!("wasmi_{delete}");
     let docs = format!("Deletes the [`{name}`].");
 
     (quote! {
@@ -40,7 +40,7 @@ pub fn declare_ty(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
     let name = ty.to_string();
     let prefix = &name[..name.len() - 2];
     let copy = quote::format_ident!("{}_copy", &prefix);
-    let prefixed_copy = format!("wasmi_{}_copy", &prefix);
+    let prefixed_copy = format!("wasmi_{copy}");
     let docs = format!(
         "Creates a new [`{name}`] which matches the provided one.\n\n\
         The caller is responsible for deleting the returned value via [`{prefix}_delete`].\n\n\
@@ -66,37 +66,37 @@ pub fn declare_ref(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
     let name = ty.to_string();
     let prefix = &name[..name.len() - 2];
     let same = quote::format_ident!("{}_same", prefix);
-    let same_prefixed = format!("wasmi_{}_same", prefix);
+    let same_prefixed = format!("wasmi_{same}");
     let same_docs = format!(
         "Returns `true` if the given references are pointing to the same [`{name}`].\n\n\
         This is not yet supported and aborts the process upon use."
     );
     let get_host_info = quote::format_ident!("{}_get_host_info", prefix);
-    let get_host_info_prefixed = format!("wasmi_{}_get_host_info", prefix);
+    let get_host_info_prefixed = format!("wasmi_{get_host_info}");
     let get_host_info_docs = format!(
         "Returns the host information of the [`{name}`].\n\n\
         This is not yet supported and always returns `NULL`."
     );
     let set_host_info = quote::format_ident!("{}_set_host_info", prefix);
-    let set_host_info_prefixed = format!("wasmi_{}_set_host_info", prefix);
+    let set_host_info_prefixed = format!("wasmi_{set_host_info}");
     let set_host_info_docs = format!(
         "Sets the host information of the [`{name}`].\n\n\
         This is not yet supported and aborts the process upon use."
     );
     let set_host_info_final = quote::format_ident!("{}_set_host_info_with_finalizer", prefix);
-    let set_host_info_final_prefixed = format!("wasmi_{}_set_host_info_with_finalizer", prefix);
+    let set_host_info_final_prefixed = format!("wasmi_{set_host_info_final}");
     let set_host_info_final_docs = format!(
         "Sets the host information finalizer of the [`{name}`].\n\n\
         This is not yet supported and aborts the process upon use."
     );
     let as_ref = quote::format_ident!("{}_as_ref", prefix);
-    let as_ref_prefixed = format!("wasmi_{}_as_ref", prefix);
+    let as_ref_prefixed = format!("wasmi_{as_ref}");
     let as_ref_docs = format!(
         "Returns the [`{name}`] as mutable reference.\n\n\
         This is not yet supported and aborts the process upon use."
     );
     let as_ref_const = quote::format_ident!("{}_as_ref_const", prefix);
-    let as_ref_const_prefixed = format!("wasmi_{}_as_ref_const", prefix);
+    let as_ref_const_prefixed = format!("wasmi_{as_ref_const}");
     let as_ref_const_docs = format!(
         "Returns the [`{name}`] as immutable reference.\n\n\
         This is not yet supported and aborts the process upon use."
--- a/crates/c_api/macro/lib.rs
+++ b/crates/c_api/macro/lib.rs
@@ -22,13 +22,12 @@ pub fn declare_own(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
     let ty = extract_ident(input);
     let name = ty.to_string();
     let delete = quote::format_ident!("{}_delete", &name[..name.len() - 2]);
-    let prefixed_delete = format!("wasmi_{delete}");
     let docs = format!("Deletes the [`{name}`].");
 
     (quote! {
         #[doc = #docs]
         #[no_mangle]
-        #[cfg_attr(feature = "prefix-symbols", export_name = #prefixed_delete)]
+        #[cfg_attr(feature = "prefix-symbols", wasmi_c_api_macros::prefix_symbol)]
         pub extern "C" fn #delete(_: ::alloc::boxed::Box<#ty>) {}
     })
     .into()
@@ -40,7 +39,6 @@ pub fn declare_ty(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
     let name = ty.to_string();
     let prefix = &name[..name.len() - 2];
     let copy = quote::format_ident!("{}_copy", &prefix);
-    let prefixed_copy = format!("wasmi_{copy}");
     let docs = format!(
         "Creates a new [`{name}`] which matches the provided one.\n\n\
         The caller is responsible for deleting the returned value via [`{prefix}_delete`].\n\n\
@@ -52,7 +50,7 @@ pub fn declare_ty(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
 
         #[doc = #docs]
         #[no_mangle]
-        #[cfg_attr(feature = "prefix-symbols", export_name = #prefixed_copy)]
+        #[cfg_attr(feature = "prefix-symbols", wasmi_c_api_macros::prefix_symbol)]
         pub extern "C" fn #copy(src: &#ty) -> ::alloc::boxed::Box<#ty> {
             ::alloc::boxed::Box::new(src.clone())
         }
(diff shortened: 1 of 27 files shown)


Robbepop:
@xdoardo Thanks for the PR!

Unforunately it seems that both new features for the C-API are conflicting each other which itself conflicts with Rust API design guidelines that crate features must be additives.

How could we best resolve this? 🤔

Does this mean that symbols like `wasm_store_delete` will now be called `wasmi_wasm_store_delete`?

xdoardo:
Hello again @Robbepop , thanks for the response. Let me give a bit of context so to explain better why and how I came to this PR. 

In [this](https://github.com/wasmerio/wasmer/pull/5243) PR I add support for multiple backends (or runtimes) to Wasmer: among others, we have support for [V8](https://v8.dev/), [WAMR](https://github.com/bytecodealliance/wasm-micro-runtime) and wasmi. All of them are embedded using their implementation of the `wasm_c_api` standard. 

Of course, using the respective pristine builds, the three libraries mentioned above all export the `wasm_...` symbols, resulting in a duplicate symbol errors when trying to link them all together in a single `wasmer` build. For `v8` and `wamr` I was able to use `objcopy` to [rename](https://github.com/wasmerio/wasmer/blob/5198-cli-support-wasm_c_api-backends-together-with-native-ones/lib/api/build.rs#L350) these `wasm_...` symbols giving them a prefix.  The reason for this PR is, simply, that the same mechanism - which, I'll admit, is also pretty hacky - I think does not play well when the library being embedded in the binary is a Rust library (i.e. does not play too well with cargo: it's not as easy as downloading the library from an URL and running `objcopy` on it). 

The other nice side of this endeavour is that with `bindgen` I can add a `#[link_name = "wasmi_wasm_..."]`  attribute to the symbols imported from the header. In practice, the result is that the code generated from `bindgen` is: 
```
#[link_name = "\u{1}_wasmi_wasm_byte_vec_new_empty"]
pub fn wasm_byte_vec_new_empty(out: *mut wasm_byte_vec_t);
```
and from the library side I can simply import `wasm_byte_vec_new_empty`; in turn, this allows the library code calling the backends pretty much identical for all `wasm_c_api` implementers.
 
To conclude: https://github.com/wasmi-labs/wasmi/pull/1312 had pretty much the same meaning, at least for my use case. However, without the shims created by `bindgen`, the result of using a library that was specifically thought to be used by C-like code (and not Rust without `bindgen`, as in my case) is...not good, at least in my experience: long story short, the result  is a vast amount of null checks for raw pointers to be able to cast them back to "normal" rust references (again, at least for my use case).

The next question, then, might be: why not use `bindgen` to generate the shims and then use `#[link_name = ...]` to link to mangled `wasmi_c_impl` symbols? The reason is that to do that I need to have a precise mapping I can compute at build time, that is, I'd need to have a way to get the mangled name of a Rust function in a `build.rs` file, which is probably not unfeasible but not ideal.

-- 

This comment has quite a bit of technical details which I tried to report briefly but in a meaningful manner. Of course, if something is not clear or, even better, if I missed some obvious and easier way to do what I'm trying to do without jumping through all these hoops, I'm happy to provide more context.

xdoardo:
> Both features conflict with each other and if this feature is superior then we only want this solution. 

I agree. I didn't remove the previous feature in this PR because I wanted to come to a consensus before doing it: my previous PR was, in hindsight, not a good solution to the problem at hand. Sorry for that. 

> Though, I am very uncertain if I really want to change the C-API headers which are a given by the Wasm spec standard. This probably is a really bad idea.

I agree. My thought here is that the asymmetry between the symbols exported in the library and the symbols declared in the C-API header can make a lot of sense or no sense at all depending on the use case. In my case it does, as I can simply copy the code for another C-API implementer, change `include!(".../wamr_bindings.rs");` to `include!(".../wasmi_bindings.rs");`  and have a working wasmi-backed backend, given that I add the `link_name` attributes while generating shims with `bindgen`.  On the other hand, I also understand that other users might expect the symbols exported in the C-API header to be the same symbols exported from the library, without any asymmetry. 

The choice, here, is between changing a standard API or expect users to know what they're doing when enabling this feature (the third option being removing both symbols-related features).  We should also take into account that the current `mangle-symbols` feature already induces an asymmetry between the symbols in the header and those in the library, with the difference that the symbol exposed in the library are pretty much unknown to users (again, something like `__ZN96_$LT$wasm_store_delete$GT$17hdfa21c4755fac718E`). I think that the second option, that is giving users a bit of responsibility with this feature, is the best as it allows new use-cases and does not change the standard API.

Robbepop:
> We should also take into account that the current mangle-symbols feature already induces an asymmetry between the symbols in the header and those in the library [..]

Good point. I think it would be fine if we revert your other PR and add docs that describe how to use this new `prefix` feature without changing our C headers. This way users will understand why it exists and how to make use of the same solution that you used to make use of it.

Robbepop:
> Uhm, I'm not sure why `Build` in CI is failing. I don't see any difference between that step and the `build (all features)` steps in the various `Test (*-latest)` jobs. Any clue?

You can view the logs here:

- `miri` failure: https://github.com/wasmi-labs/wasmi/actions/runs/12036488681/job/33557871738?pr=1315#step:8:30
- `build` failure: https://github.com/wasmi-labs/wasmi/actions/runs/12036488681/job/33557866591?pr=1315#step:6:180

The `build` failure is weird, I don't know why it fails.

Would you be so kind and separate out the revert of the other PR into another PR? This would make this PR simpler to review.

crates/c_api/macro/lib.rs
```diff
@@ -62,31 +66,37 @@ pub fn declare_ref(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
...
         This is not yet supported and always returns `NULL`."
     );
     let set_host_info = quote::format_ident!("{}_set_host_info", prefix);
+    let set_host_info_prefixed = format!("wasmi_{}_set_host_info", prefix);
     let set_host_info_docs = format!(
         "Sets the host information of the [`{name}`].\n\n\
         This is not yet supported and aborts the process upon use."
     );
     let set_host_info_final = quote::format_ident!("{}_set_host_info_with_finalizer", prefix);
+    let set_host_info_final_prefixed = format!("wasmi_{}_set_host_info_with_finalizer", prefix);
```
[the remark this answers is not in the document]
xdoardo: fe7117e

crates/c_api/src/engine.rs
```diff
@@ -19,6 +19,7 @@ wasmi_c_api_macros::declare_own!(wasm_engine_t);
 ///
 /// Wraps [`wasmi::Engine::default`].
 #[no_mangle]
+#[cfg_attr(feature = "prefix-symbols", export_name = "wasmi_wasm_engine_new")]
```
Robbepop, line 22: I am currently thinking about a solution to prevent having to duplicate the names of all those numerous functions for this `cfg_attr`. Technically we could craft a proc. macro attribute that can do this automatically, however, the big downside is that this proc. macro would require a dependency on `syn` which is huge. 🤔

crates/c_api/macro/lib.rs
```diff
@@ -150,3 +158,39 @@ pub fn declare_ref(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
...
+    let prefixed_fn_name = format!("wasmi_{}", fn_name.unwrap());
+
+    let mut attr: proc_macro::TokenStream = quote! {
+        #[cfg_attr(feature = "prefix-symbols", export_name = #prefixed_fn_name)]
+    }
+    .into();
+    attr.extend(input);
+
+    attr
+}
```
Robbepop, line 196: Thank you! I think this is fine for merging even though it does not really check for function signatures. What I had in my mind is to do parsing similar to `syn` but overly specific to our use case as to not have to use `syn` as dependency. Shouldn't be too much work but I am willing to implement this as a follow-up PR so we can get this merged.

Robbepop (approved):
Please fix the `clippy` CI job here: https://github.com/wasmi-labs/wasmi/actions/runs/12049971351/job/33597861395?pr=1315#step:5:18

Then we should be fine with merging this.
Thanks a lot!

crates/c_api/macro/lib.rs
```diff
@@ -27,6 +27,7 @@ pub fn declare_own(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
     (quote! {
         #[doc = #docs]
         #[no_mangle]
+        #[wasmi_c_api_macros::prefix_symbol]
```
Robbepop, line 30: Please use `#[cfg_attr(feature = "prefix-symbols", prefix_symbol)]` in all usage sites as I recommended earlier. The reason is that this way we avoid running the proc. macro numerous times which is exceptionally expensive when not needed.

Robbepop (changes requested):

crates/c_api/macro/lib.rs
Robbepop: With this `cfg_attr` guard the `prefix_symbol` proc. macro should expand to a non-`cfg_attr` but a normal `cfg` attribute.

xdoardo: You're totally right.

xdoardo:
> Could you please remove this massive `o` file?

Yep, sorry, my bad!

crates/c_api/src/extern.rs
```diff
@@ -57,6 +60,10 @@ pub extern "C" fn wasm_extern_as_func(e: &mut wasm_extern_t) -> Option<&mut wasm
 ///
 /// Returns `None` if `e` is not a [`wasm_func_t`].
 #[no_mangle]
+#[cfg_attr(
+    feature = "prefix-symbols",
+    export_name = "wasmi_wasm_extern_as_func_const"
+)]
```
Robbepop, line 66: This should use the `#[cfg_attr(feature = "prefix-symbols", wasmi_c_api_macros::prefix_symbol)]` attribute. Probably an oversight due to text replacement without multi-lines.

crates/c_api/src/extern.rs
```diff
@@ -89,6 +99,10 @@ pub extern "C" fn wasm_extern_as_table(e: &mut wasm_extern_t) -> Option<&mut was
 ///
 /// Returns `None` if `e` is not a [`wasm_table_t`].
 #[no_mangle]
+#[cfg_attr(
+    feature = "prefix-symbols",
+    export_name = "wasmi_wasm_extern_as_table_const"
+)]
```
Robbepop, line 105: This should use the `#[cfg_attr(feature = "prefix-symbols", wasmi_c_api_macros::prefix_symbol)]` attribute. Probably an oversight due to text replacement without multi-lines.

crates/c_api/src/frame.rs
```diff
@@ -27,6 +28,10 @@ pub extern "C" fn wasm_frame_func_index(_frame: &wasm_frame_t<'_>) -> u32 {
 ///
 /// This API is unsupported and will panic upon use.
 #[no_mangle]
+#[cfg_attr(
+    feature = "prefix-symbols",
+    export_name = "wasmi_wasm_frame_func_offset"
+)]
```
Robbepop, line 34: This should use the `#[cfg_attr(feature = "prefix-symbols", wasmi_c_api_macros::prefix_symbol)]` attribute. Probably an oversight due to text replacement without multi-lines.

crates/c_api/src/func.rs
```diff
@@ -141,6 +142,10 @@ pub unsafe extern "C" fn wasm_func_new(
 /// It is the caller's responsibility not to alias the [`wasm_functype_t`]
 /// with its underlying, internal [`WasmStoreRef`](crate::WasmStoreRef).
 #[no_mangle]
+#[cfg_attr(
+    feature = "prefix-symbols",
+    export_name = "wasmi_wasm_func_new_with_env"
+)]
```
Robbepop, line 148: This should use the `#[cfg_attr(feature = "prefix-symbols", wasmi_c_api_macros::prefix_symbol)]` attribute. Probably an oversight due to text replacement without multi-lines.

crates/c_api/src/types/extern.rs
```diff
@@ -75,6 +76,10 @@ pub extern "C" fn wasm_externtype_kind(et: &wasm_externtype_t) -> wasm_externkin
 
 /// Returns a mutable reference to the [`wasm_externtype_t`] as [`wasm_functype_t`].
 #[no_mangle]
+#[cfg_attr(
+    feature = "prefix-symbols",
+    export_name = "wasmi_wasm_externtype_as_functype"
+)]
```
Robbepop, line 82: This should use the `#[cfg_attr(feature = "prefix-symbols", wasmi_c_api_macros::prefix_symbol)]` attribute. Probably an oversight due to text replacement without multi-lines.

Robbepop: Also many more occurrences in the same file.

crates/c_api/src/types/import.rs
```diff
@@ -53,18 +54,24 @@ pub extern "C" fn wasm_importtype_new(
 
 /// Returns a shared reference to the module namespace of the [`wasm_importtype_t`].
 #[no_mangle]
+#[cfg_attr(
+    feature = "prefix-symbols",
+    export_name = "wasmi_wasm_importtype_module"
+)]
```
Robbepop, line 60: This should use the `#[cfg_attr(feature = "prefix-symbols", wasmi_c_api_macros::prefix_symbol)]` attribute. Probably an oversight due to text replacement without multi-lines.

crates/c_api/src/types/memory.rs
```diff
@@ -63,19 +63,28 @@ impl CMemoryType {
...
     let memory_type = MemoryType::new(limits.min, limits.max()).unwrap();
     Box::new(wasm_memorytype_t::new(memory_type))
 }
 
 /// Returns a shared reference to the table limits of the [`wasm_memorytype_t`].
 #[no_mangle]
+#[cfg_attr(
+    feature = "prefix-symbols",
+    export_name = "wasmi_wasm_memorytype_limits"
+)]
```
Robbepop, line 77: This should use the `#[cfg_attr(feature = "prefix-symbols", wasmi_c_api_macros::prefix_symbol)]` attribute. Probably an oversight due to text replacement without multi-lines.

Also many more occurrences in the same file.

crates/c_api/src/types/table.rs
```diff
@@ -79,18 +80,30 @@ pub extern "C" fn wasm_tabletype_new(
 
 /// Returns a shared reference to the element type of the [`wasm_tabletype_t`].
 #[no_mangle]
+#[cfg_attr(
+    feature = "prefix-symbols",
+    export_name = "wasmi_wasm_tabletype_element"
+)]
```
Robbepop, line 86: This should use the `#[cfg_attr(feature = "prefix-symbols", wasmi_c_api_macros::prefix_symbol)]` attribute. Probably an oversight due to text replacement without multi-lines.

Also many more occurrences in the same file.

Robbepop (changes requested):

crates/c_api/src/types/table.rs
xdoardo: Another day, another mess with `:substitute`!

Robbepop (approved):
Thanks a lot for all the work you've put into this!
LGTM

Robbepop:
@xdoardo Seems like this solution isn't great either:
https://github.com/wasmi-labs/wasmi/actions/runs/12527322280/job/34940791110?pr=1336

Background info: https://github.com/rust-lang/rust/issues/47446

But since `#[export_name = ..]` binds stronger I think the simplest solution is to guard `#[no_mangle]` by `not(feature = "prefix-symbols")`.

xdoardo:
Yeah, it makes sense that `#[no_mangle]` and `#[export_name]` used together cause an error. I'll prepare a patch asap.

PR #1321, opened 2024-11-28
Add new security audit report

Robbepop:


NEWS.md
```diff
@@ -1,6 +1,18 @@
 # Wasmi News
 
-## 2024-02-07  -  Announcement: Transfer of Ownership
+## 2024-11-27 - Wasmi Security Audit for v0.36+
+
+In the last months Wasmi v0.36.0 and later versions have been audited by
```
anupsdf, line 5: Maybe remove 'In the last months' in the beginning since in the future years when someone reads this it won't be last few months.

Robbepop: I thought the date right above it was enough to avoid this but I can remove this.

PR #1328, opened 2024-12-01
Add optional `.wat` module support to `Module::new`

Robbepop:
A test-only `wat2wasm` function has been implemented several times in the Wasmi codebase, especially in tests which is an indicator that this functionality was crucially missing from Wasmi's module parsing.

Thus we added this feature optionally behind the new `wat` crate feature in Wasmi.
This also improves Wasmi's Wasmtime API mirroring because Wasmtime also provides `.wat` parsing support in its `Module::new` API.

We explicitly do _not_ support `.wat` parsing in `Module::new_streaming` since there is no streaming parsing support for `.wat` files via the used `wat` crate.


Robbepop:
The `miri` job fails due to timeout, so not because of this PR.

PR #1353, opened 2025-02-03
Use `u64` for linear memory APIs

Robbepop:
Follow-up of https://github.com/wasmi-labs/wasmi/pull/1349.

Not sure if this PR should actually be merged without Wasm `memory64` proposal support.


Robbepop:
It is a better idea to move this work into a future PR to implement Wasm `memory64` proposal support.

Closed.

PR #1371, opened 2025-02-14
Add support for the Wasm `memory64` proposal

Robbepop:
Closes https://github.com/wasmi-labs/wasmi/issues/1357.


Robbepop:
Since the official Wasm spec testsuite was ported over to Wasm 3.0 and Wasmi does not (yet) support Wasm 3.0 we need to have 2 different Wasm spec testsuite Git submodules in our tree: the main one tracking the most up-to-date state so we keep most tests up-to-date, and another one specifically based on [`b66a09a`](https://github.com/WebAssembly/testsuite/commit/b66a09a0ee27c482cda8ab178c4701ea9339afb7) which is the last commit which had all the necessary `memory64` tests neatly in one place.

Robbepop:
I just ran the benchmarks to compare `main` with this PR's branch and saw no performance difference across the entire set of tests. Noticably there were minor improvements but these were all within the margin of error.

PR #1393, opened 2025-03-13
Clean-up `wasmi_core` implementation and remote `UntypedVal` eval API

Robbepop:
- Cleans up the low-level traits underlying to the new `wasm` API.
- Re-designed low-level `Integer` trait.
- Minimally re-design low-level `Float` trait.
- Make `TypedVal` use the new `wasm` eval API for its eval API instead of the old `UntypedVal` eval API.
- Remove the old `UntypedVal` eval API.
- Port all translation tests to the new `wasm` eval API.
- Port Wasm const-expr eval to new `wasm` eval API.


Robbepop:
The flaky `Miri (spec)` CI job failed due to timeout.

PR #1397, opened 2025-03-19
Fix const evaluation in `mul_wide` instructions

Robbepop:
Fixed https://github.com/bytecodealliance/wasmtime/issues/10418.

The const-evaluator pushed the `lo` and `hi` values in reversed order back into the translation stack during const-eval. I checked the other `wide-arithmetic` instructions but there it was all working as expected.

This PR adds a regression test.


Robbepop:
The `clippy` CI job does _not_ fail due to changes made by this PR.

PR #1398, opened 2025-03-19
Explicitly drop `ControlFlow` when not needed in the Wasmi executor

Robbepop:
This fixes (or silences) a `clippy` warning.


Robbepop:
The flaky `miri` CI job failed due to timeout. We have to bump the timeout threshold once again.

PR #1449, opened 2025-03-30
experiment: Add `PrunedStore`

Robbepop:
Closes https://github.com/wasmi-labs/wasmi/issues/1448
Unblocks https://github.com/wasmi-labs/wasmi/issues/1433

### DIFF
--- a/crates/wasmi/src/store.rs
+++ b/crates/wasmi/src/store.rs
@@ -136,7 +136,7 @@ pub struct TypedStore<T> {
     /// Stored host function trampolines.
     trampolines: Arena<TrampolineIdx, TrampolineEntity<T>>,
     /// User provided host data owned by the [`Store`].
-    data: T,
+    data: Box<T>,
     /// User provided hook to retrieve a [`ResourceLimiter`].
     limiter: Option<ResourceLimiterQuery<T>>,
     /// User provided callback called when a host calls a WebAssembly function
@@ -876,7 +876,7 @@ where
             inner: StoreInner::new(&engine),
             typed: TypedStore {
                 trampolines: Arena::new(),
-                data: T::default(),
+                data: Box::new(T::default()),
                 limiter: None,
                 call_hook: None,
             },
@@ -891,7 +891,7 @@ impl<T> Store<T> {
             inner: StoreInner::new(engine),
             typed: TypedStore {
                 trampolines: Arena::new(),
-                data,
+                data: Box::new(data),
                 limiter: None,
                 call_hook: None,
             },
@@ -915,7 +915,7 @@ impl<T> Store<T> {
 
     /// Consumes `self` and returns its user provided data.
     pub fn into_data(self) -> T {
-        self.typed.data
+        *self.typed.data
     }
 
     /// Installs a function into the [`Store`] that will be called with the user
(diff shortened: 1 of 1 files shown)


Robbepop:
I just benchmarked the Wasmi executor that is now using the new `PrunedStore` type and is therefore fully non-generic.

Performance is affected a lot. It seems that Wasmi can still achieve the same performance, but op-codes are affected very differently. For example, `counter` dropped significantly whereas `tiny_keccak` remained stable and `global_bump` even improved ever so slightly.
This usually means that we can fix these performance rollercoaster with some inspection and `#[inline(never)]`, `#[inline(always)]` or normal `#[inline]` annotations. However, this is super flaky and might change with every Rust version. This would be solved if Rust had [`explicit-tail-calls`](https://github.com/rust-lang/rust/issues/112788).

Robbepop:
I tried https://github.com/wasmi-labs/wasmi/tree/rf-add-continue-next-to-hot-loop to be more explicit about Wasmi's state machine nature in its interpreter loop, but contrary to posts I have read this didn't seem to work and just generated about the same codegen which on the other hand is to be expected given that semantically the code is the same.

Robbepop:
After the most recent commits, performance is somewhat on par with `main` again.

## Benchmarks

### Same or Better

#### Real Use Cases
image

#### Loads and Stores
image

### Regressions

image

image

Robbepop:
Despite the remaining performance issues (e.g. `counter`) I think this is good to go. The performance improvements balance out the regressions kind of. We can have follow-up PRs to fix the remaining performance issues.

zamazan4ik:
@Robbepop excuse me for pinging you so late here - I just discovered this PR. I am interested in your findings regarding PGO applicability for Wasmi. I was surprised that PGO didn't help for Wasmi since PGO works great for other interpreters like CPython and V8 (and for [many](https://github.com/zamazan4ik/awesome-pgo) other things).

> I tried to PGO compile Wasmi. This helped when used with a very limited set of benchmarks. For example, when running PGO with the tiny_keccak test case it yielded a performance improvement of ~15%. However, as soon as the benchmark set was expanded there were no performance gains anymore.

Could you please describe a bit more about your PGO optimization process? What was the training workload in this case? Just guessing - maybe we can try to prepare a better training workload set so PGO will work better. Was it Instrumentation or Sampling PGO? I suppose it was Instrumentation but need to ask anyway.

Maybe you have some ready-to-use scripts/whatever else, so I can try to reproduce your results. I am not saying at all that you did something in a wrong way - I am just curious.

Thank you!

Robbepop:
@zamazan4ik I have created an issue for PGO for Wasmi: https://github.com/wasmi-labs/wasmi/issues/1551 Let us continue the talk about PGO for Wasmi there.

To answer your question: it is too long ago to remember all the details. I tried out PGO for Wasmi more of as a side quest to fix the performance problems which I ultimately fixed with some inline annotations if I remember correctly.

PGO indeed resulted in some local performance wins, e.g. for `tiny_keccak` by roughly 15% or so. However, this was only achieved when I used _only_ `tiny_keccak` benchmark as PGO input. I am pretty sure that I haven't done PGO properly since it was my first time trying this out. I am pretty sure a knowledgeable person (which you seem to be about PGO) might yield much better results. I just took Wasmi's benchmarks or subsets of them as PGO input and toyed around with those. Maybe something like `coremark.wasm` is a better PGO input? I have too little experience about PGO to answer those questions.

PR #1455, opened 2025-04-01
chore: remove redundant words

highcloudwind:
remove redundant words

### DIFF
--- a/crates/wasmi/src/module/mod.rs
+++ b/crates/wasmi/src/module/mod.rs
@@ -101,17 +101,17 @@ impl ModuleHeader {
         &self.inner.funcs[func_idx.into_u32() as usize]
     }
 
-    /// Returns the [`GlobalType`] the the indexed global variable.
+    /// Returns the [`GlobalType`] the indexed global variable.
     pub fn get_type_of_global(&self, global_idx: GlobalIdx) -> &GlobalType {
         &self.inner.globals[global_idx.into_u32() as usize]
     }
 
-    /// Returns the [`MemoryType`] the the indexed Wasm memory.
+    /// Returns the [`MemoryType`] the indexed Wasm memory.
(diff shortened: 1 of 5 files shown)


Robbepop (changes requested):
Thanks for the PR and the improvements. Though some of those should replace the `the the` with `of the` and other should replace the pair with `that the` instead.

crates/wasmi/src/module/mod.rs
```diff
@@ -101,17 +101,17 @@ impl ModuleHeader {
         &self.inner.funcs[func_idx.into_u32() as usize]
     }
 
-    /// Returns the [`GlobalType`] the the indexed global variable.
+    /// Returns the [`GlobalType`] the indexed global variable.
```
Robbepop, line 104:
```suggestion
    /// Returns the [`GlobalType`] of the indexed global variable.
```

Robbepop: The others in the same file should be fixed the same.

crates/wasmi/tests/integration/fuel_metering.rs
```diff
@@ -35,7 +35,7 @@ fn default_test_setup(wasm: &[u8]) -> (Store<()>, Func) {
     (store, func)
 }
 
-/// Asserts the the call was successful.
+/// Asserts the call was successful.
```
Robbepop, line 38:
```suggestion
/// Asserts that the call was successful.
```

Robbepop:
Closed due to inactivity. Please feel free to re-open once you implemented the review suggestions. :)

highcloudwind:
@Robbepop Sorry for the late reply. Here's the new PR. https://github.com/wasmi-labs/wasmi/pull/1459  Please review again.

PR #1460, opened 2025-04-08
Move Wasmi's `memory` internals to `wasmi_core`

Robbepop:
Partially implements https://github.com/wasmi-labs/wasmi/issues/1410.


Robbepop:
I am not happy with this PR's diff. I am going to try to split it up into more smaller PRs and see if that works.

Issue #538, opened 2022-10-25
Implement backtraces in `wasmi`


Robbepop:
Even though `wasmi` is an interpreter VM for Wasm we are missing proper debugging functionality such as backtraces.
This is a heavily requested feature by smart contract users so that they can finally properly debug and analyze their smart contract exeuctions.

Having backtraces is not sufficient by its own since we also are required to show proper function and local variable names for a nice debugging UX. For this purpose we also need to support the Wasm `name` section that provides a Wasm VM with those names for a given Wasm blob.

Furthermore any backtrace implementation for `wasmi` should ideally _not_ conflict with the performance in case those backtraces are not required. Therefore we need to find a design that has zero cost for non-debug executions.

@cmichi

Robbepop:
Past discussions: https://github.com/paritytech/wasmi/issues/28

orsinium:
It might be a better idea to generate a [coredump](https://github.com/WebAssembly/tool-conventions/blob/main/Coredump.md) on trap. Then the debug information doesn't need to be attached to the wasm binary and parsed by wasmi. That's one of the debug modes that is supported by [wasmtime](https://docs.wasmtime.dev/examples-debugging-core-dumps.html) and [wasmer](https://github.com/wasmerio/wasmer/blob/main/docs/en/examples-coredump.md).

Would you be interested in me sponsoring your work on this issue? I'm making [Firefly Zero](https://fireflyzero.com/), a wasmi-powered handheld game console, and tracebacks is the most requested feature gamedevs have so far. With the current implementation, if a trap occurs, I can only provide information about the guest function that failed and the last called host function.

Robbepop:
Hi @orsinium, very cool to hear about Firefly Zero and that it uses Wasmi. :) I absolutely love the mascot and the idea.

I personally have not yet taken a look into the technical aspects of coredumps what is going to be necessary for Wasmi to support producing them when encountering a trap but will take a look as I think it is a very valuable addition to Wasmi.

orsinium:
> Implementing it yourself and sponsoring a PR? (great if it meets the quality) Or sponsoring this feature monetarily for me to implement?

The latter. I looked around to see if there is a link to financially sponsor you or wasmi and haven't found any. I want to support the most important project that powers Firefly Zero, especially if it can bring closer some of the important features like the one in question.

I could also look into implementing coredumps for wasmi myself but I'll have the capacity only in October.

orsinium:
[...]

PR #1461, opened 2025-04-08
Add Wasm `coredump` builder

orsinium:
Implements https://github.com/wasmi-labs/wasmi/issues/538.

This is the first step to adding support for coredumps. The PR introduces a `wasmi::coredump` module with `serialize` function. The function can be used to serialize the given process info as coredump. For example:

```rust
let frame = Frame {
    func_idx: 6,
    code_offset: 123,
};
let thread = Thread {
    name: "main",
    frames: &[frame],
};
let proc = Process {
    name: "/usr/bin/true.exe",
    threads: &[thread],
    memories: &[Memory { min: 0, max: None }],
    data: &[],
};
let mut coredump = Vec::new();
serialize(&mut coredump, &proc);
```

The provided code is not used anywhere just yet.

Related to #538

### DIFF
--- a/crates/wasmi/src/coredump.rs
+++ b/crates/wasmi/src/coredump.rs
@@ -1,0 +1,171 @@
+#[derive(Debug, Default)]
+pub struct Process<'a> {
+    pub name: &'a str,
+    pub threads: &'a [Thread<'a>],
+    pub memories: &'a [Memory],
+    pub data: &'a [u8],
+}
+
+#[derive(Debug)]
+pub struct Memory {
+    pub min: u64,
+    pub max: Option<u64>,
+}
+
+#[derive(Debug)]
+pub struct Thread<'a> {
+    pub name: &'a str,
+    pub frames: &'a [Frame],
+}
+
+#[derive(Debug)]
+pub struct Frame {
+    pub func_idx: u32,
+    pub code_offset: u32,
+    // TODO: locals
+    // TODO: stack
+}
+
+use alloc::vec::Vec;
+
+const HEADER: [u8; 8] = [
+    0x00, 0x61, 0x73, 0x6D, // Magic
+    0x01, 0x00, 0x00, 0x00, // Version
+];
+
+/// Convert the given `proc` info into a coredump and write it into `w`.
+///
+/// ## Example
+///
+///     let frame = Frame {
+///         func_idx: 6,
+///         code_offset: 123,
+///     };
+///     let thread = Thread {
+///         name: "main",
+///         frames: &[frame],
+///     };
+///     let proc = Process {
+///         name: "/usr/bin/true.exe",
+///         threads: &[thread],
+///         memories: &[Memory { min: 0, max: None }],
+///         data: &[],
+///     };
+///     let mut coredump = Vec::new();
+///     serialize(&mut coredump, &proc);
+pub fn serialize(w: &mut Vec<u8>, proc: &Process) {
+    w.extend(HEADER);
+    write_core_section(w, proc);
+    write_corestack_sections(w, proc);
+    write_memory_section(w, proc);
+    write_data_section(w, proc);
+}
+
+fn write_core_section(w: &mut Vec<u8>, proc: &Process) {
+    let mut data = Vec::new();
+    data.push(0x0);
+    write_utf8(&mut data, proc.name);
+    w.push(0); // custom section ID
+    encode_custom_section(w, "core", &data);
+}
+
+fn write_corestack_sections(w: &mut Vec<u8>, proc: &Process) {
+    for stack in proc.threads {
+        let mut data = Vec::new();
+        encode_stack(&mut data, stack);
+        w.push(0); // custom section ID
+        encode_custom_section(w, "corestack", &data);
+    }
+}
+
+fn write_memory_section(w: &mut Vec<u8>, proc: &Process) {
+    let mut section: Vec<u8> = Vec::new();
+    for mem in proc.memories {
+        let mut flags = 0;
+        if mem.max.is_some() {
+            flags |= 0b0001;
+        }
+        section.push(flags);
+        write_u64(&mut section, mem.min);
+        if let Some(max) = mem.max {
+            write_u64(&mut section, max);
+        }
+    }
+    w.push(5); // memory section ID
+    let count = proc.memories.len();
+    encode_section(w, count as u32, &section);
+}
+
+fn write_data_section(w: &mut Vec<u8>, proc: &Process) {
+    let mut section = Vec::new();
+    section.push(0x00); // active
+    section.push(0x41); // "i32.const" instruction
+    write_u64(&mut section, 0); // i32.const value (zero)
+    section.push(0x0B); // "end" instruction.
+    write_u64(&mut section, proc.data.len() as u64);
+    section.extend(proc.data);
+
+    w.push(11); // data section ID
+    encode_section(w, 1, &section);
+}
+
+fn encode_custom_section(w: &mut Vec<u8>, name: &str, data: &[u8]) {
+    let encoded_name_len = encoding_size(u32::try_from(name.len()).unwrap());
+    let total_size = encoded_name_len + name.len() + data.len();
+    write_u64(w, total_size as u64);
+    write_u64(w, name.len() as u64);
+    w.extend_from_slice(name.as_bytes());
+    w.extend(data);
+}
+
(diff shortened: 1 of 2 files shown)


orsinium:
I also have a few simple smoke tests but they use third-party std dependencies, so I decided to not bring them into wasmi.

```rust

#[cfg(test)]
mod tests {
    use crate::*;

    #[test]
    fn test_write_u64() {
        use super::write_u64;
        for x in &[0u64, 1, 2, 3, 4, 10, 21, 49, 1033, 1231245, u64::MAX] {
            let mut act = Vec::new();
            write_u64(&mut act, *x);
            let mut exp = Vec::new();
            leb128::write::unsigned(&mut exp, *x).unwrap();
            assert_eq!(act, exp);
        }
    }

    #[test]
    fn test_serialize() {
        let frame = Frame {
            func_idx: 6,
            code_offset: 123,
        };
        let thread = Thread {
            name: "main",
            frames: &[frame],
        };
        let proc = Process {
            name: "/usr/bin/true.exe",
            threads: &[thread],
            memories: &[Memory { min: 0, max: None }],
            data: &[],
        };
        let mut act = Vec::new();
        serialize(&mut act, &proc);

        use wasm_coredump_builder::*;
        let mut coredump = CoredumpBuilder::new().executable_name("/usr/bin/true.exe");
        let mut thread = ThreadBuilder::new().thread_name("main");
        let frame = FrameBuilder::new().codeoffset(123).funcidx(6).build();
        thread.add_frame(frame);
        coredump.add_thread(thread.build());
        let exp = coredump.serialize().unwrap();

        assert_eq!(act, exp);
    }
}
```

orsinium:
Also, clippy naturally fails because the code I've added is unused.

Robbepop:
> I also have a few simple smoke tests but they use third-party std dependencies, so I decided to not bring them into wasmi.
> 
> ```rust
> #[cfg(test)]
> mod tests {
>     use crate::*;
> 
>     #[test]
>     fn test_write_u64() {
>         use super::write_u64;
>         for x in &[0u64, 1, 2, 3, 4, 10, 21, 49, 1033, 1231245, u64::MAX] {
>             let mut act = Vec::new();
>             write_u64(&mut act, *x);
>             let mut exp = Vec::new();
>             leb128::write::unsigned(&mut exp, *x).unwrap();
>             assert_eq!(act, exp);
>         }
>     }
> 
>     #[test]
>     fn test_serialize() {
>         let frame = Frame {
>             func_idx: 6,
>             code_offset: 123,
>         };
>         let thread = Thread {
>             name: "main",
>             frames: &[frame],
>         };
>         let proc = Process {
>             name: "/usr/bin/true.exe",
>             threads: &[thread],
>             memories: &[Memory { min: 0, max: None }],
>             data: &[],
>         };
>         let mut act = Vec::new();
>         serialize(&mut act, &proc);
> 
>         use wasm_coredump_builder::*;
>         let mut coredump = CoredumpBuilder::new().executable_name("/usr/bin/true.exe");
>         let mut thread = ThreadBuilder::new().thread_name("main");
>         let frame = FrameBuilder::new().codeoffset(123).funcidx(6).build();
>         thread.add_frame(frame);
>         coredump.add_thread(thread.build());
>         let exp = coredump.serialize().unwrap();
> 
>         assert_eq!(act, exp);
>     }
> }
> ```

What is the difference between this PR and what `wasm_coredump_builder` does?
Tests would be nice but indeed 3rd party dependencies should be avoided. Ideally we had a few tests that do not depend on 3rd party dependencies.

Robbepop:
@orsinium thanks a lot for your PR!

All in all this PR still needs some work before we can merge. How willing are you to do this? Otherwise I could drive this to finalization.
I also need to know what differs this from `wasm_coredump_builder`. Seems to me (superficially) that both crates more or less do the same thing.

The main question is how users would interact with this feature via Wasmi's API and what the performance implications are (if any) if enabled.

orsinium:
> All in all this PR still needs some work before we can merge. How willing are you to do this? 

I'd be happy to properly integrate it with wasmi but it's a big project and I don't know where to start. I hoped that `Trap` would maybe contain the instruction offset but it doesn't. I'm sure the interpreter knows which instruction caused the trap, so if you can put this information into the trap error, the problem is pretty much solved.

> I also need to know what differs this from wasm_coredump_builder

I've added no_std support into wasm_coredump_builder but it still depends on leb128 which requires std. I tried to add no_std support there as well but they don't want to depend on embedded_io for std, and doing it that way would require so much work:

https://github.com/gimli-rs/leb128/issues/25

So, when I started to dig into both crates, I learned that they don't really do much. So, here is a simple implementation, without std or any dependencies. If you don't want to maintain it as part of wasmi, I can release it as a separate crate.

> The main question is how users would interact with this feature via Wasmi's API and what the performance implications are (if any) if enabled.

My preferred API would be `trap.write_coredump(&mut output)` where `Trap` is `trap` and `write_coredump` is a wrapper around `coredump::serialize`. For that to work (for initial PoC), all Trap needs to know is the function and instruction offset that cause the trap. Using this information, we can form one `Frame`, put it into one `Thread`, put it into one `Process`, and serialize.

> I suppose this serializes to this encoding:
> https://github.com/WebAssembly/tool-conventions/blob/main/Coredump.md
> right?

Yes! It's pretty much a wasm file with some custom sections.

Robbepop:
Concerning your implementation questions: Wasmi does not really store a mapping of the actual functions on the call stack for performance reasons. However, with the information that you can find on the call stack it is possible to compute this information and create the backtrace. This will be slow but having a trap + backtrace situation usually is not a case that needs to be optimized for anyways as long as the backtrace eventually is created, right?

The function call stack can be found here:

- `CallStack`: https://github.com/wasmi-labs/wasmi/blob/v0.44.0/crates/wasmi/src/engine/executor/stack/calls.rs#L24
- `CallFrame`: https://github.com/wasmi-labs/wasmi/blob/v0.44.0/crates/wasmi/src/engine/executor/stack/calls.rs#L196
- `CodeMap`: https://github.com/wasmi-labs/wasmi/blob/main/crates/wasmi/src/engine/code_map.rs#L79
- `CompiledFuncEntity`: https://github.com/wasmi-labs/wasmi/blob/main/crates/wasmi/src/engine/code_map.rs#L763

Here the `CallFrame::instr_ptr` is a pointer into the instructions of a function on the call stack. This `instr_ptr` can be used to backtrack the information about the concrete function on the stack by querying all the `CodeMap`'s function entities and see for all `CompiledFuncEntity` (because uncompiled ones cannot be executed) which is the one that contains the `instr_ptr` in its `instr` slice. This will require to split the `instr` slice into its begin and end pointer and then see if they contain `instr_ptr` with `begin >= instr_ptr && instr_ptr < end`.

Though this won't provide you with the function's name since Wasmi does not (yet) support the Wasm name section and thus cannot associate functions to their names. This will only help to associate it with the index of the function in the `CodeMap` which might be enough information though.
Also information about the function's signature is missing because it is implied in Wasmi's bytecode and not needed for execution. I have no idea how we could backtrace this information. If it is critical we probably have to add it to the executor.

orsinium:
> Though this won't provide you with the function's name

We don't need the function name. All we need is the function index and the instruction index within the function. That's it, that's all we put in coredump. The function name, function signature, line of code etc are inferred by a third-party coredump inspector based on dwarf files or other debug info. It would be helpful to have values of locals and the call stack as well but it's totally fine to keep it out of PoC.

Robbepop:
> > Though this won't provide you with the function's name
> 
> We don't need the function name. All we need is the function index and the instruction index within the function. That's it, that's all we put in coredump. The function name, function signature, line of code etc are inferred by a third-party coredump inspector based on dwarf files or other debug info. It would be helpful to have values of locals and the call stack as well but it's totally fine to keep it out of PoC.

- I suppose you need the Wasm function index. For that, as described in my last post, we'd have to add this field to the `CodeMap` because right now this information is simply not available.
- For the instruction index within the function you probably also mean the Wasm instruction index. This is very tricky and unlikely to be implemented easily since Wasm instructions are not Wasmi instructions. A single Wasmi instruction may represent multiple Wasm instructions. So it will be very hard to provide a proper mapping between the two.

Robbepop:
Concerning user facing API of how to get the Wasm coredump, I think we should simply do what Wasmtime does and embed `WasmCoreDump` to Wasmi's `Error` type when this option is enabled.
https://docs.rs/wasmtime/latest/wasmtime/struct.WasmCoreDump.html

PR #1490, opened 2025-04-28
Fix `rust-analyzer` issues

Robbepop:
Not a Wasmi bug or issue but this fixes an annoying issue I am seeing with `rust-analyzer` lately.


Robbepop:
I kinda missed that this commit invalidly contained some renamings. This was not supposed to be but it accidentally happened.

PR #1498, opened 2025-05-02
Allow function resumption after running out of fuel

Robbepop:
Closes https://github.com/wasmi-labs/wasmi/issues/696.

cc @rib @tomaka @AbleTheAbove

This PR implements this `ResumableCall` design:
image


Robbepop:
At this point the feature is mostly implemented and even integrated into Wasmi's Wast runner to use the resumable call API to execute the official Wasm spec testsuite and ... it _mostly_ works. A few (2-3) test cases do not yet work properly and need to be investigated.

PR #1501, opened 2025-05-05
Update to `wasm-tools` v229

Robbepop:


Robbepop:
Due to the massive performance regressions between `wasmparser` v228 -> v229 I decided _not_ to update to `wasm-tools` v229 for the upcoming Wasmi v0.45.0 release.

Robbepop:
I am closing this since this PR is already very old and stale thus might need another try if we ever want to port to a newer `wasmparser` version. Alternatively we could fork `wasmparser` and have our own specialized version of it: https://github.com/wasmi-labs/wasmi/issues/1514

polarathene:
> might need another try if we ever want to port to a newer `wasmparser` version.

It might be worth verifying if the regression is still present given it's been around a year since the attempt to upgrade to v235?

That said if it's not worth the time to get into sync with upstream I'd understand that too 👍

Robbepop:
@polarathene 

> > might need another try if we ever want to port to a newer `wasmparser` version.
> 
> It might be worth verifying if the regression is still present given it's been around a year since the attempt to upgrade to v235?
> 
> That said if it's not worth the time to get into sync with upstream I'd understand that too 👍 

The main problem was not even the performance regression but certain API changes that were really difficult to deal with for how Wasmi uses the wasmparser crate. So porting became too much of a hassle for too little (or even negative) gains thus I did not bother.

PR #1502, opened 2025-05-06
Set `WasmFeatures` for `wasmparser::Parser` in `ModuleParser`

Robbepop:
This was an oversight that caused bugs when upgrading to `wasmparser` v0.229.0.


Robbepop:
Wasmi translation performance is not affected by this change.

PR #1507, opened 2025-05-08
Drop `T: 'static` bounds from `wasmi::Store` again

Robbepop:
Closes https://github.com/wasmi-labs/wasmi/issues/1503.

cc @juntyr


juntyr:
Thanks @Robbepop!

PR #1512, opened 2025-05-10
Replace register-alloc with purely stack-based translation

Robbepop:
Closes https://github.com/wasmi-labs/wasmi/issues/1484.

### DIFF
--- a/crates/wasmi/src/engine/translator/mod.rs
+++ b/crates/wasmi/src/engine/translator/mod.rs
@@ -10,6 +10,7 @@ mod labels;
 mod provider;
 mod relink_result;
 mod stack;
+mod stack2;
 #[macro_use]
 mod utils;
 mod visit;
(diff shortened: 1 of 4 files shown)


Robbepop:
I benchmarked the new Wasmi `FuncTranslator` against the `main` one.

Especially the `translate/x/eager/unchecked` benchmarks yield roughly 20-35% regressions compared to `main`.

# `main` - `translate/spidermonkey/eager/unchecked`

flamegraph main

# `pr` - `translate/spidermonkey/eager/unchecked`

flamegraph pr

After seeing this difference I put `#[inline(never)]` on all `visit` trait methods.

# `pr.v2` - `translate/spidermonkey/eager/unchecked`

flamegraph pr2

This resulted in significant performance gains across the board. However, `translate/x/eager/unchecked` is still roughly 10-25% slower than on `main`. Further investigation is required.

PR #1528, opened 2025-05-30
Bump Wasmtime dependencies to v33

Robbepop:


Robbepop:
It seems that Wasmtime removed its faster `SinglePass` register allocator (due to bugs) which we use in Wasmi's differential fuzzing since it prevents fuzzing timeouts by being _WAY_ faster than the other register allocator. We likely want to skip updating Wasmtime dependencies until the faster `SinglePass` register allocator is available again.

PR #1540, opened 2025-06-15
Deprecate `Module::new_streaming[_unchecked]` APIs

Robbepop:
The `Module::new_streaming[_unchecked]` APIs existed in Wasmi since the very beginning (thought differently named).
They implement the streaming Wasm module creation interface which is supposedly more efficient than the non-streaming API if Wasm bytes transmission and Wasm module translation are both slow. The idea is to stream in bytes in order to build up a Wasm module incrementally. However, the streaming nature adds overhead to parsing which kinda nullifies the gains. Also Wasmi translation is extremely fast, so streaming usually adds more overhead than is reduced.

As I am not 100% certain that nobody would ever have a serious use for this API, I will deprecate it first and let users know that they shall notify me here in case they have a valid use case for this API.

Removing the streaming APIs make room for a slightly more efficient and lightweight Wasm parser which could simplify the task at: https://github.com/wasmi-labs/wasmi/issues/1514

### DIFF
--- a/crates/wasmi/src/module/mod.rs
+++ b/crates/wasmi/src/module/mod.rs
@@ -244,6 +244,14 @@ impl Module {
     /// - If Wasmi cannot translate the Wasm bytecode.
     ///
     /// [`Config`]: crate::Config
+    #[deprecated(
+        since = "0.48.0",
+        note = "\
+            This API has been deprecated because it is inefficient and unserused. \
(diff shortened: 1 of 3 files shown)


orsinium:
(this comment is moved into https://github.com/wasmi-labs/wasmi/issues/1515)

PR #1543, opened 2025-06-17
Use niche optimization for `Instr` via `NonZeroU32`

Robbepop:


Robbepop:
Benchmarks show a consistent 5-6% regression in translation performance. Therefore we probably should _not_ merge this, as this PR was meant as optimization.

PR #1583, opened 2025-07-30
Improve Munal OS combination mark for README

Robbepop:


Robbepop:
Permission was granted to use the new logo, therefore we can merge this now.

PR #1590, opened 2025-08-01
Update Firefly Zero logo

orsinium:


Robbepop:
@orsinium thanks for improving the Firefly Zero logo!

A little personal nit: I think the font size could still be increased. This is how it looks on my screen:

image

A bit too small for my eyes to have a pleasant feeling when reading the brand name. But we can still merge if you want it just like that.

orsinium:
I've updated the logo. Now instead of Lux it has a lamp which has fewer fine details and matches the text size. Now I'm happy with the result.

image

image

orsinium:
A few force-pushes later, I've managed to sign the commit since the project requires it ✅

Robbepop:
@orsinium Looks way way better now. Great idea to use the smaller lightbulp logo! :)

PR #1638, opened 2025-08-14
Allocate fewer function local constants

Robbepop:
This allocates fewer function local constants by encoding more `copy` instructions if they can encode their immediate values inline, e.g. `copy_imm32`, `copy_i64imm32` or `copy_f64imm32`.

### DIFF
--- a/crates/wasmi/src/engine/translator/func/visit.rs
+++ b/crates/wasmi/src/engine/translator/func/visit.rs
@@ -790,7 +790,7 @@ impl<'a> VisitOperator<'a> for FuncTranslator {
             }
         }
         // Case: fallback to generic `memory.grow` instruction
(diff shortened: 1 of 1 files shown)


Robbepop:
Benchmarks for compiling `spidermonkey.wasm`'s `(func 0)`:

| What | `main` | PR | Diff |
|:--|:-:|:-:|:-:|
| Total number of instructions | 14760 | 14864 | +104 |
| Stack slots reserved for function local constants | 1476 | 1467 | -9 |
| Stack slots reserved for parameters and locals | 57 | 57 | 0 |
| Stack slots (total) | 1533 | 1524 | -9 |

Not a big win with this PR but still the right direction. Long way to go for significant improvements.

PR #1655, opened 2025-09-10
Integrate Wasmi IR 3.0 into `wasmi`

Robbepop:


Robbepop:
All tests are passing again and with this the PR is no longer a draft and could potentially be merged once we see fit.

jprochazk:
Hi! Excuse the drive-by, I read https://wasmi-labs.github.io/blog/posts/wasmi-v1.0/ and stumbled upon this.

> - Interpretation loop must be panic-free for tail calls to be properly generated without the become keyword.
>   - This means that we have to wrap host function calls with ::core::panic::catch_unwind and convert potential panics into flags signalling bad state to the interpreter.

If you can live with 4 parameters on each dispatch handler instead of 6, then you can continue to use regular `fn`, which supports unwinding. Windows calling convention reserves 4 int registers. It might even make codegen a little better for instructions which don't currently use those extra parameters, as they'd be able to use those registers for intermediate values.

`catch_unwind` is a _very_ heavy function to call, and wrapping every native call in it may be a lot worse than paying the cost of an extra indirection for `mem0` in every instruction handler. At least it's worth thinking about :)

Robbepop:
> Hi! Excuse the drive-by, I read https://wasmi-labs.github.io/blog/posts/wasmi-v1.0/ and stumbled upon this.

No excuses needed, that's the whole point of the article. :)

> If you can live with 4 parameters on each dispatch handler instead of 6, then you can continue to use regular `fn`, which supports unwinding. Windows calling convention reserves 4 int registers. It might even make codegen a little better for instructions which don't currently use those extra parameters, as they'd be able to use those registers for intermediate values.
> 
> `catch_unwind` is a _very_ heavy function to call, and wrapping every native call in it may be a lot worse than paying the cost of an extra indirection for `mem0` in every instruction handler. At least it's worth thinking about :)

Nothing set in stone, yet. However, the plan is to use `sysv64` calling convention on Windows which has 6 integer registers for use.
While benchmarking I found out that it is already good enough to simply not have panics in hot-paths of execution handlers. So it is not a must-have to not have panics at all but it is important to "outline" all the functions that may panic, e.g. via `#[inline(never)]` and `#[cold]`.

Thank you for your input! I didn't know how costly `catch_unwind` would actually be. Maybe not the best idea then to even consider using it.

PR #1673, opened 2025-10-17
chore: fix comment for crates/wasmi/src/instance/mod.rs

suranmiao:
fix comment for crates/wasmi/src/instance/mod.rs


Robbepop (approved):
Thank you for the fix.

Issue #1672, opened 2025-10-11
wasm-c-api wasm_func_call error


haroel:
In an iOS/Objective-C++ project, I use wasmi, and I compiled the iOS static library for wasmi myself.
I used a test project to verify whether wasmi can work properly. Here is the calling code (addition operation).

```

unsigned char add_wasm[] = {
  0x00, 0x61, 0x73, 0x6d, 0x01, 0x00, 0x00, 0x00, 0x00, 0x0f, 0x08, 0x64,
  0x79, 0x6c, 0x69, 0x6e, 0x6b, 0x2e, 0x30, 0x01, 0x04, 0x00, 0x00, 0x00,
  0x00, 0x01, 0x0a, 0x02, 0x60, 0x00, 0x00, 0x60, 0x02, 0x7f, 0x7f, 0x01,
  0x7f, 0x03, 0x03, 0x02, 0x00, 0x01, 0x07, 0x1b, 0x02, 0x11, 0x5f, 0x5f,
  0x77, 0x61, 0x73, 0x6d, 0x5f, 0x63, 0x61, 0x6c, 0x6c, 0x5f, 0x63, 0x74,
  0x6f, 0x72, 0x73, 0x00, 0x00, 0x03, 0x61, 0x64, 0x64, 0x00, 0x01, 0x0a,
  0x0d, 0x02, 0x03, 0x00, 0x01, 0x0b, 0x07, 0x00, 0x20, 0x00, 0x20, 0x01,
  0x6a, 0x0b
};
unsigned int add_wasm_len = 86;

int WasmiWrapper::runAddFunction(int a, int b) {
    pImpl->lastError = "";
    int result = -1;
    // 确保模块已加载
    if (!loadModule({})) {
        return result;
    }
    
    // 将 Wasm 模块字节码转换为 wasm_byte_vec_t
    wasm_byte_vec_t wasm_bytes;
    
    // 调试：打印字节码信息
    std::cout << "add_wasm_len: " << add_wasm_len << std::endl;
    std::cout << "add_wasm 前4字节: " << std::hex 
              << (int)add_wasm[0] << " " 
              << (int)add_wasm[1] << " " 
              << (int)add_wasm[2] << " " 
              << (int)add_wasm[3] << std::dec << std::endl;
    
    // 使用 wasm_byte_vec_new 创建副本
    wasm_byte_vec_new(&wasm_bytes, add_wasm_len, (wasm_byte_t*)add_wasm);
    
[...]

PR #1676, opened 2025-10-23
Fix `simd` usage in Wasmi's C-API

Robbepop:
Closes https://github.com/wasmi-labs/wasmi/issues/1672.

Wasmi implemented `simd` support superficially while the official C-API does not. This caused `wasm_val_t` and `wasm_valtype_t` types to be misaligned which caused issues.

@haroel can you confirm that this PR fixes the underlying issue?

### DIFF
--- a/crates/c_api/src/types/val.rs
+++ b/crates/c_api/src/types/val.rs
@@ -34,10 +34,6 @@ pub enum wasm_valkind_t {
     ///
     /// Wraps [`ValType::F64`].
     WASM_F64 = 3,
-    /// A Wasm `v128` value type.
-    ///
-    /// Wraps [`ValType::V128`].
-    WASM_V128 = 4,
     /// A Wasm external reference type.
     ///
     /// Wraps [`ValType::ExternRef`].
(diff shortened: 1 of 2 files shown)


haroel:
> Closes #1672.
> 
> Wasmi implemented `simd` support superficially while the official C-API does not. This caused `wasm_val_t` and `wasm_valtype_t` types to be misaligned which caused issues.
> 
> @haroel can you confirm that this PR fixes the underlying issue?

Yes, but my usage scenario is relatively simple,  with only 3 exported interfaces. I'm not sure if all possible usage cases are covered.

@Robbepop Can you create a GitHub Action unit test for wasm-c-api? I think many developers use the wasm-c mode.

Robbepop:
@haroel I have just recently thought about the need to add a test framework for Wasmi's Wasm C-API implementation. There does not exist one, yet, which is also the reason why the C-API is flagged as experimental in Wasmi's README. Shouldn't be too hard to get the infrastructure started. I have seen on the Wasmtime side, what they do for their C-API tests and it looks reasonable. Adding tons of tests will be the true ground work. Obviously, high prio to have tests but there are tons of other high prio work items and a day has only so many hours.

PR #1724, opened 2025-12-06
Fix `Send` and `Sync` impls of `ResumableCall*` related types

Robbepop:
Closes https://github.com/wasmi-labs/wasmi/issues/1720.

cc @Robert-MacWha


Robbepop:
note to self: I need to split this PR into 2 - one for the actual fix and one for the new docs and refactorings.

PR #1727, opened 2025-12-09
fix `Config::set_max_stack_height` check

gramseyer:
Previous implementation of set_max_stack_height made it impossible to raise the max stack height.


Robbepop:
@gramseyer thanks for the fix, changes look good to me!

I will backport this fix to Wasmi `v1.0`.

Robbepop (approved):

PR #1728, opened 2025-12-10
Add conversion from `ResumableHostError` to `dyn HostError` via `ErrorKind::as_host`

gramseyer:
In earlier versions of wasmi (I can confirm this worked on 0.36.5, at least), one could use Error::downcast to access a HostError returned from a host function.  On 1.0.3, however, the downcast fails.  It appears that the root cause is that the ErrorKind returned from a host function has changed to a ResumableHostTrap( HostError ), and as_host() returns None.
Is this change intended behavior?

This change makes the HostError accessible from an ErrorKind::ResumableHostTrap, as it is from an ErrorKind::Host.

### DIFF
--- a/crates/wasmi/src/engine/resumable.rs
+++ b/crates/wasmi/src/engine/resumable.rs
@@ -66,6 +66,16 @@ impl ResumableHostTrapError {
         self.host_error
     }
 
+    /// Returns a shared reference to the underlying [`Error`].
+    pub(crate) fn as_error_ref(&self) -> &Error {
+        &self.host_error
+    }
+
+    /// Returns a mutable reference to the underlying [`Error`].
+    pub(crate) fn as_error_mut(&mut self) -> &mut Error {
+        &mut self.host_error
+    }
+
     /// Returns the [`Func`] of the [`ResumableHostTrapError`].
     pub(crate) fn host_func(&self) -> &Func {
         &self.host_func
--- a/crates/wasmi/src/error.rs
+++ b/crates/wasmi/src/error.rs
@@ -76,6 +76,11 @@ impl Error {
         &self.kind
     }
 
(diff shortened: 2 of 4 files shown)


gramseyer:
Apologies for the noise with the formatting commits; I haven't figured out how to replicate cargo-fmt with the same settings as you're using in CI.

Robbepop:
> Apologies for the noise with the formatting commits; I haven't figured out how to replicate cargo-fmt with the same settings as you're using in CI.

No problem! Ideally you should simply use `cargo +nightly fmt` locally.

Robbepop:
Hi @gramseyer and thank you for the PR!

It is important that we allow for this use-case again. From what I have seen this use-case is currently not possible since `ResumableHostError` is converted to `Error` as `ErrorKind::ResumableHostError` when calling a Wasm function non-resumably. Ideally, it was converted into `ErrorKind::Host`, instead, since a non-resumable function call shouldn't return resumable errors to begin with.

So the main issue is that this `impl` is being used for non-resumable calls:

```rust
impl From<ExecutionOutcome> for Error {
    fn from(error: ExecutionOutcome) -> Self {
        match error {
            ExecutionOutcome::Host(error) => error.into(),
            ExecutionOutcome::OutOfFuel(error) => error.into(),
            ExecutionOutcome::Error(error) => error,
        }
    }
}
```

Ideally we use a conversion from resumable errors to non-resumable errors instead.

Robbepop:
I opened a PR that implements the proposed changes: https://github.com/wasmi-labs/wasmi/pull/1730

This fixes your provided test and is a more minimal solution for the problem.

Robbepop:
Closed in favor of https://github.com/wasmi-labs/wasmi/pull/1730.

#1730 does not add additional public APIs and is a more minimal solution.
Thanks @gramseyer again for opening the PR and adding the test case!

PR #1730, opened 2025-12-10
Return non-resumable `Error`s in `Func::call` et. al.

Robbepop:
cc @gramseyer


Robbepop:
@gramseyer does this need to be backported to Wasmi v1.0?

gramseyer:
Thank you for the quick fix!  I don't know if a backport is necessary (personally, I can work off of the main branch of this repo)

Robbepop:
> Thank you for the quick fix! I don't know if a backport is necessary (personally, I can work off of the main branch of this repo)

@gramseyer I checked on the `v1.x-dev` branch and your provided test works there so I assume the semantics already are working as intended for Wasmi v1.0. :)

PR #1742, opened 2025-12-19
Add `PartialEq` and `Eq` to `ExternType`

BoilingFusion:
Also derive the PartialEq and Eq trait for the ExternType enum.

### DIFF
--- a/crates/wasmi/src/instance/exports.rs
+++ b/crates/wasmi/src/instance/exports.rs
@@ -114,7 +114,7 @@ impl Extern {
 /// The type of an [`Extern`] item.
 ///
 /// A list of all possible types which can be externally referenced from a WebAssembly module.
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq, Eq)]
 pub enum ExternType {
     /// The type of an [`Extern::Global`].
     Global(GlobalType),


BoilingFusion:
I could not find the run-local-ci.sh script so I didn't run that. Its supposed to be at the script folder at the root of the repo right?

Robbepop:
Hi @BoilingFusion and thank you for the feature request.
I just checked the Wasmtime API and it seems they are not implementing thise traits for the `ExternType` type. Since the Wasmtime API mirror is an imporant feature for Wasmi this will not merge before Wasmtime has not adopted those trait impls.

What you can do is to open a feautre request in the Wasmtime repo. Maybe though there are technical reasons that explain why those trait impls are missing.

BoilingFusion:
Hello @Robbepop thank you for the quick response. From what I can tell from the Wasmtime docs it looks like it has to do with the sub-typing of the Wasmtime [`RefType`](https://docs.rs/wasmtime/39.0.1/wasmtime/struct.RefType.html).
I'll look a bit further into it.

Robbepop:
@BoilingFusion if subtyping prevents `PartialEq` and `Eq` impls on `ExternType` in Wasmtime, then we won't be able to implement those traits in Wasmi either since the Wasm proposals that add these features are planned to be implemented in Wasmi in the future.

Robbepop:
@BoilingFusion please keep me updated about your findings so that we can either close this or keep this open until Wasmtime adds this API in case the API is aligned with Wasmtime semantics.

BoilingFusion:
So far [this commit](https://github.com/bytecodealliance/wasmtime/commit/ff93bce067403bba6bf61cfd7f871e14933a8ad3) gave a good overview:
The `Eq` and `PartialEq` traits where replaced with [`ValType::matches`](https://docs.rs/wasmtime/39.0.1/wasmtime/enum.ValType.html#method.matches) and `FuncType::matches`. This was done since the `matches` functions respects sub-typing which is whats desirable in most cases.

I opened [this feature request](https://github.com/bytecodealliance/wasmtime/issues/12193) to ask if Wasmtime plans on adding `matches` functions to `ExternType` and other type types since that is basically the feature I would want.

Since Wasmtime doesn't have `Eq` and `PartialEq` implemented on `ValType` wasmi should also lack those implementations right? I think the best thing would be to wait with removing those implementations until wasmi has sub-types support and corresponding `matches` functions.

So since `Eq` and `PartialEq` will most likely be replaced by `matches` anyway this pull request can be closed. Maybe replaced with one requesting a `matches` implementation on `ExternType` (depending on if Wasmtime plans on adding that).

Robbepop:
@BoilingFusion thank you for the detailed explanation!

Closed in favor of https://github.com/wasmi-labs/wasmi/issues/1746.

PR #1750, opened 2025-12-27
Make Wasmi's `Table` use `Ref` instead of `Val`

Robbepop:
Closes https://github.com/wasmi-labs/wasmi/issues/1747.

This also introduces the lower-level `TypedRef` and `UntypedRef` types which are 64-bit independent of `simd` crate feature enablement unlike `UntypedVal` which was used prior.

A follow-up PR would reduce the 64-bit `UntypedRef` to a 32-bit version that drops the `Store`-owner information as it is the same for all references stored in a single `Table` and also within a single Wasm execution. The `Store`-owner information can reside in a new dedicated `wasmi::Table` type wrapping `wasmi_core::Tabel`. The remaining question is where the information is stored during Wasm executions to restore it upon returning or calling host functions.

### DIFF
--- a/crates/c_api/src/ref.rs
+++ b/crates/c_api/src/ref.rs
@@ -33,7 +33,7 @@ pub struct wasm_ref_t {
 wasmi_c_api_macros::declare_own!(wasm_ref_t);
 
 impl wasm_ref_t {
-    /// Creates a new boxed [`wasm_ref_t`] from the given [`WasmRef`].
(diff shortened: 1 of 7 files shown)


Robbepop:
The `fuzz` bug detected in the translation fuzzer has nothing to do with the changes introduced in this PR.
Fixed in https://github.com/wasmi-labs/wasmi/pull/1751

PR #1755, opened 2025-12-31
Use 64-bit cells for Wasmi's executor

Robbepop:
Closes https://github.com/wasmi-labs/wasmi/issues/1734.

## Benchmarks

This PR significantly improves performance when the `simd` crate feature is enabled, especially for use cases that do not make use of `simd` functionality. Now turning on `simd` will no longer pessimize performance or memory consumption.

### `tiny_keccak` Benchmark

| SIMD | `main` | PR |
|:-:|--:|--:|
| `false` | 117ms | 117ms |
| `true` | 131ms | 117ms |

## Outstanding Bugs:

- [x] Fix calling host functions via executor.
    - The underlying problem is that host functions still use `len_params` and `len_results` referring to the number of parameters and results respectively. However, with the changes introduced in this PR this needs to change to the respective number of cells for parameters and results instead.

## Follow-Up PRs

- Resolve remaining `TODO`s that are not that important.
    - [ ] Return `BoundedSlotSpan` in the `copy_if_immediate` method.
    - [ ] Remove `StackLayout::local_to_slot` method and uses.
    - [ ] Remove `ControlFrameBase::len_branch_params`.
        - Requires refactoring of `encode_copies` API.
    - [ ] Encode `LocalOperand` copy in local preservation without `Operand::from`.

## ToDo

- [x] Resolve remaining `TODO`s in the code base.
    - [x] Replace `ControlFrameBase::len_branch_values` entirely with `ControlFrameBase::branch_slots`.
    - [x] Add proper `panic` messages and remove usage of `unwrap`.
- [x] Define a new set of types and traits for copying between `Cell`s and Wasm types.
    - New traits: `StoreToCells`, `LoadFromCells`, `LoadFromCellsByValue`, `CellsReader`, `CellsWriter`
    - New types: `Cell`, `CellError`, `LoadByVal`
- [x] Support new traits and types in Wasmi's executor and marshalling between Wasmi and host.
- [x] `global.get` operator needs `global_get64` and `global_get128` forms.
- [x] `global.set` operator needs `global_set64` and `global_set128` forms.
- [x] Use `copy_span_{asc,des}` instead of `copy` for values of type `v128`.
- [x] `return_slot` operator will be used for 64-bit values.
    - To return `v128` values one needs to do `copy_span`, followed by a simple `return` operator.
    - The Wasmi translation needs to be updated to reflect this.
- [x] Add `select128` operator for `v128` type values which does not fuse with `cmp` operators for simplicity.
- [x] Wasmi translator need support for multi-cell mappings for `V128` typed values.
    - [x] `StackLayout` now stores the cell offsets for all registered locals.
    - [x] `Stack` now stores the temporary cell offsets for all operands.
        - `Operand` now stores and provides both `temp_slot: Slot` and `pos: StackPos`.
        - `LocalOperand`s cell offsets need to be computed via `StackLayout`.
- [x] Wasmi's `IntoFunc` trait needs to be updated to the new interface.
- [x] `PrunedStore` API needs to be updated to the new interface.
- [x] Removed `wasmi_core`'s `EncodeUntypedSlice` and `DecodeUntypedSlice`.
- [x] Removed `value` components of the `WasmTyList` trait.
- [x] Removed `CallParams` and `CallResults` traits.

### DIFF
--- a/crates/wasmi/src/engine/translator/func/mod.rs
+++ b/crates/wasmi/src/engine/translator/func/mod.rs
@@ -811,9 +811,9 @@ impl FuncTranslator {
     fn preserve_all_locals(&mut self) -> Result<(), Error> {
         let consume_fuel_instr = self.stack.consume_fuel_instr();
         for local in self.stack.preserve_all_locals() {
-            debug_assert!(matches!(local, Operand::Local(_)));
             let result = local.temp_slots().head();
-            let Some(copy_instr) = Self::make_copy_instr(result, local, &mut self.layout)? else {
+            let Some(copy_instr) = Self::make_copy_instr(result, local.into(), &mut self.layout)?
+            else {
                 unreachable!("`result` and `local` refer to different stack spaces");
             };
             self.instrs
--- a/crates/wasmi/src/engine/translator/func/stack/operands.rs
+++ b/crates/wasmi/src/engine/translator/func/stack/operands.rs
@@ -529,8 +529,7 @@ impl PreservedAllLocalsIter<'_> {
 }
 
 impl Iterator for PreservedAllLocalsIter<'_> {
-    // TODO: return `LocalOperand` instead
-    type Item = Operand;
+    type Item = LocalOperand;
 
     fn next(&mut self) -> Option<Self::Item> {
         if !self.has_remaining_locals() {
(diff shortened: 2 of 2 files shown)


Robbepop:
At this point all tests are passing again. Benchmarks clearly show that performance is no longer decreased when enabling the `simd` crate feature which was the whole point of this PR.

For the sake of not exploding the size of this PR any further I am going to merge it and handle the remainings TODOs in follow-up PRs instead.

PR #1785, opened 2026-02-14
Add `wasmi_wasi::add_to_externals` and use it in the Wasmi CLI application

Robbepop:
Part of https://github.com/wasmi-labs/wasmi/issues/1782.

### DIFF
--- a/crates/wasi/src/sync/snapshots/preview_1.rs
+++ b/crates/wasi/src/sync/snapshots/preview_1.rs
@@ -177,15 +177,15 @@ macro_rules! add_funcs_to_externals {
         );+ $(;)?
     ) => {
         /// Add Wasi APIs imported by `module` to the `externals` buffer.
-        /// 
+        ///
         /// # Note
-        /// 
+        ///
         /// - This only supports Wasi (preview 1) or `wasip1`.
         /// - The `externals` buffer is cleared before being populated.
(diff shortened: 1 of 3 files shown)


crates/wasi/src/sync/snapshots/preview_1.rs
```diff
@@ -59,57 +61,78 @@ pub trait AddWasi<T> {
 /// Look [here](https://github.com/WebAssembly/WASI/blob/snapshot-01/phases/snapshot/docs.md) for more details.
 pub fn add_wasi_snapshot_preview1_to_linker<T, U>(
     linker: &mut Linker<T>,
-    wasi_ctx: impl Fn(&mut T) -> &mut U + Send + Sync + Copy + 'static,
+    get_ctx: impl Fn(&mut T) -> &mut U + Send + Sync + Copy + 'static,
 ) -> Result<(), Error>
 where
+    U: 'static,
+    T: Send + 'static,
```
bjorn3, line 68: Why was this bound added? In my case the store context is `!Send` and has a lifetime.

Robbepop: dang! great catch @bjorn3 . this was on my radar but I completely forgot about fixing it. will ship a fix in the next 2.0 release.

PR #1800, opened 2026-02-18
Disable `wasmi_wast` default features be default

Robbepop:
I was wondering why `cargo build -p wasmi_cli` was using all those debug features such as `portable-dispatch` and `extra-checks` without specifying them explicitly.

Turned out, that since `wasmi_cli` depends on `wasmi_wast` it pulled in all of its default features.
The fix is easy: simply disable default features in the workspace dependency.

So `cargo build -p wasmi_cli` now builds with the proper feature set.


Robbepop:
The `miri` CI jobs failed due to this bug in `cargo-nextest`: https://github.com/nextest-rs/nextest/issues/3073

PR #1804, opened 2026-02-19
Don't use `From` as supertrait in `Handle` trait

Robbepop:
This should resolve this user issue: https://github.com/DouglasDwyer/wasm_runtime_layer/pull/70

cc @juntyr : I have tested this on a local branch and it indeed fixes your issue.


Robbepop:
The `miri` CI jobs failed due to this bug in `cargo-nextest`: https://github.com/nextest-rs/nextest/issues/3073

PR #1834, opened 2026-04-02
Fix operand order in `v128_select_ssss`

shumbo:
This PR fixes a miscompilation in SIMD `v128.select`.

`v128_select_ssss` is executed with operands in the order `(result, condition, true_val, false_val)`, but the translator currently emits `(result, condition, false_val, true_val)` when lowering `v128.select`. This causes the instruction to select the opposite branch.

Other select variants, including `u32_select_*` and `u64_select_*`, already use the correct operand order, so the issue is specific to the `v128` lowering path.

Regression test:

```wast
(module
  (func (export "v128_select") (param i32) (result v128)
    (select (result v128)
      (v128.const i64x2 0x1111111111111111 0x2222222222222222)
      (v128.const i64x2 0x3333333333333333 0x4444444444444444)
      (local.get 0)
    )
  )
)
(assert_return (invoke "v128_select" (i32.const 1)) (v128.const i64x2 0x1111111111111111 0x2222222222222222))
(assert_return (invoke "v128_select" (i32.const 0)) (v128.const i64x2 0x3333333333333333 0x4444444444444444))
```

```
$ git rev-parse HEAD
9f6ff82fc11baea9a3c1232e13f2304fdc49d432
$ cargo run --bin wasmi --features simd -- wast test.wast
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.05s
     Running `target/debug/wasmi wast test.wast`
Error: failed directive on test.wast:10:1

Caused by:
    0: evaluation mismatch of `v128` values
    1: expected = [1229782938247303441, 2459565876494606882], actual = [3689348814741910323, 4919131752989213764]
```

The test passes with this patch.

### DIFF
--- a/crates/wasmi/src/engine/translator/func/mod.rs
+++ b/crates/wasmi/src/engine/translator/func/mod.rs
@@ -2052,7 +2052,7 @@ impl FuncTranslator {
             let false_val = self.copy_if_immediate(false_val)?;
             self.push_instr_with_result(
                 ty,
-                |result| Op::v128_select_ssss(result, condition, false_val, true_val),
+                |result| Op::v128_select_ssss(result, condition, true_val, false_val),
                 FuelCostsProvider::base,
             )?;
             return Ok(());


Robbepop:
@shumbo thank you a lot for the bug fix and especially for the additional regression test!

Robbepop:
The audit CI job is failing for reasons not associated to this PR's changes.

Robbepop:
I tested and verified the bug and fix locally. @shumbo thank you again for the bug report and fix and sorry for taking so long to properly respond.

Robbepop (approved):
LGTM

PR #1855, opened 2026-05-07
Introduce accumulator based interpreter architecture

### LINKED ISSUE #1737: Pre-multiply `Slot` indices by 8
Currently `Slot` is a thin-wrapper around a `u16` value representing indices into a functions stack slots.
During execution Wasmi has to multiply those stack slot indices with 8 to get the byte offset.
This usually isn't a big deal on most platforms as machine instructions such as `lea` provide multiplies of power-of-tows out of the box (up to a limit).

This will also necessitate having `u32` indices for `Slot`s since otherwise their index values will likely go out of range for common Wasm functions.

However, there might still be the case that I-cache pressure will drop due to the following reasons:

- `u16` values have to be zero-extended on most platforms before being used.
- Dropping the multiply-by-8 might actually give room to an improved instruction selection.
- Some micro architectures (e.g. embedded devices) might not have the same capabilities such as modern x86 or ARM and thus might see considerably better performance.

This issue tracks progress of this potential optimization.

### LINKED ISSUE #1738: Improve alignment and memory layout of Wasmi operators
With https://github.com/wasmi-labs/wasmi/pull/1650 new Wasmi operators have been introduced together with 2 different modes of bytecode layout: `indirect-dispatch` and `!indirect-dispatch`.

- `indirect-dispatch` encodes all operators with a 16-bit op-code.
- `!indirect-dispatch` encodes all operators with a pointer sized function pointer to the execution handler.

Thus, `indirect-dispatch` offsets all operands by 2 bytes whereas `!indirect-dispatch` offsets operands by either 4 bytes (32-bit) or 8 bytes (64-bit).

Ideally, operators are laid out differently depending on those constraints.
For example, `i64.add_ssi` has 3 operators:

1. `result: Slot` (2 bytes)
2. `lhs: Slot` (2 bytes)
3. `rhs: i64` (8 bytes)

This requires different layouts on different configs:

1. `indirect-dispatch`:

```rust
struct I64Add_Ssi {
    pub result: Slot,
    pub lhs: Slot,
    #[cfg(target_pointer_width = "64")]
    _pad: Padding<4>,
    pub rhs: i64,
}
```

Where `Padding<N>` is defined as:
```rust
#[derive(Debug, Default, Copy, Clone)]
pub struct Padding<const N: usize> {
    marker: [u8; N],
}
```
And has a `Decode` and `Encode` impl that simply forwards the decoder or encoder cursor while decoding or encoding nothing. Its bytes are always zeroed and cannot be changed. It simply exist to create padding bytes in between operands.

2. `!indirect-dispatch`

This is a bit more tricky since we either have a 4-byte offset (32-bit) or an 8-byte offset (64-bit).

```rust
struct I64Add_Ssi {
    pub result: Slot,
    pub lhs: Slot,
    #[cfg(target_pointer_width = "64")]
    _pad: Padding<4>,
    pub rhs: i64,
}
```

or we could also re-order fields which is what the Rust compiler would do:

```rust
#[cfg(target_pointer_width = "64")]
struct I64Add_Ssi {
    pub rhs: i64,
    pub result: Slot,
    pub lhs: Slot,
    _pad: Padding<4>, // so the next operator starts at alignment of 4
}

#[cfg(target_pointer_width = "32")]
struct I64Add_Ssi {
    pub result: Slot,
    pub lhs: Slot,
    pub rhs: i64,
}
```

All in all this can get complex very quickly due to the different modes of memory layout.

### LINKED ISSUE #1739: Use accumulator-based interpreter architecture
Currently, Wasmi uses stack slots (a.k.a. `Slot`) to load and store values from the stack when executing operation of a function. Each function on the call stack has its own region of stack slots that it can access.

For example, there is the `i32.add_ssi` operation that has 3 operands:

1. `result: Slot`
2. `lhs: Slot`
3. `rhs: i32`

In this example, `result` and `lhs` are both `Slot` and thus `lhs` is loaded from the stack and `result` is stored into the associated stack slot, whereas `rhs` is an immediate value.

This is more efficient than a stack-based interpreter architecture but we can do better.

A more efficient interpreter architecture is used by both Wasm3 and Makepad's Stitch interpreter. During translation of Wasm to their internal bytecodes they always put the top-most value of the stack into a so-called "register" value. This value can be easily put into its own registers in execution handlers during the interpreter execution loop and thus be access (load/store) way faster than loading and storing from and to stack slots. This whole interpreter architecture is nothing new and those registers are simply called accumulators. For the sake of simplicity it is usually enough to have a single accumulator, thus a single such register value.

Right now, Wasmi's execution handlers already are using 6 integer registers:

- `vmstate`: pointer to the execution context which stores auxiliary data and allows to access the store
- `ip`: the instruction pointer
- `sp`: the stack pointer
- `mem0_ptr`: pointer to the start of `(memory 0)`
- `mem0_len`: length of `(memory 0)`
- `inst`: pointer to the currently used instance

If we add another integer register for the `accumulator` we end up with 7 integer registers which puts further pressure on machine codegen of the compiler and likely results in less efficient code to dispatch to execution handlers. Ideally, we'd remove one of the 6 old register uses or move them under `vmstate`. There are 2 solutions:

1. Move `inst` under `vmstate`. All instance related bytecode will perform worse, likely way worse.
2. Wait until https://github.com/wasmi-labs/wasmi/issues/1492 is merged which allows to remove the `inst` value altogether without requiring a replacement.
3. Turn `inst` into a `f64` field and bit-interpret it back into its actual value when used. This way we no longer use a GPR (general purpose register) for it which is limited to 6. There can be 8 vector/float registers for the SysV calling convention that we use. If this is faster than 1) we should do this instead.

This issue tracks progress and evaluation of this optimization and the research behind it.

Robbepop:
Closes https://github.com/wasmi-labs/wasmi/issues/1739.
Closes https://github.com/wasmi-labs/wasmi/issues/1737.
Closes https://github.com/wasmi-labs/wasmi/issues/1738.

## ToDo: Definition, Translation & Execution

- [x] Deduplicate logic for `if` and `br_if` op-code fusion.
- [x] Remove `#![expect(dead_code)]` and remove unused code in general.
- [x] Wasm `block`, `if` and `loop` with branch params should use accumulator registers.
    - This should reduce the amount of times a value actually needs to be copied for conditional branches.
    - Ideally we'd expand this for any amount of branch parameters and use the maximum available amount of registers. E.g. when branch parameter types are `i32, i64, f32, f64` we'd use all 3 registers and only have to do 1 slot-based copy operation.
- [x] Optimization: unblock op-code fusion in `if` translation.
    - Translating `if` currently performs `preserve_[all_]{locals,regs}` before trying to perform op-code fusion. The calls to `preserve_all_{locals,regs}` might encode copies which flushes the staged operand, thus invalidating any otherwise op-code fusion which was otherwise valid.
- [x] Optimization: unblock op-code fusion in `local.set` translation
    - Op-code fusion is blocked whenever the overwritten local has to be preserved.
    - Steps:
        1. We should evaluate if fusion is possible and if that's the case, drop the staged `Op`.
        2. Then preserve the overwritten local.
        3. Finally, encode the fused `Op`.
        4. If op-code fusion is not possible according to step 1. then follow the current mechanics.
    - Unfortunately with this PR this will only yield better codegen for Wasm `simd` code.
- [x] Optimization: add special `copy_sr` variants with fixed slots.
    - The idea is that due to register preservation the new Wasmi translator emits lots of those `copy_sr` operators.
    - Therefore optimizing this particular operator variant may improve performance significantly.
    - Benchmarks show performance improvements across the board between 5-40%.
    - Fixed slots up to `(slot 9)` are now encoded as variant, e.g. `u64_copy_s9r`.
- [x] Optimization: don't preserve local operands that are parameters to `block` or `if`.
    - This is experimental and I am not sure at this point whether this optimization actually is sound.
- [x] Optimization: don't preserve register operands that are parameters to `block` or `if`.
    - This is experimental and I am not sure at this point whether this optimization actually is sound.
- [x] Optimization: memorize if a register is currently stored in which `local` if any. If that's the case, use the associated register instead of the `local` as operand in an operator translation.
    - This might trigger paths that are currently marked as `unreachable`, e.g. `i32_add_rrr` because with this `lhs = rhs = register` can happen etc..
    - For ideal codegen we should not add `i32_add_rrr` to avoid combinatory explosion but to something like `i32_mul_rri` with `i = 2` instead where it makes sense to do so.
    - This memoization needs to be reset at some code paths, e.g. maybe on control flow edges?
- [x] #1737
- [x] Encoded operators should be aligned via padding if indirect-dispatch` is diabled.
    - While this increases memory usage of Wasmi bytecode, benchmarks showed positive effects.
- [x] Eagerly encode staged operators with the ability to reset the encoding upon unstaging.
    - This should be way more efficient than the current solution since fewer op-code fusions happen with this PR.
    - Translation performance improved by up to 10% with a median at ~3%.
- [x] Preserve all state between `Operand` and `StackOperand`.
    - This is helpful when re-pushing operands back onto the stack, e.g. in `translate_reinterpret`.
- [x] Keep track of `Operand::Reg` operands on the `Stack`
    - This is required in order to copy them to `Operand::Temp` when there are conflicts.
- [x] Optimization: avoid re-flushing an `Operand::Reg` on the stack that has already been set to a local via `local.tee`.
- [x] Turn some `From<RawVal>` where bounds back into `From<TypedRawVal>`.
- [x] Fix operator translation bugs.
    - [x] non-SIMD operators
    - [x] SIMD operators
    - [x] Preserve register state in between resumable calls.
    - [x] Preserve register state across calls.
- [x] Operator Definition, Translation & Execution
    - [x] Unary Ops
    - [x] Binary Ops
        - [x] Div/Rem
        - [x] Copysign
        - [x] Shift
        - [x] All Others
    - [x] Compare-Branch Ops
    - [x] Load Ops
    - [x] Store Ops
    - [x] Simd Ops
        - [x] Splat Ops
        - [x] Select Op
        - [x] Extract Lane Ops
        - [x] Replace Lane Ops
        - [x] Shift Ops
        - [x] Load Ops
        - [x] Store Ops
        - [x] Unary Ops
        - [x] Binary Ops
    - [x] Select Ops
    - [x] Call Ops
    - [x] Conversion Ops
    - [x] Copy Ops
    - [x] Return Ops
    - [x] Table Get/Set
    - [x] Table Bulk Ops
    - [x] Memory Bulk Ops
    - [x] Global Ops
    - [x] Branch Table

## (Failed) Experiments

- [ ] https://github.com/wasmi-labs/wasmi/issues/1853
    - Branch: https://github.com/wasmi-labs/wasmi/tree/rf-add-i32-eqz-nez-ops
    - Conclusion:
        - Benchmarks: do not improve
        - Memory Usage: outstanding
- [ ] Add `SIZE` (or `LEN`) assoc. constant to `wasmi_ir::Encode` trait.
    - This way one can query the fixed encoding size of any type which is important to decouple some code fragments, such as `br_table` execution.
    - Problem: `enum` types should have `size = max(size(variants)..)`, however that computation in `wasmi_ir`'s `build.rs` script may be quite compute intense with over 1000 different operator variants.
- [ ] **Experiment:** Introduce a `{i32,i64,f32,f64}_copy_br_i32_[not_]eq_r{s,r}{s,i}` operator.
    - This is an operator that conditionally branches and also copies a single value if the branch is taken.
    - This can be used for cases where `br_if` conditionally branches to a destination that has exactly a single branch parameter. Today 3 Wasmi IR operators are emitted, with this, only 1 is needed.
    - Control flow structures with a single branch parameter are pretty common since they are enabled since Wasm MVP.
    - This requires `block`, `if` and `loop` to put results also into registers and not just temps.
    - This experiment was conducted on https://github.com/wasmi-labs/wasmi/tree/rf-copy-br-if-operators.  
      Unfortunately it was found not to improve any benchmark to a significant extend. Thus it was not merged.

## Follow-Ups

- [x] https://github.com/wasmi-labs/wasmi/issues/1889
- [x] https://github.com/wasmi-labs/wasmi/issues/1886
- [x] https://github.com/wasmi-labs/wasmi/issues/1880
- [x] https://github.com/wasmi-labs/wasmi/issues/1888
- [x] https://github.com/wasmi-labs/wasmi/issues/1890
- [x] https://github.com/wasmi-labs/wasmi/issues/1703
- [x] https://github.com/wasmi-labs/wasmi/issues/1795
- [x] https://github.com/wasmi-labs/wasmi/issues/1887
- [ ] https://github.com/wasmi-labs/wasmi/issues/1891


Robbepop:
Initial benchmarks show interesting results:
image

While the Rust compiled Wasm blobs such as `tiny_keccak`, `reverse_complement` and `regex_redux` have improved significantly or are on par, some artificial benchmarks such as fibonacci variants perform worse now.

Note that these benchmarks have been conducted with 32-bit `Slot` adjustments which are not included in this PR.

Robbepop:
Some more benchmarks with the current state of this PR:
Wasmi (PR) vs Wasmi (`main`)

- **Green:** Wasmi (PR) is faster
- **Red:** Wasmi (`main`) is faster

image

Robbepop:
Some more benchmarks with the current state of this PR:
Wasmi (PR) vs Wasmi (`main`)

- **Green:** Wasmi (PR) is faster
- **Red:** Wasmi (`main`) is faster

image

Robbepop:
Some more benchmarks with the current state of this PR:
Wasmi (PR) vs Wasmi (`main`)

- **Green:** Wasmi (PR) is faster
- **Red:** Wasmi (`main`) is faster

image

Robbepop:
Some more benchmarks with the current state of this PR:

tl;dr: Some of the performance wins are decisive enough to spark confidence that the remaining regressions will fade once all planned optimizations are in place.

## CoreMark Scores (Apple M2 Pro)

| Runtime  | Version         | Type        | Score |
|:---------|----------------:|:-----------:|------:|
| Wasmtime | `v45`           | JIT         | 30086 |
| Winch    | `v45`           | JIT         | 13598 |
| Wasm3    | `v0.5`          | Interpreter |  2919 |
| Wasmi    | `v2.0.0-beta.3` | Interpreter |  2544 |
| Stitch   | `v0.1`          | Interpreter |  2228 |
| Wasmi    | `v1.0.9`        | Interpreter |  2027 |
| Pulley   | `v45`           | Interpreter |  1786 |
| Tinywasm | `v0.9`          | Interpreter |   937 |

> 1. Wasmi `v2.0.0-beta.3` represents this PR.
> 2. Stitch regressed since Rust 1.92. Before that it had a score of ~3000.

## Wasmi Benchmarks

Wasmi (PR) vs Wasmi (`main`)

- Green: Wasmi (PR) is faster
- Red: Wasmi (`main`) is faster

image

image

PR #1866, opened 2026-05-30
Use `hint::select_unpredictable` in cmp+br and select

Robbepop:
Unfortunately this requires us to bump our MSRV from 1.86 to 1.88.
Althought, I couldn't find any differences in assembly codegen on `aarch64` it could be beneficial on other platforms.
I wanted to open this draft PR as a reference for later when we bump to 1.88 anyways.


Robbepop:
Rust and LLVM already use the same codegen even without the `hint::select_unpredictable` methods.
Therefore, for now we don't need to merge this PR. In the future, we might see more codegen stability from using these methods.
This should certainly be used once Wasmi's MSRV is updated to 1.88+.

PR #1898, opened 2026-06-19
Re-design `CodeMap` to profit from append-only semantics

### LINKED ISSUE #1128: Make `CodeMap` lock-free
Since https://github.com/wasmi-labs/wasmi/pull/1122 Wasmi's internal `CodeMap`, that stores all the data of functions stored in the `Engine`, is put entirely behind a `Mutex` based lock for all accesses.
Before #1122, it was put behind an `RwLock` and was `read`-locked whenever Wasmi exeucted a function which was the root cause for the dead-lock that #1122 has fixed.

The downsides of #1122 is that the performance of call-intense workload have regressed by roughly 5-10% which was acceptable given the importance of fixing the dead-lock.

However, ideally we want a lock-free version of `CodeMap` with hopefully improved performance, especially for accessing pre-compiled Wasm functions. Access of pre-compiled Wasm functions is part of the hot-path of the Wasmi executor and thus its performance is critical.

An inspiring blog post about a lock-free append-only `Vec` implementation in Rust:
https://ibraheem.ca/posts/a-lock-free-vector/

This issue is about tracking progress for the lock-free `CodeMap` implementation.

### LINKED ISSUE #1880: Re-design `CodeMap` and profit from append-only semantics
tl;dr: With this proposed design it should be possible to reach Stitch level performance for direct Wasm `call` and `return_call` operators targeting internal (non-imported) Wasm functions.

## Summary

Every Wasm→Wasm call currently acquires an engine-global `spin::Mutex` to resolve the
callee's bytecode from the `CodeMap`. This issue proposes restructuring the `CodeMap`
into an append-only, pointer-stable bucket structure — still behind a top-level `Mutex`
for its container APIs — with a per-function **atomic state machine** governing each
entity. Because entity addresses are stable for the engine's lifetime, `call_internal`
and `return_call_internal` ops can then embed a **direct pointer** to the callee's
`FuncEntity` instead of an `InternalFunc` index, bypassing the mutex, the index
computation, and the `CodeMap` dereference entirely on the internal-call hot path.

The work splits into two PRs:

- **PR 1** — append-only `CodeMap` + per-entity atomic state (foundational: stable
  addresses, lock-free compilation waiting, fixes a latent concurrency bug)
- **PR 2** — embed `FuncEntity` pointers in internal call ops (IR + translator +
  executor; this is where the per-call speedup lands)

## Motivation

The call hot path today (branch: `rf-accumulator-based-interpreter-arch`):

1. `call_internal` handler decodes an `InternalFunc` (u32 index) and calls `call_wasm`
   (`crates/wasmi/src/engine/executor/handler/exec.rs:281-297`).
2. `call_wasm` → `compile_or_get_func!` → `CodeMap::get`
   (`crates/wasmi/src/engine/executor/handler/utils.rs:66-72`, `:657-670`).
3. `CodeMap::get` → `get_compiled` locks `self.funcs: Mutex<Arena<EngineFunc, FuncEntity>>`,
   indexes the arena, matches on the `FuncEntity` enum, and transmutes the resulting
   reference's lifetime to escape the `MutexGuard`
   (`crates/wasmi/src/engine/code_map.rs:74-77`, `:320-333`, `:359-364`).

Costs of this design:

- **Two atomic RMWs per call** (lock + unlock) on a *single engine-global cache line*,
  even in single-threaded execution where the lock is never contended.
- **Cross-thread serialization**: all threads executing on the same `Engine` — even on
  disjoint modules/functions — bounce the same mutex cache line on every call.
- **Waiting for lazy compilation hammers the global lock**: `wait_for_compilation`
  spins by repeatedly locking the mutex (`code_map.rs:413-434`), contending with every
  call on every other thread.
- The escape-the-guard lifetime transmute (`adjust_cref_lifetime`) is already justified
  by an informal "append-only" argument — the design wants pointer stability; this
  proposal makes that property structural.

The mutex is genuinely needed only for the *container*: allocating entries during
module build and resolving indices to entities. The *per-entity* lifecycle
(uncompiled → compiling → compiled/failed) can be governed by an atomic per entity,
which is also exactly what makes direct entity pointers usable without any lock.

## Current design recap

```rust
pub struct CodeMap {
    funcs: Mutex<Arena<EngineFunc, FuncEntity>>,   // spin::Mutex
    features: WasmFeatures,
}

enum FuncEntity {
    Uninit,
    Uncompiled(UncompiledFuncEntity),
    Compiling,
    FailedToCompile,
    Compiled(CompiledFuncEntity),
}
```

- `alloc_funcs` appends `Uninit` entries during module build
  (`crates/wasmi/src/module/builder.rs:239`); they are then initialized as
  `Compiled` (eager) or `Uncompiled` (lazy) before the module is shared.
- Lazy compilation: the executing thread takes the `UncompiledFuncEntity` *out*
  (state → `Compiling`), compiles **without holding the lock**, then re-locks to
  store the result.
- Out-of-fuel during lazy compilation reverts `Compiling → Uncompiled` so a resumable
  call can retry later (`code_map.rs:392-395`).

### Latent bug in the current design

If thread A reverts `Compiling → Uncompiled` on out-of-fuel while thread B (a different
store with sufficient fuel) is sitting in `wait_for_compilation`, B observes
`Uncompiled` and **panics** ("unexpected function state", `code_map.rs:429-431`)
instead of retrying compilation itself. The proposed state machine fixes this
naturally: waiters loop back into the full dispatch and race to claim compilation.

### Part A — append-only, pointer-stable function storage (mutex retained)

Replace the `Arena` (a growable `Vec`, which reallocates and therefore moves entities)
with doubling-capacity buckets of boxed slices, kept behind a top-level `Mutex` as
today:

```rust
pub struct CodeMap {
    funcs: Mutex<Funcs>,
    features: WasmFeatures,
}

struct Funcs {
    /// Bucket `b` has capacity `2^(B0 + b)`. Buckets are allocated on demand,
    /// never reallocated, never freed until the `CodeMap` drops.
    buckets: Vec<Box<[FuncEntity]>>,
    /// Number of allocated (not necessarily initialized) entities.
    len: u32,
}

/// log2 of the first bucket's capacity (64 entries).
const B0: u32 = 6;
```

Keeping the top-level `Mutex` (instead of lock-free `AtomicPtr` bucket publication)
deliberately trades a little hot-path generality for a lot of API simplicity: growth,
index resolution, and bookkeeping stay ordinary safe-ish code, and no
publication/`Acquire`-load dance is needed on the bucket table. The outer `Vec` may
reallocate its pointer table when buckets are appended — that is fine, because the
**entities themselves live in the boxed slices, whose allocations never move**.
`&FuncEntity` / `NonNull<FuncEntity>` obtained once therefore remain valid for the
engine's lifetime. Likewise, since the structure is append-only, an in-bounds index
stays in-bounds forever.

Index → (bucket, slot) is branch-free:

```rust
#[inline]
fn locate(index: u32) -> (usize, usize) {
    let i = u64::from(index) + (1 << B0);
    let msb = 63 - i.leading_zeros();      // floor(log2(i))
    let bucket = (msb - B0) as usize;
    let slot = (i - (1u64 << msb)) as usize;
    (bucket, slot)
}
```

- Bucket 0 holds indices `0..64`, bucket 1 holds `64..192` (128 entries), bucket `b`
  holds `2^(B0+b)` entries — the same doubling growth profile (and worst-case slack)
  as the current `Vec`-backed arena. 27 buckets cover the entire `u32` index space.
- **`alloc_funcs`** (cold, module build): lock, allocate any missing buckets as
  `Box<[FuncEntity]>` with entries default-initialized to the `UNINIT` state, bump
  `len`, return the `EngineFuncSpan`.
- **Index-based access** (`get`, `init_func_as_*`, `entity_ptr`): lock, `locate`,
  take a `NonNull<FuncEntity>` / `&FuncEntity`, unlock. Everything after address
  resolution runs on the per-entity atomic protocol (Part B), *outside* the lock.
- **`Drop`**: automatic — `Vec<Box<[FuncEntity]>>` drops normally; per-entity cleanup
  is handled by `FuncEntity`'s own `Drop` (Part B).

### Part B — per-entity atomic state machine over a union

The `AtomicU8` becomes the **sole discriminant**; the payload is a union rather than
an enum, so no second discriminant is stored:

```rust
pub struct FuncEntity {
    /// Synchronization authority *and* discriminant. One of the consts below.
    state: AtomicU8,
    /// Payload; which field is active (if any) is determined by `state`.
    data: UnsafeCell<FuncEntityData>,
}

union FuncEntityData {
    /// Active in `UNINIT`, `COMPILING` and `FAILED` states.
    uninit: (),
    /// Active in the `UNCOMPILED` state.
    uncompiled: ManuallyDrop<UncompiledFuncEntity>,
    /// Active in the `COMPILED` state.
    compiled: ManuallyDrop<CompiledFuncEntity>,
}

// Safety: see "Safety" section.
unsafe impl Send for FuncEntity {}
unsafe impl Sync for FuncEntity {}

mod state {
    pub const UNINIT: u8 = 0;
    pub const UNCOMPILED: u8 = 1;
    pub const COMPILING: u8 = 2;
    pub const FAILED: u8 = 3;
    pub const COMPILED: u8 = 4;
}
```

Two details the union forces (vs. an enum), both small:

- Union fields holding non-`Copy` types must be wrapped in `ManuallyDrop`, and an
  `uninit: ()` field is needed so bucket entries can be default-constructed.
- `FuncEntity` needs a manual `Drop` impl: read `state` (exclusive access at drop
  time, `Relaxed`/`get_mut` suffices) and `ManuallyDrop::drop` the active field.
  `UNINIT`, `COMPILING` and `FAILED` drop nothing — which is exactly right, since in
  `COMPILING`/`FAILED` the payload was moved out by the compiling thread (a panic
  mid-compilation leaves `COMPILING` behind and the moved-out data is dropped by the
  unwinding compile call, so this is panic-safe by construction).

Note that the `UnsafeCell` is **still required despite the top-level mutex**: the
lazy-compilation winner writes the compiled result through a *shared* `&FuncEntity`
(possibly reached via a raw pointer embedded in bytecode, with no lock held), and
readers access the payload through shared references concurrently with `state` loads.
Relatedly, after an entity has been created, the code must **never form
`&mut FuncEntity`** (e.g. through the `MutexGuard`): a `&mut` would assert exclusive
access over the whole entity — including `state` — while other threads concurrently
perform atomic loads on it, which is undefined behavior under Rust's aliasing rules.
All mutation goes through the `UnsafeCell`, gated by the protocol below.

State machine (replaces `CodeMap::compile_or_wait` / `wait_for_compilation`):

```text
FuncEntity::get_or_compile(fuel, features):     // takes &self — no lock held
  loop on state.load(Acquire):
    COMPILED   → read `data.compiled` (immutable since publication)          [fast path]
    FAILED     → Err(LazyCompilationFailed)                                  [sticky]
    UNCOMPILED → compare_exchange(UNCOMPILED → COMPILING, AcqRel, Acquire)
                  ├─ winner: exit loop, proceed below
                  └─ loser:  re-loop
    COMPILING  → core::hint::spin_loop(); re-loop                            [wait]
    UNINIT     → unreachable_unchecked (entities are initialized before
                                        they become reachable from executing code)

winner (exclusive payload access while state == COMPILING):
  uncompiled = ManuallyDrop::take(&mut (*data.get()).uncompiled)   // move out
  result = uncompiled.compile(fuel, features)                      // NO lock held, as today
  match result:
    Ok(compiled)   → write `data.compiled`;   state.store(COMPILED,   Release); Ok
    Err(OutOfFuel) → write `data.uncompiled`; state.store(UNCOMPILED, Release); Err
    Err(other)     →                          state.store(FAILED,     Release); Err
```

- The `UNCOMPILED` re-loop is what fixes the latent waiter-panic bug: a waiter that
  observes the out-of-fuel revert simply races to claim compilation itself.
- Waiting no longer touches the global mutex at all — waiters spin on the entity's own
  cache line. Spinning matches current behavior (the existing code spin-locks a
  `spin::Mutex` in a loop) and keeps `no_std` compatibility; under the `std` feature
  this could later become `yield_now`/futex parking without changing the protocol.
- `init_func_as_compiled` / `init_func_as_uncompiled`: resolve the entity under the
  mutex, write the payload, then `state.store(…, Release)` (with a debug-mode
  `compare_exchange(UNINIT → …)` to preserve today's double-init assertion).
  Exclusivity holds by construction — module build, span not yet shared.

**Resulting fast path** (with a resolved `&FuncEntity`): 1 `Acquire` load (a plain
load on x86, `ldarb` on AArch64) + one well-predicted branch + 2 loads (`ops` pointer,
`len_stack_slots`). No RMW, no stores, no shared-line contention.

**Honest trade-off of keeping the mutex**: *index-based* lookups still lock briefly to
resolve index → address. So in PR 1, imported calls, indirect calls, and root calls
keep roughly today's locking cost (the lock's critical section shrinks to a pointer
computation, and compilation/waiting move off the lock — but the lock+unlock RMWs
remain). The per-call win arrives with PR 2 for internal calls, and via the
`WasmFuncEntity` pointer-caching follow-up (below) for the remaining call kinds.

### Part C (PR 2) — direct `FuncEntity` pointers in internal call ops

- With stable entity addresses, `call_internal` / `return_call_internal` no longer need
an index — or the mutex — at all:
- With mutable operator sequences, we can introduce new `compile` and `return_compile` operators that primarily compile their underlying known Wasm function and upon finalization replacing themselves with `call_internal` or `return_call_internal` respectively.

**IR (`wasmi_ir`)** — the crate is engine-agnostic, so the field is an opaque
pointer-sized newtype:

- Add `FieldTy::EngineFuncRef` to `crates/ir/build/ty.rs` and switch the `Func` field of
  `CallInternal` (`crates/ir/build/isa.rs:600-606`) and `ReturnCallInternal`
  (`crates/ir/build/isa.rs:618-624`) from `FieldTy::InternalFunc` to it.
- Define `EngineFuncRef(usize)` (opaque, `Debug` prints hex) next to the index types in
  `crates/ir/src/index.rs`; encode/decode as `usize` — already a supported fixed-width
  primitive (`crates/ir/src/encode.rs:137-139`).

**Translator (`wasmi`)**:

- `translate_call` / `translate_return_call`
  (`crates/wasmi/src/engine/translator/func/mod.rs:1649-1673`,
  `crates/wasmi/src/engine/translator/func/visit.rs:352`, `:1699`) currently map
  `FuncIdx → EngineFunc` via `self.module.get_engine_func(func_idx)`. Add
  `CodeMap::entity_ptr(&self, EngineFunc) -> NonNull<FuncEntity>` (lock, locate,
  unlock) and emit the resolved pointer as `EngineFuncRef`.
- All of a module's entities are allocated before any of its functions are translated
  (`crates/wasmi/src/module/builder.rs:239`), eagerly *and* lazily, so the pointer
  always exists at translation time — including self-calls and forward calls.
- The translator already reaches the engine for lazy compilation via
  `ModuleHeader::engine().upgrade()`; cache the `CodeMap` handle in `FuncTranslator` to
  avoid a per-call-op `Weak` upgrade.

**Executor**:

- `call_internal` / `return_call_internal` handlers
  (`crates/wasmi/src/engine/executor/handler/exec.rs:281-297`, `:368-385`) decode
  `EngineFuncRef`, convert to `NonNull<FuncEntity>`, and call
  `FuncEntity::get_or_compile(fuel, features)` directly — no mutex, no `locate`, no
  `EngineFunc` conversion. (`VmState` already carries `code: &CodeMap` for `features`
  access and for the remaining index-based paths.)
- Imported / indirect / root calls keep the index-based `CodeMap::get`
  (lock → resolve address → unlock → atomic protocol).

## Safety

**`Sync`/`Send` for `FuncEntity`** — justified by the protocol:

- The payload is written only (a) during module build before the entity is reachable
  by any other thread (publication happens-before via the `Module` handoff plus the
  `Release` state stores), and (b) by the unique winner of the
  `UNCOMPILED → COMPILING` CAS, which holds logical exclusivity until its `Release`
  store transitions away from `COMPILING`.
- The payload is read-shared only in the `COMPILED` state, after which it is immutable.
- All readers use `Acquire` loads of `state`, pairing with the writers' `Release`
  stores to establish happens-before on the payload.
- Aliasing discipline: after creation, no `&mut FuncEntity` is ever formed; all payload
  access goes through the `UnsafeCell`, and which union field is active is dictated by
  `state` alone.
- Requires `UncompiledFuncEntity: Send` (moved to whichever thread wins compilation)
  and `CompiledFuncEntity: Sync` (shared after publication) — both already hold today:
  entities live behind a `spin::Mutex` (whose `Sync` requires `Send`), and
  `CompiledFuncRef`s already cross the guard boundary via the lifetime transmute.

**Pointer stability / lifetime** — entities live in `Box<[FuncEntity]>` buckets that
are never reallocated or freed until the `CodeMap` drops (the outer `Vec` may move its
*pointer table*, never the entities); the structure is append-only, so an in-bounds
index or resolved pointer stays valid for the engine's lifetime. The bytecode embedding
the pointers (PR 2) is owned by the *same* `CodeMap`, so encoded pointers cannot
outlive their targets; the `Engine` (and thus the `CodeMap`) is kept alive by every
`Store` during execution.

**Provenance** — encoding a pointer into the byte stream and re-materializing it on
decode should use `ptr::expose_provenance` / `ptr::with_exposed_provenance` to stay
well-defined under strict provenance. Precedent: the default direct-dispatch mode
already embeds raw op-handler *function pointers* into the encoded bytecode, so
per-process, non-relocatable bytecode is an established property of Wasmi IR, not a new
constraint introduced here.

## Trade-offs and alternatives considered

- **Top-level `Mutex` vs. fully lock-free container**: a lock-free variant
  (`AtomicPtr` bucket publication, `Acquire` bucket loads) would make *index-based*
  reads lock-free too, but complicates every container API. Since the hot internal
  calls use direct pointers anyway, the mutex is kept and index paths accept a brief
  lock. If imported/indirect call locking ever shows up in profiles, the answer is the
  `WasmFuncEntity` pointer-caching follow-up rather than a lock-free container.
- **Union vs. enum-with-atomic-mirror**: an alternative keeps today's `FuncEntity`
  enum inside the cell with the atomic mirroring its discriminant (automatic `Drop`,
  less unsafe). The union avoids storing the discriminant twice and makes the atomic
  the single source of truth; the cost is a small manual `Drop` and `ManuallyDrop`
  plumbing. The union is the chosen design.
- **Op size**: `EngineFuncRef` is 8 bytes on 64-bit targets vs 4 bytes for
  `InternalFunc` (+4 bytes per internal call op; unchanged on 32-bit). Given that
  direct-dispatch mode already pays `usize` per op for handler pointers, this is in
  line with the existing size/speed stance. Whether `indirect-dispatch` (the
  compactness-focused configuration, `slot16`) should instead keep index-based call ops
  is left as an open question — doing so would double the call-op surface, so the
  default proposal is to embed pointers unconditionally.
- **Call-site self-patching** (`CallInternal` → rewrite to a resolved form after first
  execution) was rejected: it requires mutable bytecode (`UnsafeCell` over all op
  buffers plus racy patching between concurrent executors) for little gain over the
  translate-time pointer, which is available because entities are pre-allocated.
- **Per-`alloc_funcs` segments** (one bucket per module) would make index math require
  a search/offset table; power-of-two buckets keep O(1) `locate`.

## Implementation plan

**PR 1 — append-only storage + per-entity atomic state**
(`crates/wasmi/src/engine/code_map.rs`):

1. Introduce `Funcs` (doubling buckets of `Box<[FuncEntity]>`), `locate`, growth path.
2. Rework `FuncEntity` into `{ AtomicU8, UnsafeCell<union> }` with the state machine
   above (incl. manual `Drop`); port `init_func_as_compiled/_uncompiled` and `get`;
   delete `compile_or_wait`/`wait_for_compilation`/`adjust_cref_lifetime`.
3. Keep `CodeMap::get(fuel, EngineFunc) -> Result<CompiledFuncRef, Error>` signature
   stable so the executor (`handler/utils.rs:66-72`, `handler/func.rs:165`) is
   untouched.
4. Add a multi-threaded regression test: N threads lazily executing the same module's
   functions, including a fuel-limited store that triggers the out-of-fuel revert while
   other threads wait (reproduces the current waiter panic).

**PR 2 — pointer-embedded internal calls**:

1. `wasmi_ir`: `EngineFuncRef` type + `FieldTy` variant + ISA field swap
   (`crates/ir/build/isa.rs`, `crates/ir/build/ty.rs`, `crates/ir/src/index.rs`,
   encode/decode).
2. `CodeMap::entity_ptr`, translator resolution in `translate_call` /
   `translate_return_call`.
3. Executor handlers decode the pointer and call `FuncEntity::get_or_compile` directly.
4. Follow-up candidates (separate issues): cache `NonNull<FuncEntity>` in
   `WasmFuncEntity` so imported/indirect/root calls also bypass the mutex and index
   resolution; store the `EngineFunc` index inside `FuncEntity` for diagnostics.

## Verification

- `cargo test -p wasmi` and `cargo test -p wasmi_wast` (full spec suite).
- Call-heavy benchmarks before/after each PR:
  `cargo bench --bench benches -p wasmi execute/fibonacci` (recursive, dominated by
  `call_internal`), plus the other `execute/*` call benches; the headline per-call
  gains are expected from PR 2.
- The new multi-threaded lazy-compilation + fuel test from PR 1.
- If feasible, run the `code_map` unit tests under Miri
  (`cargo +nightly miri test -p wasmi code_map`) to validate the `UnsafeCell`/union
  protocol and pointer provenance.

## Open questions

- Should `indirect-dispatch` builds keep index-based internal call ops to preserve
  bytecode compactness, at the cost of a second op flavor?
- `B0` tuning (first-bucket capacity 16 vs 64) — affects slack for many-small-module
  embeddings; 64 is proposed as the default.
- Should `WasmFuncEntity` pointer caching (making imported/indirect calls mutex-free
  as well) be folded into PR 2 or kept as a follow-up?

Robbepop:
Closes https://github.com/wasmi-labs/wasmi/issues/1880
Closes https://github.com/wasmi-labs/wasmi/issues/1128

This implements the new append-only `CodeMap`, the optimization for `[return_]call_internal` operators to use references to `FuncEntity` in their operands and avoidance of `Mutex` locking in Wasmi's executor.

This does _not_ implement the self-modifying Wasmi bytecode which introduces `[return_]compile` operators as this was deemed unnecessarily complex and potentially risky concerning UB.

- [x] Implement new `CodeMap`
    - New `CodeMap` is ~10-15% slower than old one for look-ups via `EngineFunc`.
- [x] Use `CodeView` instead of `CodeMap` in Wasmi's executor.
    - This way the costly `Mutex` lock can be avoided for calls.
- [x] Make `CodeMap` completely lock-free.
- [x] Make `[return_]call_internal` operators use `&FuncEntity` operands instead of `EngineFunc`.
    - This way these operators profit fully from the append-only semantics.
- [ ] ~~**Experimental:** Feed `&FuncEntity` into `Store`'s function definitions once known to be compiled.~~
    - ~~This way we could also speed-up calls to imported Wasm functions from other Wasm module instances.~~


Robbepop:
Initial benchmarks show significant improvements for call-intense workloads after having implemented:

Make `[return_]call_internal` operators use `&FuncEntity` operands instead of `EngineFunc`.

## Benchmarks

Performance of call-intense workload improve significantly.
Host-call performance drops only slightly due to the synchronization overhead of `CodeView::refresh`.

image

image

PR #1906, opened 2026-06-22
Automatically create release artifacts on `tag` pushes via CI

Robbepop:
Closes https://github.com/wasmi-labs/wasmi/issues/1828.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,254 @@
+name: Release - Build CLI Artifacts
+
+# Builds the `wasmi` CLI for every supported platform and feature set whenever a
+# semver release tag (`vX.Y.Z` or `vX.Y.Z-suffix`) is pushed, then attaches the
+# archives to a *draft* GitHub Release for review before publishing.
+#
+# It can also be run manually (`workflow_dispatch`) to (re)build the artifacts for
+# an existing tag, e.g. when a tagged release run failed.
+#
+# Safety / least-privilege notes:
+#   * No personal access tokens or secrets are used — only the ephemeral,
+#     repo-scoped `GITHUB_TOKEN` provided automatically by Actions.
+#   * The default permission for the whole workflow is `contents: read`. Only the
+#     small `release` job (which compiles nothing) is granted `contents: write`,
+#     so third-party build dependencies never see a write-capable token.
+#   * The Release is created with the built-in `gh` CLI, so the write token is
+#     never handed to a third-party action.
+#   * All third-party actions are pinned by full commit SHA.
+
+on:
+  push:
+    tags:
+      # Coarse glob (GitHub tag filters are not full regex); the `setup` job below
+      # validates the tag against a strict semver regex.
+      - 'v[0-9]*.[0-9]*.[0-9]*'
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: 'Existing git tag to (re)build & attach artifacts to, e.g. v2.0.0'
+        required: true
+
+# Deny-by-default: read-only for every job unless a job opts into more.
+permissions:
+  contents: read
+
+concurrency:
+  group: release-${{ github.event.inputs.tag || github.ref_name }}
+  cancel-in-progress: false
+
+jobs:
+  # ---------------------------------------------------------------------------
+  # Resolve & validate the release tag once, share it with the other jobs.
+  # ---------------------------------------------------------------------------
+  setup:
+    name: Resolve release tag
+    runs-on: ubuntu-latest
+    outputs:
+      tag: ${{ steps.resolve.outputs.tag }}
+      prerelease: ${{ steps.resolve.outputs.prerelease }}
+    steps:
+      - name: Resolve and validate tag
+        id: resolve
+        shell: bash
+        run: |
+          set -euo pipefail
+          if [ "${{ github.event_name }}" = "workflow_dispatch" ]; then
+            tag="${{ github.event.inputs.tag }}"
+          else
+            tag="${{ github.ref_name }}"
+          fi
+          if ! printf '%s' "$tag" | grep -Eq '^v[0-9]+\.[0-9]+\.[0-9]+(-.+)?$'; then
+            echo "::error::Tag '$tag' is not a valid release tag (expected vX.Y.Z or vX.Y.Z-suffix)"
+            exit 1
+          fi
+          # Mark tags with a suffix (e.g. -beta.2, -rc.1) as prereleases.
+          if printf '%s' "$tag" | grep -q '-'; then
+            prerelease=true
+          else
+            prerelease=false
+          fi
+          echo "tag=$tag" >> "$GITHUB_OUTPUT"
+          echo "prerelease=$prerelease" >> "$GITHUB_OUTPUT"
+          echo "Resolved release tag: $tag (prerelease=$prerelease)"
+
+  # ---------------------------------------------------------------------------
+  # Build + package the CLI for each platform × feature set (read-only token).
+  # ---------------------------------------------------------------------------
+  build:
+    name: ${{ matrix.platform.token }} (${{ matrix.featureset.name }})
+    needs: setup
+    runs-on: ${{ matrix.platform.runner }}
+    env:
+      # Optimize the release builds without editing the (commented-out) profile in
+      # crates/cli/Cargo.toml. The workspace already sets `codegen-units = 1`.
+      CARGO_PROFILE_RELEASE_LTO: fat
+      CARGO_PROFILE_RELEASE_STRIP: symbols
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,254 @@
...
+          $ErrorActionPreference = "Stop"
+          $name = "wasmi-${{ needs.setup.outputs.tag }}-${{ matrix.platform.token }}-${{ matrix.featureset.name }}"
+          $staging = "dist/$name"
+          New-Item -ItemType Directory -Force -Path $staging | Out-Null
+          Copy-Item "target/${{ matrix.platform.target }}/release/wasmi.exe" "$staging/"
+          Copy-Item "LICENSE-APACHE","LICENSE-MIT" $staging/
+          Copy-Item "crates/cli/README.md" "$staging/README.md"
+          Compress-Archive -Path $staging -DestinationPath "dist/$name.zip"
+          Remove-Item -Recurse -Force $staging
+          "ARTIFACT=dist/$name.zip" | Out-File -FilePath $env:GITHUB_ENV -Append -Encoding utf8
```
polarathene, line 199: [`cargo-dist`](https://github.com/axodotdev/cargo-dist) or similar would have simplified this? (_handles various parts including [archiving for different platforms with `.tar.gz`/`.zip`](https://axodotdev.github.io/cargo-dist/book/artifacts/archives.html#archive-formats)_) [Here's an reference config](https://github.com/rust-cross/cargo-zigbuild/blob/main/dist-workspace.toml) used by `cargo-zigbuild`.

**FWIW:** `just` [defines their matrix](https://github.com/casey/just/blob/68d95b73462694efe974132f16f448f34943cb42/.github/workflows/release.yaml#L32-L75) and has a [`package` step](https://github.com/casey/just/blob/68d95b73462694efe974132f16f448f34943cb42/.github/workflows/release.yaml#L116-L124) that [runs a bash script](https://github.com/casey/just/blob/master/bin/package) (_on all OS runners Linux, macOS, and Windows_) followed by a [separate step to publish artifacts to GH Releases](https://github.com/casey/just/blob/68d95b73462694efe974132f16f448f34943cb42/.github/workflows/release.yaml#L126-L134). They then upload the artifact for other jobs (_adding some follow-up changes to the GH release after the matrix job completes + building/publishing container images_).

.github/workflows/release.yml
```diff
@@ -0,0 +1,254 @@
...
+            features: run,wat,wast,wasi,validate,stable,simd
+          # `minimal` == smallest useful CLI: run validated modules only.
+          - name: minimal
+            features: run,validate
+        # Exactly the subset of Wasmtime's target-triples that matches our
+        # {x86_64, aarch64} × {Linux, macOS, Windows} scope. All native runners.
+        platform:
+          - token: x86_64-linux
+            target: x86_64-unknown-linux-gnu
+            runner: ubuntu-22.04
```
polarathene, line 103: Is the decision to use an old Ubuntu here due to glibc?

If you use [`cargo-zigbuild`](https://github.com/rust-cross/cargo-zigbuild) to build via Zig instead of GCC, you can quite easily cross-compile to different archs, libc ABI and even platforms 😎 (_without the overhead of QEMU or need to run native runners_)

```bash
# Glibc 2.34 (Aug 2021) is generally a good lower bound:
cargo zigbuild --target aarch64-unknown-linux-gnu.2.34
```

The [`cargo-zigbuild` project also provides a Docker image with the macOS SDK](https://github.com/rust-cross/cargo-zigbuild/blob/main/README.md#installation), so you should be able to build for all targets from a single runner if you'd like a simpler setup:

```bash
docker run --rm -it --volume ./:/io --workdir /io ghcr.io/rust-cross/cargo-zigbuild \
  cargo zigbuild --release --target x86_64-apple-darwin
```

Although, besides the benefit of not having to use older distros for glibc linking I guess using multiple native runners works too for faster CI runs if the builds take time 🤔 

---

For context that is a 4.9GB container image with all the targets setup. Presently packaged with Rust `1.93.0`, glibc (2.44), and using Debian 13 (Trixie) as the distro:

```console
$ docker --version
Docker version 28.0.1, build 068a01e

$ docker run --rm -it ghcr.io/rust-cross/cargo-zigbuild bash

$ du -shx /
4.9G    /

$ grep PRETTY_NAME /etc/os-release
PRETTY_NAME="Debian GNU/Linux 13 (trixie)"

$ ld --version
GNU ld (GNU Binutils for Debian) 2.44
Copyright (C) 2025 Free Software Foundation, Inc.

$ cargo --version
cargo 1.93.0 (083ac5135 2025-12-15)

$ cargo-zigbuild --version
cargo-zigbuild 0.23.0

$ zig version
0.16.0

$ gcc --version
gcc (Debian 14.2.0-19) 14.2.0

# LLVM 19:
clang --version
Debian clang version 19.1.7 (3+b1)

# Size breakdown (largest contributors, approx 4.2GB of weight):
$ du -shx /opt/MacOSX11.3.sdk /usr/local/{zig-x86_64-linux-0.16.0,rustup/toolchains} /usr/lib/{llvm-19,x86_64-linux-gnu}
597M    /opt/MacOSX11.3.sdk
396M    /usr/local/zig-x86_64-linux-0.16.0
1.8G    /usr/local/rustup/toolchains
797M    /usr/lib/llvm-19
650M    /usr/lib/x86_64-linux-gnu

# Target support:
$ rustup show
Default host: x86_64-unknown-linux-gnu
rustup home:  /usr/local/rustup

installed toolchains
--------------------
1.93.0-x86_64-unknown-linux-gnu (active, default)

active toolchain
----------------
name: 1.93.0-x86_64-unknown-linux-gnu
active because: it's the default toolchain
installed targets:
  aarch64-apple-darwin
  aarch64-pc-windows-gnullvm
  aarch64-unknown-linux-gnu
  aarch64-unknown-linux-musl
  arm-unknown-linux-gnueabihf
  arm-unknown-linux-musleabihf
  x86_64-apple-darwin
  x86_64-pc-windows-gnu
  x86_64-unknown-linux-gnu
  x86_64-unknown-linux-musl
```

Since the build environment is linux, the windows targets differ (_`-gnu`/`-gnullvm` instead of `-msvc`_).

polarathene (commented):
You've probably got higher priorities and if you're happy with your workflow setup then never mind me :)

Just sharing some info if it's helpful 😁

Robbepop:
> You've probably got higher priorities and if you're happy with your workflow setup then never mind me :)
> 
> Just sharing some info if it's helpful 😁

Hi @polarathene and thanks a lot for your comments and suggestions!

First, I am not very experienced with anything related to CI infrastructure, so please take everything I say with a considerable grain of salt.

The reason I did not use `cargo-zigbuild` and `cargo-dist` (or just `dist`) is because I wanted to have a lean and simple setup that is solid and robust enough to not break constantly due to 3rd party tool changes or breakages etc..

So my goals were to keep 3rd party tool usage and dependencies very limited.
The reason an old `ubuntu-22` is used is, as you correctly pointed out, because of lower-floor `glibc`.
Similar to Wasmtime, we build static self-contained `musl` builds, too.

With the current setup we are also able to quickly smoke-test the built binaries (simple `--help`) before releasing them. That's not possible if we used a single cross-compiling runner.

I just didn't feel like I get anything I really need for my minimal setup from `cargo-zigbuild` or `cargo-dist` in all honesty. Maybe, once the setup is more complex or so it will be useful for me. But until then, I'd like to keep things simple.

polarathene:
> With the current setup we are also able to quickly smoke-test the built binaries (simple `--help`) before releasing them. That's not possible if we used a single cross-compiling runner.

Actually with QEMU that should work for testing all archs in one runner for Linux, whereas for Windows WINE enables that too. For macOS I think you'd have to do that on the native runner.

> I just didn't feel like I get anything I really need for my minimal setup from `cargo-zigbuild` or `cargo-dist` in all honesty. Maybe, once the setup is more complex or so it will be useful for me. But until then, I'd like to keep things simple.

Well there's the glibc concern tackled to enable you to use newer build environments :)

The version of musl linked I think in this case isn't tied to the older distro package, since Rust bundles it's own for static linking, you'd only need to reach out to `cargo-zigbuild` for glibc host to musl target builds when the target is a foreign arch.

If GH provides all runners natively for what you want to support with wasmi then there's no issue 👍 (_I assume at some point there might be requests for riscv64, which is where it might be useful but even then you could approach it without tools like `cargo-zigbuild` if that were an issue_).

> The reason I did not use `cargo-zigbuild` and `cargo-dist` (or just `dist`) is because I wanted to have a lean and simple setup that is solid and robust enough to not break constantly due to 3rd party tool changes or breakages etc..

Perfectly valid reason, as long as you're comfortable maintaining what you have that's totally fine 😁

PR #1927, opened 2026-06-26
Make accumulator registers (`ireg` etc..) caller saved upon calls

### LINKED ISSUE #1925: Make `ireg`, `freg32` and `freg64` caller-saved upon calls
Currently, when taking a call, Wasmi preserves all its registers to the stack.

Given that https://github.com/wasmi-labs/wasmi/issues/1888 was an unsuccessful optimization,
function calls are never populating the caller's registers.

Therefore, Wasmi has no caller-saved nor callee-saved registers. Registers just always vanish upon calls.
That's a potential performance problem for call-intense executions.

What Stitch (and potentially Wasm3) do, is: they save their register state in call frame and restore those states after returning back to the caller. Thus, they implement caller-saved semantics for those accumulator registers.

Wasmi should try to do the same and benchmark if that has positive effects on performance.

The obvious downside to this is that call and return operator executions will be impacted (due to register store/restore and function frame sizes on the call stack will have to store 160 (64 + 64 + 32) additional bits.

Robbepop:
Closes https://github.com/wasmi-labs/wasmi/issues/1925.

### DIFF
--- a/crates/wasmi/src/engine/executor/handler/exec.rs
+++ b/crates/wasmi/src/engine/executor/handler/exec.rs
@@ -414,7 +414,7 @@ execution_handler! {
         let (_, crate::ir::decode::ReturnCallImported { params, func }) = unsafe { decode_op(ip) };
         let func = fetch_func(instance, func);
         let func_entity = resolve_func(state.store, &func);
-        let (callee_ip, sp, new_instance) = match func_entity {
+        let (callee_ip, sp, new_instance, ireg, freg32, freg64) = match func_entity {
             FuncEntity::Wasm(func) => {
                 let wasm_func = func.func_body();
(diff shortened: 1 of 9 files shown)


Robbepop:
Unfortunately benchmarks show that this optimization causes more regressions than it helps with performance:

image

The reason being that the added costs of having to carry around `ireg`, `freg32` and `freg64` caller-saved accumulator registers in each `Frame` on Wasmi's call stack is unconditional whereas the old Wasmi only preserved live accumulator registers, thus was conditional.

Furthermore, I thought Stitch was translating and executing Wasm function calls exactly as this PR was implemented, however, upon further inspection, I noticed that Wasmi as is on `main` right now haves exactly how Stitch does.

In conclusion: won't merge. Closed.

PR #1934, opened 2026-06-27
Add `memory64` crate feature to Wasmi

### LINKED ISSUE #1931: Allow to disable Wasm `memory64` support via crate feature
Wasmi currently and unconditionally supports Wasm `memory64`.
Just like most other Wasm proposals it can be enabled or disabled via Wasmi's `Config` API.

However, enabling Wasm `memory64` has real consequences for execution performance in that Wasmi has to do an additional bounds check for each executed load and store operation. As load and store operations are quite common operations this might add up quickly to measurable performance degradation.

If it was possible to statically disable Wasm `memory64` support at compile-time, Wasmi could improve its codegen for load and store operators without having to introduce too many new operator variants into its IR.

This way, users who don't need support for Wasm `memory64` could simply disable it and have all the performance advantages of other Wasm interpreters that don't support this proposal such as Stitch and Wasm3.

This issue tracks progress and evaluation of this feature.

## Assembly

The assembly for Wasmi's `u64_load_mem0_offset16_rr` operator is this:

```asm
wasmi::engine::executor::handler::exec::u64_load_mem0_offset16_rr:
Lfunc_begin2432:
        .cfi_startproc
        ldrh w8, [x1, #8]
        adds x8, x5, x8
        b.hs LBB2432_2
        subs x9, x4, x8
        ccmp x9, #8, #0, hs
        b.hs LBB2432_3
LBB2432_2:
        mov w0, #2
        ret
LBB2432_3:
        ldr x5, [x3, x8]
        ldur x6, [x1, #16]
        add x1, x1, #16
        br x6
```

Experiments with disalbed `memory64` support at compile-time saw the following assembly:

```asm
wasmi::engine::executor::handler::exec::u64_load_mem0_offset16_rr:
Lfunc_begin2432:
        .cfi_startproc
        ldrh w8, [x1, #8]
        add x8, x8, w5, uxtw
        subs x9, x4, x8
        ccmp x9, #8, #0, hs
        b.hs LBB2432_2
        mov w0, #2
        ret
LBB2432_2:
        ldr x5, [x3, x8]
        ldr x6, [x1, #16]!
        br x6
```

Which displays a significant improvement.

Robbepop:
Closes https://github.com/wasmi-labs/wasmi/issues/1931.


Robbepop:
Honest summary:

- No measurable benchmark improvments beyond noise on M2 Pro.
    - However, M2 Pro has extremely advanced out-of-order execution, branch predictor, a huge I-cache and large pipeline, which narrow those potential performance gains.
- Only 2kB less binary size in Wasmi's executor.
    - ~54kB -> ~52kB: ~4% drop

So the gains aren't amazing, however, the PR's diff is kinda small, too.
Therefore, if these changes cause other non Apple Silicon platforms to perform significantly better, it may be worth it.

However, embedded users that run Wasmi on microcontrollers will probably see more benefits as these CPU have very limited cache, modest branch predictors and execute in-order.

PR #1962, opened 2026-07-03
Add imported function name to mismatch error

kajacx:
Adding the name of the function that has the bad signature is a complete lifesaver, and a little string copying on the error path is well worth it.


Robbepop (approved):
@kajacx LGTM, thanks for the improvement!

PR #1964, opened 2026-07-05
fix `u64` overflow computing `memory64` max size in `effective_address`

aizu-m:
Spotted while auditing the `memory64` load/store translation path. A 64-bit memory declared with the largest allowed maximum traps on a plainly in-bounds constant store:

    (memory i64 1 0x1000000000000)   ;; min 1 page, max 2^48 pages
    (i64.store (i64.const 8) (i64.const 42))  ;; -> MemoryOutOfBounds

The trap comes from translation, not runtime. `effective_address` sizes the memory maximum as `max << page_size_log2()` in `u64`. For `max == 2^48` pages and the default 64 KiB page size (`log2 = 16`) the shift is `2^48 << 16 == 2^64`, which wraps `u64` to `0`, so `address > max_size` holds for every non-zero constant address and the access is folded into an unconditional trap.

`wasmi_core` already does this same byte-size calculation in `u128` (see `crates/core/src/memory/mod.rs` and `ty.rs`); the translator is the one spot still doing it in `u64`. Widened it to match. Added a regression test under `tests/integration/memory64.rs`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,14 @@ Additionally we have an `Internal` section for changes that are of interest to d
 
 Dates in this file are formattes as `YYYY-MM-DD`.
 
+## Unreleased
+
+### Fixed
+
+- Fixed a `u64` overflow when sizing a 64-bit memory's maximum during translation. [#1964]
+  - A `memory64` declared with the largest allowed maximum (`2^48` pages) turned every
+    non-zero constant load/store address into an unconditional `MemoryOutOfBounds` trap.
+
 ## `2.0.0-beta.4` - 2026-07-01
 
 ### Added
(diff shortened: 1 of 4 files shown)


crates/wasmi/tests/integration/memory64.rs
```diff
@@ -0,0 +1,30 @@
...
+    "#;
+    let module = Module::new(&engine, wasm).unwrap();
+    let instance = Instance::new(&mut store, &module, &[]).unwrap();
+    let test = instance
+        .get_func(&store, "test")
+        .unwrap()
+        .typed::<(), i64>(&store)
+        .unwrap();
+    assert_eq!(test.call(&mut store, ()).unwrap(), 42);
+}
```
Robbepop, line 30: We prefer to use `.wast` tests found under `crates/wast/tests/wasmi` directory.
I will create a follow-up PR to transform your `.rs` test into a `.wast` test.

CHANGELOG.md
```diff
@@ -68,6 +76,7 @@ Dates in this file are formattes as `YYYY-MM-DD`.
 - Add `cfg`-guard badges to docs rendered at `docs.rs`. [#1913]
 - Update the `nightly` Rust toolchain used for the Wasmi CI. [#1936]
 
+[#1964]: https://github.com/wasmi-labs/wasmi/pull/1964
```
Robbepop, line 79: Links should be put under their respective release section.
I will move this in a follow-up PR.

Robbepop (approved):
@aizu-m Great catch and thanks for the fix and the test!
I will fix the nits in a follow-up PR.

PR #1991, opened 2026-07-23
Add `InstanceEntity` handle cache with eager warm-up

Robbepop:
Fork of https://github.com/wasmi-labs/wasmi/pull/1990 with eager warm-up phase for further improved performance and a simpler design and implementation with higher startup (instantiation) times as trade-off.


Robbepop:
I just conducted benchmarks over at `wasmi-benchmarks` repo and found that instantiation time is affected but in both directions: instantiation got both slower and faster compared to Wasmi v2.0.0-beta.7 on different test cases. The delta ranges between roughly -10% to +10% and was only measurable for the `lazy/unchecked` cases.

I worried that this change was affecting instantiation performance way more negatively and therefore I feel fine merging this.

PR #1998, opened 2026-07-25
Experiment: swap float-wrapper between `Inst` and `VmState`

Robbepop:
This significantly improves the `count/globals` benchmark by ~10%.
However, this also significantly regresses many call-intense benchmarks by at least the same percentage points.

This can only be merged if we can fix the performance regressions.


Robbepop:
Closed because benchmarks show significant performance regressions.

PR #2001, opened 2026-07-25
Use `NonNull<InstanceEntity>` instead of `Inst` where possible

Robbepop:
the reason `NonNull<InstanceEntity>` is better than `Inst` is because `Inst` might be a float-wrapper and thus not profiting from `Option`'s niche optimization.


Robbepop:
Closed because benchmarks show significant performance regressions.

PR #2015, opened 2026-07-31
Drop `CodeView` from Wasmi's executor and `VmState`

Robbepop:
This is one of the work items to reduce the number of dependend loads associated to `state: &mut VmState` accesses in Wasmi's executor.
This remove `CodeView` as it is no longer being used in the executor since recent commits and moves its `WasmFeatures` information into the `StoreInner` type accessed via the `PrunedStore` in the executor.


Robbepop:
The `audit` CI job failure is not associated to this PR and thus not stopping merging.

PR #2030, opened 2026-08-28
fix: match (ref.func), (ref.extern) and untyped (ref.null) result patterns

jaideeppyne:
`WastRunner::assert_result_core` doesn't handle three reference result patterns that the spec testsuite uses in `assert_return`. `RefFunc(..)` has no match arm and falls through to `_ => false`, so every `(ref.func)` expectation fails; the untyped `(ref.null)` (`RefNull(None)`) is likewise unhandled; and `RefExtern(None)` checks that the reference *is* null, which is inverted — `(ref.extern)` expects a non-null externref with an unspecified value, so a null `externref` currently compares equal and silently passes.

`gc/extern.wast` shows the intended distinction directly: the null cases are written `(ref.null extern)` (lines 42, 47) while the non-null cases use bare `(ref.extern)` (lines 43-46).

Because of the missing `RefFunc` arm, `select.wast` fails at `(assert_return (invoke "join-funcnull" (i32.const 1)) (ref.func))`:

```
expected RefFunc(None) but found FuncRef(Val(Func(Stored { .. })))
```

With this change, `select.wast` from the `wasm-3.0`, `gc` and `function-references` directories passes.

wasmi's API doesn't expose a `Func`'s index, so `(ref.func $f)` is treated like `(ref.func)` and accepts any non-null `funcref` — the same approach Wasmtime's Wast runner takes. Adds a `reference_result_patterns` regression test.

The inverted `RefExtern(None)` case is worth a closer look than the other two: it was a false pass rather than a false failure, so a null externref was satisfying an assertion that requires a non-null one.

Disclosure: this was AI assisted. An agent ran the spec testsuite against the runner and drafted the patch; I verified the behaviour and the spec citations before submitting.

### DIFF
--- a/crates/wast/src/lib.rs
+++ b/crates/wast/src/lib.rs
@@ -424,6 +424,9 @@ impl WastRunner {
                 return v128_matches_or_err(result, expected)
                     .map_err(|error| error.context("evaluation mismatch of `v128` values"));
             }
+            // An untyped `(ref.null)` matches a null reference of any type.
+            (Val::FuncRef(funcref), WastRetCore::RefNull(None)) => funcref.is_null(),
+            (Val::ExternRef(externref), WastRetCore::RefNull(None)) => externref.is_null(),
             (
                 Val::FuncRef(funcref),
                 WastRetCore::RefNull(Some(HeapType::Abstract {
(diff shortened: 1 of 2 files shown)


Robbepop (approved):
LGTM

Robbepop:
@jaideeppyne thank you for the PR and improvements!
Some nits will be fixed in a follow-up PR.

Issue #1859, opened 2026-05-21
Wasmi `v1` panics when running out of fuel lazily translating functions in resumable calls (not during resumption)


matthew-a-thomas:
Hello,

I'm probably missing something obvious, but when I call `wasmi::engine::resumable::ResumableCallOutOfFuel::resume(...)` then I get the following panic.

Could you help me understand what I'm missing?

Thank you.

```
>cargo run
   Compiling hello_wasm v0.1.0 (path\hello_wasm)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 1.31s                                                                                                                                                            
     Running `target\debug\hello_wasm.exe`
Fuel from 0 to 42

thread 'main' (26352) panicked at ~\.cargo\registry\src\index.crates.io-1949cf8c6b5b557f\wasmi-1.0.9\src\engine\executor\instrs.rs:72:32:
the currently used instance must be present
stack backtrace:
   0: std::panicking::panic_handler
             at /rustc/59807616e1fa2540724bfbac14d7976d7e4a3860/library\std\src\panicking.rs:689
   1: core::panicking::panic_fmt
             at /rustc/59807616e1fa2540724bfbac14d7976d7e4a3860/library\core\src\panicking.rs:80
   2: core::panicking::panic_display
             at /rustc/59807616e1fa2540724bfbac14d7976d7e4a3860/library\core\src\panicking.rs:259
   3: core::option::expect_failed
             at /rustc/59807616e1fa2540724bfbac14d7976d7e4a3860/library\core\src\option.rs:2245
   4: enum2$<core::option::Option<ref$<wasmi::instance::Instance> > >::expect<ref$<wasmi::instance::Instance> >
             at ~\.rustup\toolchains\stable-x86_64-pc-windows-msvc\lib\rustlib\src\rust\library\core\src\option.rs:971
   5: wasmi::engine::executor::stack::calls::CallStack::instance_expect
             at ~\.cargo\registry\src\index.crates.io-1949cf8c6b5b557f\wasmi-1.0.9\src\engine\executor\stack\calls.rs:89
   6: wasmi::engine::executor::instrs::execute_instrs
             at ~\.cargo\registry\src\index.crates.io-1949cf8c6b5b557f\wasmi-1.0.9\src\engine\executor\instrs.rs:72
   7: wasmi::engine::executor::EngineExecutor::execute_func
[...]

PR #2032, opened 2026-08-28
Make out-of-fuel resumable when lazily translating a callee

jaideeppyne:
`Func::call_resumable` cannot be resumed when fuel runs out while lazily translating a *callee* mid-execution, which is the default `CompilationMode::LazyTranslation`.

The root function translates and its instructions run, so side effects are already applied. Then translating the callee runs out of fuel and the call returns a plain `Err` rather than `Ok(ResumableCall::OutOfFuel(handle))`, even though the error already carries the correct `required_fuel`. The call is lost and only a full restart is possible.

```
LazyTranslation start=10000 -> Err(ran out of fuel while calling a resumable function: required_fuel=16849)
Eager           start=10000 -> ok [I32(2)] after 0 resumptions
Eager           start=0     -> ok [I32(2)] after 2 resumptions, granted=1207
```

Under `Eager` the identical fuel schedule resumes normally. All four call forms are affected: `call`, `return_call`, `call_indirect`, `return_call_indirect`.

This is not the case fixed by #1916 for #1859. That one is the *root* function's translation, which has no frame to resume into and is deliberately non-resumable. I checked that and left it as it is.

`code_map` builds a `ResumableOutOfFuelError`, but `From<Error> for ExecutionOutcome` folds it into `ExecutionOutcome::Error` and `compile_or_get_func_entry!` halts with `DoneReason::error(...)` instead of `DoneReason::OutOfFuel(...)`. This rewinds `ip` to the call operator and routes through the existing `out_of_fuel!` machinery, the same way `consume_fuel!` and `memory_copy_within` already do.

Found with a fuel self-oracle: a baseline run with metering off, compared against metered runs resumed at every out of fuel point, over 6 fuel schedules and 3 compilation modes, 1586 modules and 5948 calls. After the fix that oracle is clean. Four tests in `resumable_call.rs` fail 4/4 before and pass 4/4 after, and `cargo test --workspace` is green including the 642 spec tests.

## Performance, and where I would like your guidance

This regresses the call hot path and I would rather say so than have you find it.

`cargo bench --bench benches -- --baseline master`, on `execute/call`:

```
flat/0      -8.8%     nested/1    -10.8%     indirect/0  +7.2%
flat/1     -11.2%     nested/8     -6.2%     indirect/1  +8.0%
flat/8      -9.8%     nested/16    +1.3%     imported/1 +12.4%
flat/16     -7.9%                            imported/8  +1.7%
```

The `indirect/0` and `imported/1` regressions reproduced across three runs, including one at `--sample-size 200`, so they are real rather than noise. My machine was loaded, so treat the small numbers as unreliable, but not those two.

The cause is mine, not inherent: I changed `call_func_entry` and `return_call_func_entry` from returning `Control<(Ip, Sp), Break>` to taking `&mut Args` and writing `ip`/`sp` through it, so that the cold error branch can rewind `ip`. That cost is paid on every call to serve a path that almost never runs.

I did not want to guess at a cheaper shape in the executor without knowing your intent here. If you would prefer the error carried back to the handler in `exec.rs`, which already holds `args`, so the hot signatures stay scalar, I am happy to redo it that way. Or if you would rather just take the repro and fix it yourself, that is fine too.

I used Claude Code to build the differential harness and draft the patch, and I ran and reviewed everything above myself.

### DIFF
--- a/crates/wasmi/src/engine/executor/handler/args.rs
+++ b/crates/wasmi/src/engine/executor/handler/args.rs
@@ -236,13 +236,12 @@ impl Args {
     pub fn call_func_entry(
         &mut self,
         store: &mut PrunedStore,
+        op_ip: Ip,
         func: &FuncEntry,
         params: BoundedSlotSpan,
         instance: Option<Inst>,
     ) -> Control<(), Break> {
-        (self.ip, self.sp) =
-            utils::call_func_entry(store, self.ip, self.sp, params, func, instance)?;
-        Control::Continue(())
+        utils::call_func_entry(store, self, op_ip, params, func, instance)
     }
 
     /// Tail-calls `func` with `params` on `instance` with `state` using `self`.
@@ -250,12 +249,12 @@ impl Args {
     pub fn return_call_func_entry(
         &mut self,
         store: &mut PrunedStore,
+        op_ip: Ip,
         func: &FuncEntry,
         params: BoundedSlotSpan,
         instance: Option<Inst>,
     ) -> Control<(), Break> {
-        (self.ip, self.sp) = utils::return_call_func_entry(store, self.sp, params, func, instance)?;
-        Control::Continue(())
+        utils::return_call_func_entry(store, self, op_ip, params, func, instance)
     }
 
     /// Resolves the [`Func`] at `table[index]` of type `func_type` using `state`.
@@ -279,28 +278,12 @@ impl Args {
     pub fn call_wasm_or_host_func(
         &mut self,
         store: &mut PrunedStore,
+        op_ip: Ip,
         func: Func,
         func_entity: NonNull<FuncEntity>,
         params: BoundedSlotSpan,
     ) -> Control<(), Break> {
-        (
-            self.ip,
-            self.sp,
-            self.mem0_ptr,
-            self.mem0_len,
-            self.instance,
-        ) = utils::call_wasm_or_host(
-            store,
-            self.ip,
-            self.sp,
-            func,
-            func_entity,
-            params,
-            self.mem0_ptr,
-            self.mem0_len,
-            self.instance,
-        )?;
-        Control::Continue(())
+        utils::call_wasm_or_host(store, self, op_ip, func, func_entity, params)
     }
 
     /// Tail-calls `func` with `params` with `state` using `self`.
@@ -308,27 +291,12 @@ impl Args {
     pub fn return_call_wasm_or_host_func(
         &mut self,
         store: &mut PrunedStore,
+        op_ip: Ip,
         func: Func,
         func_entity: NonNull<FuncEntity>,
         params: BoundedSlotSpan,
     ) -> Control<(), Break> {
-        (
-            self.ip,
-            self.sp,
-            self.mem0_ptr,
-            self.mem0_len,
-            self.instance,
-        ) = utils::return_call_wasm_or_host(
-            store,
-            self.sp,
-            func,
-            func_entity,
-            params,
-            self.mem0_ptr,
-            self.mem0_len,
-            self.instance,
-        )?;
-        Control::Continue(())
+        utils::return_call_wasm_or_host(store, self, op_ip, func, func_entity, params)
     }
 
     /// Pops the top-most frame from the call stack.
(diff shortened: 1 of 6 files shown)


Robbepop:
Hi @jaideeppyne thank you again for the PR!

This primary issue (as I understand it) is that the state of the func entry is not reset upon running out of fuel when being called at root. This is already an edge case but one we should care for. It should be enough to make sure that the function entry is simply reset when encountering this state if that's possible, so it would simply trigger another re-compile after the user provided enough fuel for the translation to succeed.

Currently, this broken state prevent exactly that. I consider this an edge-case because immediately running into out-of-fuel at root can happen but is unlikely when given enough fuel.

The performance regressions though are unacceptable for any fix of that kind. So we have to find a better way to fix this.

jaideeppyne:
Thanks for looking so quickly.

Agreed on the performance, unconditionally. A fix for an edge case should not cost anything on the call path, and I would not want this merged as it stands.

Two corrections, one of them mine.

First, mine: my initial push was missing the tests. I reverted the tree to take a clean benchmark baseline and my patch file turned out to be source only, so the four tests I described were not actually in the PR. They are in now, and they fail 4/4 on `76f261c` and pass 4/4 with the change. Sorry for the noise.

Second, on the diagnosis. This is not the root function. `main` translates and starts executing, and it is the *callee* whose lazy translation runs out of fuel. The root case is the one #1916 handles, and I left that alone since there is no frame to resume into.

I also tried your suggestion of resetting the entry so a later call recompiles, and it turns out the entry is not actually stuck today. On unfixed `main`, a plain retry with more fuel already succeeds. The problem is what the retry costs. Same module, with a global incremented before the call:

```
first call ok? false
side effect after failed call: seen = 1
restart result = Ok(2)
side effect after restart:     seen = 2
```

The root already ran, so restarting runs it again and the side effect lands twice. That is the thing resuming avoids, and the engine already knows the exact `required_fuel` at that point, so the information needed to resume is there.

On a cheaper implementation: the cost is entirely mine. I changed `call_func_entry` and `return_call_func_entry` from returning `Control<(Ip, Sp), Break>` to taking `&mut Args`, purely so the cold branch can rewind `ip`. Every call pays for that. The handlers in `exec.rs` already hold `args` and the pre-decode `ip`, so the out of fuel case could be handled there and the hot signatures could stay scalar. I am happy to redo it that way and re-benchmark, or to drop the patch and leave you the repro if you would rather shape it yourself.

Robbepop:
What side effects are there if we are only talking about root function calls?
To my knowledge there are none and if re-calling the same function by root after re-fueling fixes the issue I actually fail to see the underlying problem in all honesty.

jaideeppyne:
Sorry, I think I explained it badly. It is not a root function call. The fuel runs out translating a function that the root *calls*, after the root has already started executing.

That is the whole module from those numbers:

```wat
(module
  (global $seen (mut i32) (i32.const 0))
  (func $helper (result i32) ...2000 instructions... (i32.const 2))
  (func (export "main") (result i32)
    (global.set $seen (i32.add (global.get $seen) (i32.const 1)))
    (call $helper))
  (func (export "seen") (result i32) (global.get $seen)))
```

`main` is small and translates fine. It runs, and `global.set` takes effect. Then it reaches `call $helper`, `$helper` has not been translated yet, and translating it exhausts the fuel. So the side effect is `main`'s own `global.set`, which has already happened by the time the call fails.

Re-calling `main` does complete, but it runs `global.set` a second time, which is why `seen` goes to 2. There is no way for the caller to avoid that, because the write already landed in the store before the error came back.

If `$helper` were eagerly translated, or if `main` were called with enough fuel, none of this arises. It needs lazy translation plus fuel running out at that particular point.

That said, this is your call. If you consider a duplicated side effect acceptable for how rare this is, that is a completely reasonable position and I will close the PR. I mainly wanted to be sure the case was clear, since it is a different one from #1859.

Robbepop:
@jaideeppyne thank you for the better explanation, now I understand the actual problem. So this is _not_ about root function calls but about nested function calls where the caller isn't root (aka the host) but another Wasm function. Indeed _that_ can have side effects and this might be a serious issue to some users encountering this problem.

So we should ideally fix it.
I will look into this.

jaideeppyne:
Glad that landed, and sorry it took two goes to say it clearly.

Yes, exactly that: the caller is another Wasm function that has already run, not the host.

Since you want to look into it yourself, take whatever is useful here. The four tests in `resumable_call.rs` cover `call`, `return_call`, `call_indirect` and `return_call_indirect`, and they fail on `76f261c` with the error already carrying the right `required_fuel`, so they should work against any shape of fix. The side effect module from the comment above is the shortest way to see why restarting is not equivalent.

I am happy to close this and leave you the tests and the repro, or to redo the fix keeping the hot signatures scalar and handling the out of fuel case in the `exec.rs` handlers, and re-benchmark. Whichever suits you, just say.

PR #2033, opened 2026-09-01
Fix release date

Robbepop:


Robbepop:
audit CI job fails due to wasmtime dependency which is just used in Wasmis fuzzer and thus non-blocking.

Issue #774, opened 2023-10-14
Implement the Wasm `function-references` proposal


Robbepop:
This is about adding support for the [`function-references` Wasm proposal](https://github.com/WebAssembly/function-references).

The `function-references` Wasm proposal has been recently moved to Phase 4 and thus stabilization can be expected to happen soon.

Having `function-references` support in `wasmi` will be an enabler for use cases that rely on the new Wasm `gc` (garbage collection) proposal.

PR #2035, opened 2026-09-03
Add `matches` methods to `ValType` and `FuncType`

### LINKED ISSUE #1746: Add `matches` method to `ValType` and `FuncType`
Blocked by https://github.com/wasmi-labs/wasmi/issues/774.

Wasmi currently has `PartialEq` and `Eq` derives on `FuncType` for comparisons.
After `function-references` Wasm proposal support has been added, those trait impls need to be removed as they do not support sub-typing. Instead, we need to add the following methods, as done in Wasmtime:

- [`ValType::matches`](https://docs.rs/wasmtime/39.0.1/wasmtime/enum.ValType.html#method.matches)
- `FuncType::matches`

cc @BoilingFusion

HernaniSamuel:
Adds `ValType::matches` and `FuncType::matches` — on both `wasmi_core::FuncType`
and the public `wasmi::FuncType` wrapper — mirroring the equivalent methods in
Wasmtime's `runtime/types.rs`.
 
Parameter types are checked contravariantly and result types covariantly, as the
Wasm function subtyping rules demand:
 
```rust
self.params().len() == other.params().len()
    && self.results().len() == other.results().len()
    && self.params().iter().zip(other.params()).all(|(a, b)| b.matches(a))
    && self.results().iter().zip(other.results()).all(|(a, b)| a.matches(b))
```
 
`ValType` is still a flat enum with no `Ref(RefType)` case, so it carries no
subtyping relation today and `ValType::matches` reduces to `self == other`.
That makes the variance above unobservable for now — `a.matches(b) ==
b.matches(a)` holds for every pair — but writing it in the correct form up
front means it stays correct once `function-references` (#774) introduces
real subtyping between reference types, instead of having to be rediscovered
then. The doc comment on `ValType::matches` records why the method currently
coincides with equality, so it doesn't read as a pointless one-liner.
 
## On the `PartialEq`/`Eq` derives
 
The issue also calls for removing the `PartialEq`/`Eq` derives on
`FuncType`. This PR intentionally leaves them in place, per @Robbepop's
comment on the issue: without `function-references` a plain equality check
is still sufficient, and these methods are worth adding ahead of time only
so the subtyping logic is already in the right shape.
 
They are also still required internally. `wasmi::FuncType` is interned via
`DedupArena<Key, FuncType>` in `crates/wasmi/src/engine/func_types.rs`, and
`DedupArena::alloc` requires `T: Hash + Ord + Clone`. Since `Ord: Eq`,
dropping `Eq` would break that dedup table — which performs a
structural-identity comparison unrelated to subtype matching. Removing the
derives is therefore best left to the `function-references` PR, which can
adjust the registry at the same time.
 
Existing `==`/`!=` call sites (such as the import signature check in
`module/instantiate/mod.rs`) are left untouched for the same reason:
swapping them to `.matches()` is behaviourally identical today and would
only inflate the diff.
 
## Naming
 
The methods are named `matches` as the issue requests, following Wasmtime.
Worth noting that `TableType` and `MemoryType` already spell the same
concept `is_subtype_of` in this codebase — happy to rename for internal
consistency if you'd prefer that.
 
## Testing
 
Unit tests alongside each method:
 
- `ValType`: all seven variants checked pairwise against equality, plus an
  explicit `FuncRef`/`ExternRef` case in both directions.
- `wasmi_core::FuncType`: identical signatures, two empty signatures, a
  differing param type, a differing result type, and differing arity (extra
  param and extra result, both directions).
- `wasmi::FuncType`: covers the delegating wrapper.
`cargo test -p wasmi_core`, `cargo test -p wasmi` and
`cargo clippy --workspace --locked --exclude fuzz -- -D warnings` all pass.
 
Closes #1746
 
---
 
**Disclosure:** this was AI assisted. An agent wrote the patch and tests; I
reviewed the implementation and the reasoning about the deferred derives
before submitting.


HernaniSamuel:
The `Audit` check here is failing on two RUSTSEC advisories in `wasmtime` 47.0.3 (RUSTSEC-2026-0268, RUSTSEC-2026-0269), fixed by bumping to `>=47.0.4`. Confirmed this is pre-existing and unrelated to this PR — the same `Audit` job is already failing on `main` as of the latest CI run there, so it's blocking every open PR right now, not just this one. Flagging in case it's not already being tracked.

Robbepop:
Hi @HernaniSamuel , sorry for taking so long to respond. Life is currently very busy for me.
Also I have to think about AI generated contributions. In general I like the LLM policy of the Rust project which can be compressed to:

> Pre-arranged, non-critical, high-quality, well-tested, and well-reviewed code changes that are originally created by an LLM are allowed, with disclosure.
> 
> It's fine to use LLMs to answer questions, analyze, distill, refine, check, suggest, review. But not to create.

HernaniSamuel:
Fair point, I get the distinction now. Should've flagged upfront that an agent would actually write the patch, not just that I had a green light on the issue.

For what it's worth, I read and verified every line before submitting, and made sure I understood the reasoning behind the change well enough to explain it myself. Happy to close this and redo it by hand if you'd rather. Just let me know how you want to proceed.

Robbepop:
> Fair point, I get the distinction now. Should've flagged upfront that an agent would actually write the patch, not just that I had a green light on the issue.
> 
> For what it's worth, I read and verified every line before submitting, and made sure I understood the reasoning behind the change well enough to explain it myself. Happy to close this and redo it by hand if you'd rather. Just let me know how you want to proceed.

@HernaniSamuel you crafting your own changes would be exactly what I intended with Wasmi issues that are approachable. This should provide a much better way to learn the codebase and improve your own programming skills via the code review. If I review AI generated code, nobody really learns anything from it. You can re-use this PR by force-pushing your own changes to it. I'd encourage you to do that.

HernaniSamuel:
Got it, thanks for explaining! I'll work on my own changes.

HernaniSamuel:
The core check converges with the earlier draft because this specific rule leaves little room for different correct implementations. I wrote tests covering each failure case separately and verified them by intentionally breaking the implementation, then documented the behavior based on my understanding of the subtyping rules. Happy to walk through the reasoning during review.

Issue #1835, opened 2026-04-07
Use Rust's unstable `rust-preserve-none` ABI for execution handlers


Robbepop:
Rust Tracking Issue: https://github.com/rust-lang/rust/issues/151401

This ABI could be very useful to further improve and stabilize performance of the tail-call based dispatch in Wasmi.

This feature has been implemented recently into the Rust compiler.
For targets without support for this ABI, it currently falls back to the Rust calling convention.
This isn't ideal since we'd use `sysv64` on `x86_64` (e.g. on Windows) but at least we could gather some experience and benchmarks if we implemented support for this feature into Wasmi behind the `nightly` (or `unstable`?) crate feature, similar to how it has been done with the unstable `become` keyword.

PR #2036, opened 2026-09-03
extern "rust-preserve-none" support

coolreader18:
Implements #1835. I got nerd-sniped by the blog post :)

I reorganized/deduplicated some macro stuff, since now there are *three* possible abis `Handler` could have. My machine is too noisy to meaningfully benchmark this, but looking at the asm it does eliminate a few instructions:

```asm
; before
wasmi::engine::executor::handler::exec::global_get_u64_r:
	mov rax, rsi
	movq r9, xmm0
	mov r10d, dword ptr [rsi + 8]
	add rsi, 16
	shl r10, 4
	mov r9, qword ptr [r9 + r10 + 64]
	mov r9, qword ptr [r9]
	jmp qword ptr [rax + 16]

; after
wasmi::engine::executor::handler::exec::global_get_u64_r:
	mov eax, dword ptr [r13 + 8]
	shl rax, 4
	mov rax, qword ptr [rsi + rax + 64]
	mov rdx, qword ptr [rax]
	mov rax, qword ptr [r13 + 16]
	add r13, 16
	jmp rax
```

```sh
$ cargo +nightly asm -p wasmi --lib memory_copy --no-default-features -F stable --simplify 4 | wc -l 
169
$ cargo +nightly asm -p wasmi --lib memory_copy --no-default-features -F unstable --simplify 4 | wc -l 
135
```


Robbepop:
Hi @coolreader18 and thank you for the PR!

I will check out your branch locally and conduct some benchmarks to observe the effect on performance of your changes. Unfortunately I only have an Apple M2 Pro machine which might not be influenced as strongly as an `x86_64` machine.

Besides that I really liked your execution handler macro deduplication approach!

The failing `audit` CI job has nothing to do with the changes introduced in this PR.

PR #2040, opened 2026-09-11
c_api: pass callback directly to create_function

vividvisit:
Pass `callback` directly to `create_function` in `wasm_func_new` instead of forwarding it through a closure with the same arguments.

This removes Clippy's `redundant_closure` finding and leaves the callback type and behavior unchanged.

Validation:

- `cargo fmt --all -- --check`
- `git diff --check`
- A scoped `cargo test -p wasmi_c_api_impl --lib` build was attempted, but the local disk filled while compiling existing dependencies.x


Robbepop:
@vividvisit thanks! unfortunately CI is unhappy with the changes.

Robbepop:
Hi @vividvisit , so what exactly did you change besides docs and a `clippy` allow after your latest force push? It seems to me that the closure is necessary after all.

vividvisit:
@Robbepop

> Hi @vividvisit , so what exactly did you change besides docs and a `clippy` allow after your latest force push? It seems to me that the closure is necessary after all.

You're right. I rechecked this against the exact Clippy toolchain used by the project, and the original closure passes without any allowance. After restoring the necessary closure, my latest commit only added a comment and an unnecessary clippy allow, so there is no meaningful change left.

I should have recognized that the original premise of the PR was incorrect instead of force-pushing a cosmetic change. Sorry for the unnecessary churn. I'll close the PR.

